PR #3, opened 2020-06-30
feat: creating fullnode messages integration to wallet service 

pedroferreira1:


fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import WebSocket from 'ws';
+import { SERVER, AWS_ACCESS_KEY, AWS_SECRET_KEY, QUEUE_URL } from './config';
+import { EventEmitter } from 'events';
+import AWS from 'aws-sdk';
+
+// we need to set a region even if we don't make any calls
+AWS.config.update({ region: 'us-east-1' });
```
obiyankenobi: We should get this from config. Also, I don't think the comment applies here. We are actually making calls to AWS.

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+      this.setState(WalletServiceState.CLOSED);
+    });
+
+    this.ws.on('message', (message: string) => {
+      const data: wsDataType = JSON.parse(message) as wsDataType;
+      this.handleMessage(data);
+    });
+  }
+
+  onOpen(): void {
```
obiyankenobi: We might want to make this method (and similar ones) private.

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+    this.state = state;
+    this.emit('state', state);
+    console.log('State updated to', state);
+  }
+
+  handleMessage(data: wsDataType): void {
+    switch(data.type) {
+      case 'subscribed':
+        this.onSubscribed(data);
+      break;
```
obiyankenobi: `break` should be indented?

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+type wsDataType = { [_: string]: any};
+
+enum WalletServiceState {
+  CLOSED,
+  CONNECTING,
+  READY,
+}
+
+
+class WalletService extends EventEmitter {
```
obiyankenobi: Do we need it to emit events? I'm more curious, as I don't think there are no downsides in doing it.

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+const sqs = new AWS.SQS({
+  apiVersion: '2012-11-05',
+  endpoint: QUEUE_URL,
+  accessKeyId: AWS_ACCESS_KEY,
+  secretAccessKey: AWS_SECRET_KEY,
+});
+
+type wsDataType = { [_: string]: any};
+
+enum WalletServiceState {
```
obiyankenobi: We need to debate this name. I think it's not the wallet service state, but the full node connection state, even though they're related.

fullnode-messages/src/index.ts
pedroferreira1: I also don't see any downsides. Currently I am emitting the events every time my state changes, e.g. the full node becomes offline, so the connection will be closed.

It's not 100% clear to me from the design exactly how you are gonna use this service but I guess it's important to listen to those events and know when the data is not being updated in real time.

fullnode-messages/src/index.ts
pedroferreira1: Yeah, I agree. I used my class name as `WalletService`, that's why I used this name (in that context makes sense). But it's not a good name.

Maybe `WalletServiceMessages` and `WalletServiceMessagesState`?

fullnode-messages/src/index.ts
pedroferreira1: No, I fixed it. Probably was some probably when indenting typescript. Weird that everything was still working fine. Fixed it, thanks

fullnode-messages/src/index.ts
pedroferreira1: Yeah, makes sense. Changed it.

fullnode-messages/src/index.ts
pedroferreira1: That's correct, I've set all methods as private, except `start`. Should we also let `sendMessage` public? (I don't think so).

fullnode-messages/src/index.ts
obiyankenobi: > It's not 100% clear to me from the design exactly how you are gonna use this service but I guess it's important to listen to those events and know when the data is not being updated in real time.

To be honest, I don't see the wallet-service monitoring this service. I'd guess it's the other way around. When it disconnects, it'll probably have to notify the wallet-service (or only notify when there's a reconnect). There's no way the wallet-service can listen to these emitted events, only the ones put in an SQS queue.

fullnode-messages/src/index.ts
obiyankenobi: I think this is very specific to the full node connection, so I'd name `FullNodeConnection` or something like that.

fullnode-messages/README.md
```diff
@@ -0,0 +1,14 @@
+# Wallet Service Messages
+
+Connects the full node with the wallet service, listening for new specific websocket messages and adding them to a SQS queue.
+
+# Run
+
+- Create your configuration file (`cp src/config.ts.template src/config.ts`);
+- Fill the variables at `src/config.ts`;
+- Run `npm run tsc` to compile ts files to js;
+- Run `npm start`.
```
msbrogli, line 10: Can't we just compile ts files to js on the `start` command?

fullnode-messages/README.md
```diff
@@ -0,0 +1,14 @@
+# Wallet Service Messages
+
+Connects the full node with the wallet service, listening for new specific websocket messages and adding them to a SQS queue.
+
+# Run
```
msbrogli, line 5: Is there a dev mode that automatically reloads if the source is changed?

fullnode-messages/src/config.ts.template
```diff
@@ -0,0 +1,14 @@
+// Full node URL to listen to the ws messages
+export const SERVER = '';
```
msbrogli, line 2: Maybe add an example? wss://node1.mainnet.hathor.network/?

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+AWS.config.update({ region: AWS_REGION });
+
+const sqs = new AWS.SQS({
+  apiVersion: '2012-11-05',
+  endpoint: QUEUE_URL,
+  accessKeyId: AWS_ACCESS_KEY,
+  secretAccessKey: AWS_SECRET_KEY,
+});
+
+type wsDataType = { [_: string]: any};
```
msbrogli: Missing a space before `}`?

fullnode-messages/src/index.ts
msbrogli: I agree with @obiyankenobi. Maybe `HathorWebsocketToSQS`?

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  // Heartbeat interval in milliseconds.
+  readonly heartbeatInterval: number = 3000;
+  // Connection timeout.
+  readonly connectionTimeout: number = 5000;
+  // Retry connection interval in milliseconds.
+  readonly retryConnectionInterval: number = 1000;
+  // Open connection timeout.
+  readonly openConnectionTimeout: number = 20000;
+  // Last time the ping message was sent
+  latestPingDate: Date | null = null;
```
msbrogli: I'd use a bool `waitingPongReply` instead of setting `latestPingDate` as null between pings. What do you think?

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+    const deduplicationId = `new-tx-${data.tx_id as string}`;
+    const params = {
+      MessageBody: JSON.stringify(data),
+      QueueUrl: QUEUE_URL,
+      MessageGroupId: 'wallet-service-new-tx',
+      MessageDeduplicationId: deduplicationId,
+    };
+
+    sqs.sendMessage(params, function(err, data) {
+      if (err) console.log(err, err.stack); // an error occurred
```
msbrogli: Maybe we should put the data in an internal queue and retry in case of error. If the queue is full, we can raise an unrecoverable error (or start syncing again after the SQS is back).

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  private sendMessage(message: string): void {
+    if (this.ws) {
+      // XXX Should we check if ws is ready to send messages?
+      this.ws.send(message);
+    }
+  }
+}
+
+const walletService: WalletService = new WalletService();
```
msbrogli: I think we should leave just this part in the index.ts and move the class to another file.

fullnode-messages/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  // Open connection timeout.
+  readonly openConnectionTimeout: number = 20000;
+  // Last time the ping message was sent
+  latestPingDate: Date | null = null;
+  // Settimeout reference when waiting for the pong response
+  timeoutTimer: NodeJS.Timeout | null = null;
+  // Setinterval reference for the ping interval
+  heartbeat: NodeJS.Timeout | null = null;
+
+  start(): void {
```
msbrogli: You haven't included the `stop()` method.

msbrogli (changes requested):

fullnode-messages/src/index.ts
obiyankenobi: I agree, `sendMessage` should not be public.

fullnode-messages/README.md
jansegre: We can ship the files (on source packaging, docker images, etc) and not have to compile them at all. Which is great because it makes `npm start` not depend on dev dependencies.

Otherwise for development I'd rather use a watch command that auto-recompiles and reloads when needed.

fullnode-messages/src/config.ts.template
jansegre: I actually don't really like the `config.ts.template`/`config.js.template` approach. It's certainly useful, but we can't make configurable and functional docker images. For HathorDice I went with a `process.env` only approach (which could definitely be improved, by using a `config` module to organize all available options), but I think it's far easier to auto-build and have deployable and configurable images.

fullnode-messages/src/index.ts
pedroferreira1: Yeah, the wallet service probably won't use it.

As we said before, adding those events has no harm for performance or the code, so I will just keep it in case we decide to use later.

fullnode-messages/src/index.ts
pedroferreira1: I think I will go with `FullNodeConnection`. What do you think @msbrogli? I think `ToSQS` is really specific.

fullnode-messages/src/index.ts
pedroferreira1: Separated in two different files.

fullnode-messages/src/index.ts
pedroferreira1: Done, can you see if it's good?

fullnode-messages/src/index.ts
pedroferreira1: Yes, it's better I agree. Updated it.

fullnode-messages/README.md
pedroferreira1: Yeah, I agree we shouldn't pack ts files, only js so `npm start` does not depend on dev dependencies. I will add the `watch` command also as @jansegre said in the other comment.

andreabadesso:
Closing this as this was replaced by the [sync daemon](https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/)

andreabadesso:
Closing this as it was replaced by the [sync daemon](https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/)

PR #4, opened 2020-07-08
feat: basic service working

obiyankenobi:


src/api/addresses.ts
```diff
@@ -0,0 +1,48 @@
...
+ * This lambda is called by API Gateway on GET /addresses
+ */
+export const get: APIGatewayProxyHandler = async (event) => {
+  let walletId: string;
+  const params = event.queryStringParameters;
+  if (params && params.id) {
+    walletId = params.id;
+  } else {
+    await closeDbConnection(mysql);
+    return {
```
pedroferreira1: All files have exactly this return. I think would be good to create a method on utils `return_200` that receives only the object and creates the object with status code and body

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+ * lamdba to asynchronously add new wallet info to database
+ *
+ * This lambda is called by API Gateway on POST /wallet
+ */
+export const load: APIGatewayProxyHandler = async (event) => {
+  let body;
+  try {
+    body = JSON.parse(event.body);
+    // event.body might be null, which is also parsed to null
+    if (!body) throw new Error('body is null');
```
pedroferreira1, line 71: For me this is the same case of a missing or invalid parameter. So it makes no sense to handle it differently, we should return the erro to the user.

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+  } catch (e) {
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.INVALID_PARAMETER, parameter: 'xpubkey' }),
```
pedroferreira1: The invalid parameter here isn't the `body`?

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+    };
+  }
+
+  const xpubkey = body.xpubkey;
+  if (!xpubkey) {
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.MISSING_PARAMETER, parameter: 'xpubkey' }),
```
pedroferreira1: I think this return is bad. You've been using this as a missing request parameter but this xpubkey is a missing key inside the body parameter, so this could give a false impression to the user who could just add a `xpubkey` parameter. You could return here 'invalid_parameter' for the body again. (If I understood correctly `body` is the only parameter here

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+  let status = await getWallet(mysql, walletId);
+  if (status) {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.WALLET_ALREADY_LOADED, status }),
+    };
+  }
+
+  const maxGap = parseInt(process.env.MAX_ADDRESS_GAP, 10);
```
pedroferreira1, line 92: I'm thinking if we should allow the user to send a custom `maxGap` in the request. If we don't do that we are saying to all the use cases that need more than 20 addresses on the gap (maybe it's not an expressive quantity) that they need to run their own wallet service to use Hathor for their use cases, which goes totally in the opposite side of what we sell, that is user friendly to create your use case.

However if we allow anything here we let people send a huge number, which could increase the database really fast. So I think we could have the `DEFAULT_ADDRESS_GAP=20` and `MAX_ADDRESS_GAP=100`. Then we just allow uses cases until 100 address gap. I think it's  a small change that does no harm to the code and allow more use cases to use it.

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+  // invoke lambda asynchronously to handle wallet creation
+  const lambda = new Lambda({
+    apiVersion: '2015-03-31',
+    endpoint: process.env.STAGE === 'local'
+      ? 'http://localhost:3002'
+      : `https://lambda.${process.env.AWS_REGION}.amazonaws.com`,
+  });
+  const params = {
+    // FunctionName is composed of: service name - stage - function name
+    FunctionName: `${process.env.SERVICE_NAME}-${process.env.STAGE}-loadWalletAsync`,
```
pedroferreira1: Why do you need this to be in another lambda?

DATABASE.md
```diff
@@ -0,0 +1,79 @@
+# Database
```
msbrogli, line 1: Isn't it better to have a `create_database.sql` file, so we can redirect to mysql client?

README.md
```diff
@@ -0,0 +1,116 @@
+# hathor-wallet-service
+
+The hathor-wallet-service is the backend for all official Hathor wallets. It's highly inspired in the [Bitcore Wallet
```
msbrogli: We've changed so many things that I'm not sure it is highly inspired anymore, hehe. But it is ok to have it here.

fullnode-bridge.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable */
```
msbrogli, line 1: Just drop a comment explaining what it is for and that it is not supposed to be used in production.

src/api/addresses.ts
```diff
@@ -0,0 +1,48 @@
...
+import 'source-map-support/register';
+
+import { ApiError } from '@src/api/errors';
+import {
+  getWallet,
+  getWalletAddresses,
+} from '@src/db';
+import { closeDbConnection, getDbConnection } from '@src/utils';
+
+const mysql = getDbConnection();
```
msbrogli, line 19: I'm not sure we should get a connection while loading the module.

src/api/addresses.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/*
+ * Get the addresses of a wallet
+ *
+ * This lambda is called by API Gateway on GET /addresses
+ */
+export const get: APIGatewayProxyHandler = async (event) => {
```
msbrogli, line 26: Does `event` has a type?

src/api/addresses.ts
```diff
@@ -0,0 +1,48 @@
...
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.MISSING_PARAMETER, parameter: 'id' }),
+    };
+  }
+
+  const status = await getWallet(mysql, walletId);
+  if (!status) {
+    await closeDbConnection(mysql);
```
msbrogli: We're copying it around. Maybe we should create an error method? Or encapsulate this whole module in a class?

src/api/balances.ts
```diff
@@ -0,0 +1,54 @@
...
+import 'source-map-support/register';
+
+import { ApiError } from '@src/api/errors';
+import {
+  getWallet,
+  getWalletBalances,
+} from '@src/db';
+import { closeDbConnection, getDbConnection } from '@src/utils';
+
+const mysql = getDbConnection();
```
msbrogli, line 19: Same comments I did for `src/api/balances.ts`.

src/api/txhistory.ts
```diff
@@ -0,0 +1,90 @@
...
+  getWalletTxHistory,
+} from '@src/db';
+import { closeDbConnection, getDbConnection } from '@src/utils';
+
+const mysql = getDbConnection();
+
+// XXX add to .env or serverless.yml?
+const MAX_COUNT = 15;
+
+// TODO get from lib
```
msbrogli: Important TODO.

src/api/txhistory.ts
```diff
@@ -0,0 +1,90 @@
...
+  const status = await getWallet(mysql, walletId);
+  if (!status) {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.WALLET_NOT_FOUND }),
+    };
+  }
+
+  const history = await getWalletTxHistory(mysql, walletId, tokenId, skip, count);
```
msbrogli, line 77: Instead of passing all these parameters over and over, isn't it better to have a class?

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.WALLET_ALREADY_LOADED, status }),
+    };
+  }
+
+  const maxGap = parseInt(process.env.MAX_ADDRESS_GAP, 10);
+
+  // add to wallet table with 'creating' status
+  status = await createWallet(mysql, walletId, xpubkey, maxGap);
```
msbrogli: You are subject to a race condition here. When you check whether a wallet exists, if it doesn't, you should leave a marker saying that it will be created. I think the best way here is to try to insert and handle the error if the key already exists.

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  // invoke lambda asynchronously to handle wallet creation
+  const lambda = new Lambda({
+    apiVersion: '2015-03-31',
```
msbrogli: Where does this magic date come from? Maybe we should create an internal method that handles that for us.

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+    // FunctionName is composed of: service name - stage - function name
+    FunctionName: `${process.env.SERVICE_NAME}-${process.env.STAGE}-loadWalletAsync`,
+    InvocationType: 'Event',
+    Payload: JSON.stringify({ xpubkey, maxGap }),
+  };
+  try {
+    // TODO setup lambda error handling. It's not an error on lambda.invoke, but an error during lambda execution
+    await lambda.invoke(params).promise();
+  } catch (e) {
+    // TODO handle
```
msbrogli: You should at least log something and return `success: false`.

src/api/wallet.ts
```diff
@@ -0,0 +1,169 @@
...
+  xpubkey: string;
+}
+
+/*
+ * This does the "heavy" work when loading a new wallet, updating the database tables accordingly. It
+ * expects a wallet entry already on the database
+ *
+ * This lambda is called async by another lambda, the one reponsible for the load wallet API
+ */
+export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
```
msbrogli: I think the name is misleading. I think `loadWallet` is used to load something that already exists. Maybe rename it to `createWallet` or `startWallet` or `prepareFreshWallet`. What do you think?

src/utils.ts
```diff
@@ -0,0 +1,116 @@
+import { Address, HDPublicKey, Networks } from 'bitcore-lib';
+import { createHash, HexBase64Latin1Encoding } from 'crypto';
+import serverlessMysql, { ServerlessMysql } from 'serverless-mysql';
+
+import { StringMap } from '@src/types';
+
+// TODO get from hathor-lib or maybe env?
```
msbrogli: Why not `hathor-wallet-lib`?

msbrogli (changes requested):

README.md
obiyankenobi: That's true, I'll update this.

src/api/addresses.ts
obiyankenobi: This is the correct way to do for lambda handlers. [This example](https://docs.aws.amazon.com/lambda/latest/dg/services-rds-tutorial.html) uses python, but the idea is the same: "Executing pymysql.connect() outside of the handler allows your function to re-use the database connection for better performance."

src/api/addresses.ts
obiyankenobi: Yes, it's already typed by APIGatewayProxyHandler: `export type APIGatewayProxyHandler = Handler<APIGatewayProxyEvent, APIGatewayProxyResult>;`

It means the parameter is `APIGatewayProxyEvent` and it returns a `APIGatewayProxyResult`.

src/api/addresses.ts
obiyankenobi: What code do you mean? Just `closeConnection` or the whole process of checking if wallet exists and returning the error?

src/api/txhistory.ts
obiyankenobi: I don't see any advantage, at least not at the moment. What other methods would we have in the class? Where else would it be used?

src/api/wallet.ts
obiyankenobi: I agree. But I'm not sure about the marker idea. Shouldn't just handling the error be enough?

src/api/wallet.ts
obiyankenobi: It's standard aws lambda invocation. We can have another method, but it'd have exactly the same code.

src/utils.ts
obiyankenobi: Yes, I meant `hathor-wallet-lib`.

src/db.ts
```diff
@@ -0,0 +1,627 @@
+import { strict as assert } from 'assert';
```
pedroferreira1, line 8: I think this file should be split. We have methods here that deal with different parts of the wallet and it's already 627 lines. If we don't split now it will be more than 1k lines before the first release, which would be bad to maintain

src/db.ts
```diff
@@ -0,0 +1,627 @@
...
+  mysql: ServerlessMysql,
+  txId: string,
+  timestamp: number,
+  walletBalanceMap: StringMap<TokenBalanceMap>,
+): Promise<void> => {
+  const entries = [];
+  for (const [walletId, tokenBalanceMap] of Object.entries(walletBalanceMap)) {
+    for (const [token, tokenBalance] of tokenBalanceMap.iterator()) {
+      // on wallet_balance table, balance cannot be negative (it's unsigned). That's why we use
+      // balance as (tokenBalance < 0 ? 0 : tokenBalance). In case the balance is negative, there
```
pedroferreira1: When the balance can be negative?

src/txProcessor.ts
```diff
@@ -0,0 +1,263 @@
...
+ * @param now - Current timestamp
+ * @param blockRewardLock - The block reward lock
+ */
+const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) => {
+  // TODO mysql error treatment
+
+  const txId = tx.tx_id;
+
+  let heightlock = null;
+  if (tx.version === 0 || tx.version === 3) {
```
pedroferreira1: This values should be move to a constants file or if we use our lib, we can get it from there

src/api/wallet.ts
obiyankenobi: That's what's being done. The error is thrown and handled in the catch block.

src/api/wallet.ts
obiyankenobi: Not per se. There isn't a `body` parameter. The user should send `{"xpubkey": "myxpub"}`. Instead of `INVALID_PARAMETER`, maybe `INVALID_PAYLOAD`?

src/api/wallet.ts
obiyankenobi: It depends on what you call a parameter. For GET requests, the parameters go in the URL. In this case, for a POST request, since the content type is json, I consider `xpubkey` to be a POST parameter, as it is expected to be in the payload. Again, the user does not send anything called `body`. The `body` in the API Gateway context is the payload.

src/api/wallet.ts
obiyankenobi: We have to debate what's our goal. I'd say most use cases are ok with 20 and if they need more, they could run their own service. Or even get in touch with us and we can increase the limit manually.

src/api/wallet.ts
obiyankenobi: Because it has to run async. There's no way to send a success message to the user and keep loading the wallet in this same function.

src/db.ts
obiyankenobi: What's your suggestion for splitting it?

src/db.ts
obiyankenobi: When I send tokens to another wallet. My wallet's balance in that tx will be negative.

src/txProcessor.ts
obiyankenobi: Ah, I didn't know we had it in the lib. Better to get it from there.

src/api/wallet.ts
obiyankenobi: Just to make it clear, `body` is what the API gateway calls the payload, both for the request and response.

pedroferreira1: Yeah, now I understand what `body` is and I agree `INVALID_PARAMETER` is not really good here. We don't even have a body.

I like `INVALID_PAYLOAD`, as you suggested

src/api/wallet.ts
pedroferreira1: As in the other comment, now I get what the `body` is. It's ok

src/db.ts
pedroferreira1: Address, transaction, token, wallet (core).

src/db.ts
pedroferreira1: Ah okay, you are talking about the tx balance in this line. Just add this in the comment because you start talking about the wallet balance, that can't be negative and that what got me confused.

"on wallet_balance table, balance cannot be negative" and then "In case the balance is negative".

src/db.ts
obiyankenobi: Address and wallet might be useful, as there are some tables for it. But transactions and token might not be useful. Also, there are other functions that don't fall into any of these categories. 

Anyway, I'd prefer to do this refactor in a separate (future) PR:
- there are already 2 other PRs using this as base, which modify `db.ts`. If I split this file, there'll be many conflicts;
- we should investigate using an ORM, as you had suggested. That could also simplify things a lot and make this file much smaller;

src/db.ts
obiyankenobi: Ok, I'll improve the comment.

src/api/wallet.ts
obiyankenobi: I don't like create because it gives me the idea that it's something that will only be done once. As we discussed, in the future, we may remove the wallet from database if it's inactive. And then if it's needed again it'd be "created" a second time. That's why I went with load.

But to be honest, I don't care much about the naming here. We can go with start or create, whatever everyone thinks it's best.

src/api/addresses.ts
```diff
@@ -0,0 +1,48 @@
...
+    walletId = params.id;
+  } else {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.MISSING_PARAMETER, parameter: 'id' }),
+    };
+  }
+
+  const status = await getWallet(mysql, walletId);
```
obiyankenobi, line 35: Need to check if status is ready.

src/api/balances.ts
```diff
@@ -0,0 +1,54 @@
...
+    walletId = params.id;
+  } else {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.MISSING_PARAMETER, parameter: 'id' }),
+    };
+  }
+
+  const status = await getWallet(mysql, walletId);
```
obiyankenobi, line 35: Need to check if status is ready.

src/api/txhistory.ts
```diff
@@ -0,0 +1,90 @@
...
+      return {
+        statusCode: 200,
+        body: JSON.stringify({ success: false, error: ApiError.INVALID_PARAMETER, parameter: 'count' }),
+      };
+    }
+    // we don't return an error if user requests more than the maximum allowed
+    count = Math.min(MAX_COUNT, parsed);
+  }
+
+  const status = await getWallet(mysql, walletId);
```
obiyankenobi, line 69: Need to check if status is ready.

src/api/txhistory.ts
obiyankenobi: Actually, do you mean creating a Wallet class so we later can call `wallet.getTxHistory(tokenId, skip, count)`? I'd just be against saving the `mysql` instance in the wallet, so we're basically just repeating `walletId`. But we could do it, so the other APIs would also use it: `wallet.getBalance(mysql, ...)`, `wallet.getAddresses(mysql, ...)`, etc.

src/api/txhistory.ts
jansegre: Why do we have to get this from the lib? If the value is ever different, it would be a major breaking change, no?

src/api/txhistory.ts
jansegre: I'm OK with the code as is. I'm not big fan of trying to abstract too much when a few simple functions are enough.

DATABASE.md
jansegre: I second that. I'd prefer an `.sql`, even if only for running locally. And also fixing the database engine to MySQL/MariaDB. Because I don't think we're willing to support an engine that we're not even going to use, and significant differences are bound to come up eventually (for example GitLab dropping support for MySQL).

src/utils.ts
```diff
@@ -0,0 +1,116 @@
...
+ * ```
+ *
+ * @param xpubkey - The xpubkey
+ * @param startIndex - Generate addresses starting from this index
+ * @param quantity - Amount of addresses to generate
+ * @returns A list with the generated addresses and corresponding index
+ */
+export const getHathorAddresses = (xpubkey: string, startIndex: number, quantity: number): StringMap<number> => {
+  const addrMap: StringMap<number> = {};
+  const xpub = HDPublicKey(xpubkey);
```
obiyankenobi: Need to treat bad xpubkeys:
```
Invalid Base58 checksum for xpub6EcBoi2vDFcCW5sPAiQpXDYYtXd1mKhUJD64tUi8CPRG1VQFDkAbL8G5gqTmSZD6oq4Yhr5PZ8pKf3Xmb3W3pGcgqzUdFNaCRKL7TZa3rex
Error: 
    at new NodeError (/Users/yan/dev/hathor/wallet-service/node_modules/bitcore-lib/lib/errors/index.js:20:41)
    at Function.HDPublicKey.getSerializedError (/Users/yan/dev/hathor/wallet-service/node_modules/bitcore-lib/lib/hdpublickey.js:235:12)
    at new HDPublicKey (/Users/yan/dev/hathor/wallet-service/node_modules/bitcore-lib/lib/hdpublickey.js:42:31)
    at HDPublicKey (/Users/yan/dev/hathor/wallet-service/node_modules/bitcore-lib/lib/hdpublickey.js:38:12)
```

jansegre (approved):

src/api/addresses.ts
msbrogli: Ok. I need to get used to programming lambda handlers. Thanks.

src/api/addresses.ts
msbrogli: I got it. Thanks!

src/api/addresses.ts
msbrogli: As far as I understood, we need to always call `closeDbConnection` before this method ends. In this case, shouldn't we use a decorator to do it?

Example:
```js
function closeDbConnectionDecorator(wrapped) {
  return function() {
    try {
        return wrapped.apply(this, arguments);
    } finally {
        closeDbConnection(mysql);
    }
  }
}
```

src/api/txhistory.ts
msbrogli: Ok. We can refactor it later if it makes sense.

src/api/wallet.ts
msbrogli: I think so. We can run an insert query and commit. We just need to handle a potential primary key duplication.

src/api/wallet.ts
msbrogli: I got it. Maybe move it to a constant?

src/api/wallet.ts
msbrogli: I feel that "you cannot load something that hasn't been created before". But I understand your points as well. Maybe we should just improve the documentation to explain what you just said.

src/db.ts
```diff
@@ -0,0 +1,627 @@
...
+ * It updates both the walletId and index of given addresses.
+ *
+ * @param mysql - Database connection
+ * @param walletId - The wallet id
+ * @param addresses - A map of addresses and corresponding indexes
+ */
+export const updateExistingAddresses = async (mysql: ServerlessMysql, walletId: string, addresses: AddressIndexMap): Promise<void> => {
+  if (Object.keys(addresses).length === 0) return;
+
+  for (const [address, index] of Object.entries(addresses)) {
```
msbrogli, line 225: What if only one address has been added? Will we update all other addresses of that wallet?

src/db.ts
```diff
@@ -0,0 +1,627 @@
...
+ * Initialize a wallet's balance.
+ *
+ * @remarks
+ * This function adds entries to wallet_balance table, using data from address_balance and address_tx_history.
+ *
+ * @param mysql - Database connection
+ * @param walletId - The wallet id
+ * @param addresses - The addresses that belong to this wallet
+ */
+export const initWalletBalance = async (mysql: ServerlessMysql, walletId: string, addresses: string[]): Promise<void> => {
```
msbrogli, line 274: Maybe we should add a lock per `walletId` here to prevent race conditions.

src/db.ts
```diff
@@ -0,0 +1,627 @@
...
+ * @remarks
+ * When a new transaction arrives, it can change the balance and tx history for the wallets. This function
+ * updates the wallet_balance and wallet_tx_history tables with information from this transaction.
+ *
+ * @param mysql - Database connection
+ * @param txId - Transaction id
+ * @param timestamp - Transaction timestamp
+ * @param walletBalanceMap - Map with the transaction's balance for each wallet (by walletId)
+ */
+export const updateWalletTablesWithTx = async (
```
msbrogli, line 318: Double check if we need to add a lock here.

src/db.ts
```diff
@@ -0,0 +1,627 @@
...
+ */
+export const getUtxosLockedAtHeight = async (
+  mysql: ServerlessMysql,
+  now: number,
+  height: number,
+): Promise<Utxo[]> => {
+  const utxos = [];
+  if (height >= 0) {
+    const results: DbSelectResult = await mysql.query(
+      'SELECT * FROM `utxo` WHERE `heightlock` = ? AND (timelock <= ? OR timelock is NULL)',
```
msbrogli, line 618: Shouldn't it be `heightlock > [height]`?

src/txProcessor.ts
```diff
@@ -0,0 +1,263 @@
...
+  const blockRewardLock = parseInt(process.env.BLOCK_REWARD_LOCK, 10);
+  for (const evt of event.Records) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    await addNewTx(evt.body, now, blockRewardLock);
+  }
+
+  await closeDbConnection(mysql);
+
+  // TODO delete message from queue
```
msbrogli, line 60: What does SQS do to prevent sending the same message to two consumers?

msbrogli (changes requested):

src/db.ts
obiyankenobi: I'm not sure I understood your questions.

This function only updates the index and walletId in the address table. There's no logic here, it's up to the calling function to determine which addresses will be updated. This function is called when creating the wallet, so existing addresses get marked as being part of a wallet.

src/db.ts
obiyankenobi: Adding locks on lambda functions is not easy to do. Anyway, this function will try to do an insert on `wallet_balance`. If it's been already added, it will fail on a repeated primary key. So treating the database error should handle the problem.

As for treating the db error, it's not being done now. There's a task on the epic (#5) to do this and transaction rollback support.

src/db.ts
obiyankenobi: I can review the logic behind it, but the docstring clearly states it only gets the blocks locked at the specific height.

src/txProcessor.ts
obiyankenobi: https://aws.amazon.com/sqs/features/#Queue_types

We should us FIFO queues.

DATABASE.md
obiyankenobi: Included `db.sql`.

README.md
obiyankenobi: Done

fullnode-bridge.js
obiyankenobi: Done

src/api/txhistory.ts
obiyankenobi: Getting it from the lib anyway, since it's easy.

src/api/wallet.ts
obiyankenobi: Added as a constant on this same file, as it's only used here. If we later use in other places, we can move to a constants file.

src/api/wallet.ts
obiyankenobi: Renamed to `createWallet`.

src/api/addresses.ts
obiyankenobi: I created a helper method `closeDbAndGetError`, so now it's easier. But that only sends errors. For success responses, since it's only done once in each API, I kept the more verbose code. This way, the helper function can be specific to errors.

src/api/wallet.ts
obiyankenobi: Done

src/db.ts
obiyankenobi: Done

src/txProcessor.ts
obiyankenobi: Done

src/api/addresses.ts
obiyankenobi: Done

src/api/balances.ts
obiyankenobi: Done

src/api/txhistory.ts
obiyankenobi: Done

src/utils.ts
obiyankenobi: This code was moved to the lib and already handles this.

src/api/addresses.ts
obiyankenobi: I created a helper method to close the connection and get the error reply (`closeDbAndGetError`), so the code is cleaner now. What do you think?

src/db.ts
obiyankenobi: No need to do it, as long as tx is processed once.

src/api/wallet.ts
obiyankenobi: Now I also update the status of the wallet on db to `error` and return `success: false`. Regarding logging, I'm not sure where it would end up when using lambdas. I need to look into this when testing on the real environment.

src/api/wallet.ts
obiyankenobi: Yeah, handling the db error should be fine.

src/db.ts
pedroferreira1: About token and transaction not being useful I agree with this code, however with future implementations of the next features maybe it will (at least transaction). But I agree with you that could be done in a different PR, just create this issue to separate this file because it's already confusing with the current code and might become worse with to next changes.

src/api/wallet.ts
pedroferreira1: Yeah, we can discuss the specific cases when they come to us in the future, it's ok for now.

package.json
```diff
@@ -0,0 +1,56 @@
...
+  "scripts": {
+    "postinstall": "npm dedupe",
+    "jest": "jest --runInBand --collectCoverage",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "check-types": "tsc --noemit",
+    "test": "run-s check-types lint jest",
+    "offline": "./node_modules/.bin/sls offline --printOutput --noPrependStageInUrl"
+  },
+  "author": "Hathor Labs",
+  "license": "MIT",
```
pedroferreira1, line 14: I think we should create the LICENSE file also, right? And also add the header in all code files. Just to follow what we've been using in all other repositories.

pedroferreira1 (approved):

src/api/wallet.ts
msbrogli: That's something really important to check, and also how to monitor this service.

src/txProcessor.ts
msbrogli: Thanks!

src/db.ts
msbrogli: Ok. Thanks!

src/db.ts
msbrogli: Ok! Thanks.

src/api/addresses.ts
msbrogli: It is good. Thanks!

src/db.ts
msbrogli: I'm trying to figure out if it will loop over all addresses of the wallet or not.

msbrogli (approved):

src/db.ts
obiyankenobi: It will loop over the addresses passed on the `addresses` parameter. That, in turn, is calculated on:
`const { addresses, existingAddresses, newAddresses } = await generateAddresses(mysql, xpubkey, maxGap);`

package.json
obiyankenobi: That's true, forgot about it.

src/db.ts
obiyankenobi: There's already a task in the Epic to revisit the db code and support transactions/commits. I think that'll be the best PR to do it.

package.json
obiyankenobi: Done

PR #6, opened 2020-07-20
feat: handle timelock

obiyankenobi:


DATABASE.md
```diff
@@ -32,6 +33,12 @@ CREATE TABLE `address_tx_history` (
   PRIMARY KEY (`address`,`tx_id`,`token_id`)
 );
 
+CREATE TABLE `info` (
```
pedroferreira1: What's this table used for?

tests/utils.ts
```diff
@@ -48,10 +48,11 @@ export const createOutput = (value: number, address: string, token = '00', timel
     },
     token,
     spent_by: null,
+    locked,
   }
 );
 
-export const createInput = (value: number, address: string, txId: string, index: number, token = '00', timelock = null, tokenData = 0): TxInput => (
+export const createInput = (value: number, address: string, txId: string, index: number, token = '00', timelock = null, locked = false, tokenData = 0): TxInput => (
```
pedroferreira1: Why do we have a `locked` field on the input?

tests/utils.ts
```diff
@@ -168,14 +172,15 @@ export const checkAddressBalanceTable = async (
   }
 
   // now fetch the exact entry
+  const baseQuery = 'SELECT * FROM `address_balance` WHERE `address` = ? AND `token_id` = ? AND `unlocked_balance` = ? AND `locked_balance` = ? AND `transactions` = ?';
   results = await mysql.query(
-    'SELECT * FROM `address_balance` WHERE `address` = ? AND `token_id` = ? AND `unlocked_balance` = ? AND `locked_balance` = ? AND `transactions` = ?',
-    [address, tokenId, unlocked, locked, transactions],
+    `${baseQuery} AND timelock_expires ${lockExpires === null ? 'IS' : '='} ?`,
```
pedroferreira1: All other queries the parameters are between ``, shouldn't timelock_expires also be between ``?

src/db.ts
```diff
@@ -333,15 +333,22 @@ export const updateWalletTablesWithTx = async (
...
+      // save the smaller value of timelock_expires, when not null
       await mysql.query(
-        'INSERT INTO `wallet_balance` SET ? ON DUPLICATE KEY UPDATE unlocked_balance = unlocked_balance + ?, locked_balance = locked_balance + ?, transactions = transactions + 1',
-        [entry, tokenBalance.unlocked, tokenBalance.locked],
+        `INSERT INTO wallet_balance SET ? ON DUPLICATE KEY UPDATE unlocked_balance = unlocked_balance + ?, locked_balance = locked_balance + ?,
+        transactions = transactions + 1, timelock_expires = CASE WHEN timelock_expires IS NULL THEN VALUES(timelock_expires)
+                                                                 WHEN VALUES(timelock_expires) IS NULL THEN timelock_expires
+                                                                 ELSE LEAST(timelock_expires, VALUES(timelock_expires))
+                                                            END`,
+        [entry, tokenBalance.unlocked, tokenBalance.locked, walletId, token],
```
pedroferreira1: In this query I see only 3 '?', which I believe are the markers for escaping vars. Why do we have 5 elements on the array?

src/db.ts
```diff
@@ -359,40 +366,92 @@ export const updateWalletTablesWithTx = async (
...
+ * that is sill marked as locked in our database. That might happen if the user sends his transaction
+ * using a service other than this one. Otherwise the locked amount would have been updated before
+ * sending.
+ *
+ * @param mysql - Database connection
+ * @param inputs - The transaction inputs
+ * @returns The locked UTXOs
+ */
+export const getTxLockedInputs = async (mysql: ServerlessMysql, inputs: TxInput[]): Promise<Utxo[]> => {
+  const entries = inputs.map((input) => [input.tx_id, input.index]);
```
pedroferreira1, line 515: I think we should try to be consistent with the names. Here we have `input.tx_id` and right above we have `utxo.txId`

src/db.ts
```diff
@@ -359,40 +366,92 @@ export const updateWalletTablesWithTx = async (
...
+ * when the wallet owner requests its balance. Therefore, we might receive a transaction with a UTXO
+ * that is sill marked as locked in our database. That might happen if the user sends his transaction
+ * using a service other than this one. Otherwise the locked amount would have been updated before
+ * sending.
+ *
+ * @param mysql - Database connection
+ * @param inputs - The transaction inputs
+ * @returns The locked UTXOs
+ */
+export const getTxLockedInputs = async (mysql: ServerlessMysql, inputs: TxInput[]): Promise<Utxo[]> => {
```
pedroferreira1: This method returns the locked utxos not the locked inputs. I think this name is misleading maybe it could be `getLockedUtxoFromInputs` don't know if it's good but I think is better.

Even the first line of the comment is saying "Get tx inputs that are still marked as locked."

src/db.ts
```diff
@@ -435,11 +494,17 @@ export const updateAddressTablesWithTx = async (
...
+      // save the smaller value of timelock_expires, when not null
       await mysql.query(
-        'INSERT INTO `address_balance` SET ? ON DUPLICATE KEY UPDATE unlocked_balance = unlocked_balance + ?, locked_balance = locked_balance + ?, transactions = transactions + 1',
-        [entry, tokenBalance.unlocked, tokenBalance.locked],
+        `INSERT INTO address_balance SET ? ON DUPLICATE KEY UPDATE unlocked_balance = unlocked_balance + ?, locked_balance = locked_balance + ?,
+        transactions = transactions + 1, timelock_expires = CASE WHEN timelock_expires IS NULL THEN VALUES(timelock_expires)
+                                                                 WHEN VALUES(timelock_expires) IS NULL THEN timelock_expires
+                                                                 ELSE LEAST(timelock_expires, VALUES(timelock_expires))
+                                                            END`,
+        [entry, tokenBalance.unlocked, tokenBalance.locked, address, token],
```
pedroferreira1: Again the same comment about the  5 parameters with 3 '?'. Only to mark if there is a problem in the other, this should be changed also.

src/db.ts
```diff
@@ -461,17 +526,27 @@ export const updateAddressTablesWithTx = async (
...
     for (const [token, tokenBalance] of tokenBalanceMap.iterator()) {
       await mysql.query(
         'UPDATE `address_balance` SET `unlocked_balance` = `unlocked_balance` + ?, `locked_balance` = `locked_balance` - ? WHERE `address` = ? AND `token_id` = ?',
         [tokenBalance.unlocked, tokenBalance.unlocked, address, token],
       );
+
+      // if this is being unlocked due to a timelock, also update the timelock_expires column
+      if (updateTimelocks) {
+        await mysql.query(
+          'UPDATE `address_balance` SET `timelock_expires` = (SELECT MIN(`timelock`) FROM `utxo` WHERE `address` = ? AND `token_id` = ? AND `locked` = TRUE)',
```
pedroferreira1: When the utxo is locked because of a heightlock and not a timelock, we shouldn't consider it on the SELECT, right? 

Besides that, if we have no more locked utxos should we also update the `timelock_expires` to `null`?

src/db.ts
```diff
@@ -485,17 +560,27 @@ export const updateAddressLockedBalance = async (
...
     for (const [token, tokenBalance] of tokenBalanceMap.iterator()) {
       await mysql.query(
         'UPDATE `wallet_balance` SET `unlocked_balance` = `unlocked_balance` + ?, `locked_balance` = `locked_balance` - ? WHERE `wallet_id` = ? AND `token_id` = ?',
         [tokenBalance.unlocked, tokenBalance.unlocked, walletId, token],
       );
+
+      // if this is being unlocked due to a timelock, also update the timelock_expires column
+      if (updateTimelocks) {
+        await mysql.query(
+          'UPDATE `wallet_balance` SET `timelock_expires` = (SELECT MIN(`timelock_expires`) FROM `address_balance` WHERE `address` IN (SELECT `address` from `address` WHERE `wallet_id` = ?) AND `token_id` = ?)',
```
pedroferreira1: What happens with the `MIN` operator with `null` values and integer values?

DATABASE.md
pedroferreira1: For what I've seen is used to save the current height, right? Or something else also?
Maybe we should update the Design to add it, I tried to find there and that's why I added this first comment.

pedroferreira1 (changes requested):

DATABASE.md
jansegre: Yeah, I also think the name is too generic. Would `lock_info` or `metadata` describe it better? I'm not really sure what it's used for.

src/db.ts
jansegre: I agree with consistency. But I get that these differences may come from exposing (or consuming) an API that uses a different naming pattern.

jansegre (approved):

DATABASE.md
andreabadesso: I believe the idea is to have a generic key-value store in the database to persist data

For now the only info stored is the current height

I agree that `metadata` would describe it better though

src/db.ts
andreabadesso: This was probably left there after the query was refactored

Removed it.

src/db.ts
andreabadesso: Agreed, replaced `getTxLockedInputs` with `getLockedUtxoFromInputs`

src/db.ts
andreabadesso: I also agree with the consistency but as there are several places mixing camelCase and snake_case, I suggest creating new issue to deal with this normalization and convert everything to camelCase as that is the usual casing for javascript

src/db.ts
andreabadesso: Fixed

src/db.ts
andreabadesso: It will return the MIN integer value if there is at least one non-null value and NULL otherwise

tests/utils.ts
andreabadesso: Maybe the idea was to have locked here to test the condition we have on the `getLockedUtxoFromInputs` method (```We have a "lazy" unlock policy so we might receive a transaction with a UTXO that is still marked as locked in our database```)

But since it is not being used anywhere on the tests, I removed it.

DATABASE.md
andreabadesso: Updated on ec887d6c606aeadb3e03b90ed86288dcb89e5745

tests/utils.ts
andreabadesso: The `value` of `timelock_expires` is using `?`, only the operation is using js template strings, so we can use `IS NULL` instead of `= NULL`

This will be refactored when we add an ORM or a QueryBuilder, I suggest not changing it on this PR

src/db.ts
obiyankenobi: I used camelCase for this code, but the tx object we get from the fullnode uses snake_case.

src/db.ts
andreabadesso: > When the utxo is locked because of a heightlock and not a timelock, we shouldn't consider it on the SELECT, right?

This function (`updateAddressLockedBalance`) is called on two situations:
1) When a new block arrives, so it will unlock UTXOs from height-locked blocks. In this case, `updateTimeLocks` will be set to false
2) When there are locked inputs (queried from `getLockedUtxoFromInputs`)

Both calls are made from [`addNewTx`](https://github.com/HathorNetwork/hathor-wallet-service/blob/ec887d6c606aeadb3e03b90ed86288dcb89e5745/src/txProcessor.ts#L83)

On case (1), we would also get height-locked inputs because we use the same `locked` column for locked inputs (wether they are locked because of time, height or being used in a tx-proposal). But that shouldn't matter because the query is only updating `address_balance`, so height-blocked inputs with no timelock would receive `NULL`.

That's the query returning the closest timelock expire:

`SELECT MIN(timelock) FROM utxo WHERE address = ? AND token_id = ? AND locked = TRUE`

There are two possible outcomes of that:

1) There is at least one "valid" timelock in any of the UTXOs, so it will return the closest.
2) There are no more valid timelocks, so it will return `NULL`

> Besides that, if we have no more locked utxos should we also update the timelock_expires to null?

That's exactly what the outcome 2 is doing.

obiyankenobi (commented):
Looks good to me.

pedroferreira1 (approved):

jansegre (approved):

src/api/balances.ts
```diff
@@ -46,7 +50,21 @@ export const get: APIGatewayProxyHandler = async (event) => {
...
+  const now = getUnixTimestamp();
+  const refreshBalances = balances.some((tb) => {
+    if (tb.balance.lockExpires && tb.balance.lockExpires <= now) {
+      return true;
+    }
+    return false;
+  });
+
+  if (refreshBalances) {
+    await updateBalances(mysql, walletId, now);
```
msbrogli, line 65: Can't we just update the balance of the `tb`s which lock has expired (instead of updating the whole wallet balance)?

src/api/balances.ts
```diff
@@ -55,3 +73,9 @@ export const get: APIGatewayProxyHandler = async (event) => {
     body: JSON.stringify({ success: true, balances }),
   };
 };
+
+const updateBalances = async (_mysql: ServerlessMysql, walletId: string, now: number) => {
```
msbrogli, line 84: Missing docstring.

src/db.ts
```diff
@@ -615,7 +833,12 @@ export const getUtxosLockedAtHeight = async (
   const utxos = [];
   if (height >= 0) {
     const results: DbSelectResult = await mysql.query(
-      'SELECT * FROM `utxo` WHERE `heightlock` = ? AND (timelock <= ? OR timelock is NULL)',
+      `SELECT *
+         FROM \`utxo\`
+        WHERE \`heightlock\` <= ?
+          AND (\`timelock\` <= ?
+               OR \`timelock\` is NULL)
+          AND \`locked\` = TRUE`,
```
msbrogli: I know mysql accepts `TRUE` and `FALSE` but isn't it better to use `1` here? I'd say it will clearly show that `locked` is a numeric field and justify the `result.locked > 0` below.

src/db.ts
```diff
@@ -627,9 +850,89 @@ export const getUtxosLockedAtHeight = async (
...
+): Promise<Utxo[]> => {
+  const utxos = [];
+  const results: DbSelectResult = await mysql.query(
+    `SELECT *
+       FROM \`utxo\`
+      WHERE (\`heightlock\` <= ?
+             OR \`heightlock\` is NULL)
+        AND (\`timelock\` <= ?
+             OR \`timelock\` is NULL)
+        AND \`locked\` = TRUE
```
msbrogli: Same here. Replace `TRUE` by `1`.

src/types.ts
```diff
@@ -65,18 +68,29 @@ export class Balance {
...
    *
    * @param b1 - First balance
    * @param b2 - Second balance
    * @returns The sum of both balances
    */
-  static sum(b1: Balance, b2: Balance): Balance {
-    return new Balance(b1.unlocked + b2.unlocked, b1.locked + b2.locked);
+  static merge(b1: Balance, b2: Balance): Balance {
+    let lockExpires = null;
+    if (b1.lockExpires === null) {
```
msbrogli, line 86: Isn't it easier to use `0` or `-1` instead of `null`? I feel it would prevent a lot of verifications.

tests/api.test.ts
```diff
@@ -115,8 +117,9 @@ test('GET /balances', async () => {
   expect(returnBody.balances).toHaveLength(0);
 
   // add 2 balances
-  await addToWalletBalanceTable(mysql, [['my-wallet', 'token1', 10, 0, 3]]);
-  await addToWalletBalanceTable(mysql, [['my-wallet', 'token2', 3, 2, 1]]);
+  const lockExpires = getUnixTimestamp() + 200;
+  await addToWalletBalanceTable(mysql, [['my-wallet', 'token1', 10, 0, null, 3]]);
```
msbrogli: I feel the method should receive an object and internally map it to an array. The way it is seems very error prone. What do you think?

tests/db.test.ts
```diff
@@ -340,65 +345,114 @@ test('updateWalletTablesWithTx', async () => {
   const ts2 = 20;
   const ts3 = 30;
 
+  await addToAddressTable(mysql, [
+    ['addr1', 0, walletId, 1],
```
msbrogli: Don't you think this is hard to read (and too error prone when one is using it)?

msbrogli (changes requested):

src/types.ts
andreabadesso: It would indeed prevent some verifications but I would argue that using null is more "semantically correct", meaning that it does not have an expiration time as oposed to "it does have an expiration time but it already elapsed". What do you think?

Also, it will come as `null` from the database

src/api/balances.ts
andreabadesso: I don't know if I understood your question correctly

`updateBalances` only updates the UTXOs that are currently locked and should be unlocked on the current height

tests/api.test.ts
andreabadesso: I agree. Refactored it

tests/db.test.ts
andreabadesso: I agree. Refactored it

src/types.ts
msbrogli: Makes sense. I agree. Thanks!

src/api/balances.ts
msbrogli: Ok. I got it.  But why are we verifying above whether there's any `tb` to be refreshed or not? Isn't it as simple as calling `updateBalances()`? Or add an option so the `getWalletBalances` also refresh locks?

msbrogli (approved):

tests/api.test.ts
```diff
@@ -115,8 +120,23 @@ test('GET /balances', async () => {
   expect(returnBody.balances).toHaveLength(0);
 
   // add 2 balances
-  await addToWalletBalanceTable(mysql, [['my-wallet', 'token1', 10, 0, 3]]);
-  await addToWalletBalanceTable(mysql, [['my-wallet', 'token2', 3, 2, 1]]);
+  const lockExpires = getUnixTimestamp() + 200;
+  await addToWalletBalanceTable(mysql, [{
```
msbrogli, line 124: image

msbrogli (approved):

PR #7, opened 2020-08-03
feat: handle authorities

obiyankenobi:


src/txProcessor.ts
```diff
@@ -99,6 +100,11 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
     await maybeUpdateLatestHeight(mysql, tx.height);
   }
 
+  if (tx.version === 2) {
+    // TODO get version from lib constants
```
obiyankenobi: You can complete the TODO now, as on the other `if`.

src/txProcessor.ts
```diff
@@ -259,14 +264,16 @@ export const unlockUtxos = async (_mysql: ServerlessMysql, utxos: Utxo[], update
...
     return {
-      value: utxo.value,
+      value: utxo.authorities > 0 ? utxo.authorities : utxo.value,
       token: utxo.tokenId,
       decoded,
       locked: false,
-      // we don't care about spent_by, token_data and script
+      // set authority bit if necessary
+      // TODO get TOKEN_AUTHORITY_MASK from lib
+      token_data: utxo.authorities > 0 ? 0b10000000 : 0,
```
obiyankenobi: Get from lib: `hathorLib.constants.TOKEN_AUTHORITY_MASK`

src/types.ts
```diff
@@ -40,35 +40,204 @@ export interface AddressInfo {
...
+
+  symbol: string;
+
+  constructor(id: string, name: string, symbol: string) {
+    this.id = id;
+    this.name = name;
+    this.symbol = symbol;
+    // TODO get HTR info from lib
+    // XXX should we force HTR info for token '00'?
+    if (this.id === '00') {
```
obiyankenobi: TODO

src/types.ts
```diff
@@ -193,14 +383,17 @@ export class TokenBalanceMap {
    * @returns The TokenBalanceMap object
    */
   static fromTxOutput(output: TxOutput): TokenBalanceMap {
-    // TODO handle authority
     // TODO check if output.decoded exists, else return null
     const token = output.token;
     const value = output.value;
+    const isAuthority = (output.token_data & 0b10000000) > 0;   // eslint-disable-line no-bitwise
```
obiyankenobi: Get `0b10000000` from lib.

src/types.ts
```diff
@@ -217,13 +410,17 @@ export class TokenBalanceMap {
...
-    // TODO handle authority
-    // TODO check if output.decoded exists, else return null
     const token = input.token;
-    const value = -input.value;
-
     const obj = new TokenBalanceMap();
-    obj.set(token, new Balance(value, 0, null));
+
+    // TODO get token mask from lib constants
+    if ((input.token_data & 0b10000000) > 0) {    // eslint-disable-line no-bitwise
```
obiyankenobi: Get `0b10000000` from lib.

src/utils.ts
```diff
@@ -76,3 +76,7 @@ export const closeDbConnection = async (mysql: ServerlessMysql): Promise<void> =
     await mysql.end();
   }
 };
+
+export const isAuthority = (tokenData: number): boolean => (
+  (tokenData & 0b10000000) > 0    // eslint-disable-line no-bitwise
```
obiyankenobi: Get `0b10000000` from lib.

src/types.ts
```diff
@@ -40,35 +40,204 @@ export interface AddressInfo {
...
+  static merge(a1: Authorities, a2: Authorities): Authorities {
+    return new Authorities(a1.array.map((value, index) => Math.sign(value + a2.array[index])));
+  }
+
+  toJSON(): Record<string, unknown> {
+    const authorities = this.toInteger();
+    return {
+      // TODO get from lib
+      mint: (authorities & 0b00000001) > 0,   // eslint-disable-line no-bitwise
+      melt: (authorities & 0b00000010) > 0,   // eslint-disable-line no-bitwise
```
obiyankenobi: Get constants from lib.

DATABASE.md
```diff
@@ -39,18 +39,20 @@ CREATE TABLE `metadata` (
   PRIMARY KEY (`key`)
 );
 
-CREATE TABLE `blocks` (
-  `tx_id` VARCHAR(64) NOT NULL,
-  `height` INT UNSIGNED NOT NULL,
-  PRIMARY KEY (`tx_id`)
-);
+// TODO name and symbol lengths are not limited on the blockchain, might be 255
```
pedroferreira1: Actually we do limit both on the protocol level. Name is at most 30 chars and symbol 5 chars.

src/types.ts
pedroferreira1: We have this comparison 4 times in the code, maybe we should create a method for that, e.g. `isTokenDataAuthority`

tests/txProcessor.test.ts
```diff
@@ -382,3 +425,41 @@ test('txProcessor', async () => {
...
+
+  // receive event
+  const evt = JSON.parse(JSON.stringify(eventTemplate));
+  evt.Records[0].body = tokenCreationTx;
+  await txProcessor.onNewTxEvent(evt);
+
+  for (const [index, output] of tokenCreationTx.outputs.entries()) {
+    let value = output.value;
+    let authorities = 0;
+    if ((output.token_data & 0b10000000) > 0) {     // eslint-disable-line no-bitwise
```
pedroferreira1: Get 0b10000000 from lib.

src/types.ts
pedroferreira1: I just saw the method in `src/utils`, can you use it here and in the other places?

src/db.ts
```diff
@@ -435,24 +461,38 @@ export const addUtxos = async (
   txId: string,
   outputs: TxOutput[],
   heightlock: number = null): Promise<void> => {
+  // outputs might be empty if we're destroying authorities or melting tokens
```
pedroferreira1: If you are melting tokens you still have the HTR deposit back as an output, right?

pedroferreira1 (approved):

src/types.ts
andreabadesso: Changed it to get hathor data from the lib, also changed the next lines to use it:

```
    const hathorConfig = hathorLib.constants.HATHOR_TOKEN_CONFIG;

    if (this.id === hathorConfig.uid) {
      this.name = hathorConfig.name;
      this.symbol = hathorConfig.symbol;
    }
```

src/types.ts
andreabadesso: Done

DATABASE.md
andreabadesso: Updated docs to consider 30 chars for the name and 5 chars for the symbol

src/db.ts
andreabadesso: Right, but might be empty if we're destroying authorities. Removed the last part of the comment

.env
```diff
@@ -0,0 +1,7 @@
+STAGE=production
+MAX_ADDRESS_GAP=10
+BLOCK_REWARD_LOCK=300
+DB_ENDPOINT=localhost
+DB_NAME=hathor
+DB_USER=root
+DB_PASS=hathor
```
obiyankenobi: Not sure if you should commit this file.

src/types.ts
```diff
@@ -217,13 +417,16 @@ export class TokenBalanceMap {
...
   static fromTxInput(input: TxInput): TokenBalanceMap {
-    // TODO handle authority
-    // TODO check if output.decoded exists, else return null
     const token = input.token;
-    const value = -input.value;
-
     const obj = new TokenBalanceMap();
-    obj.set(token, new Balance(value, 0, null));
+
+    if (isAuthority(input.token_data)) {    // eslint-disable-line no-bitwise
```
obiyankenobi: `// eslint-disable-line no-bitwise` left

src/db.ts
```diff
@@ -384,8 +387,10 @@ export const updateWalletTablesWithTx = async (
       const entry = {
         wallet_id: walletId,
         token_id: token,
-        unlocked_balance: (tokenBalance.unlocked < 0 ? 0 : tokenBalance.unlocked),
-        locked_balance: (tokenBalance.locked < 0 ? 0 : tokenBalance.locked),
+        unlocked_balance: (tokenBalance.unlockedAmount < 0 ? 0 : tokenBalance.unlockedAmount),
+        locked_balance: tokenBalance.lockedAmount,
+        unlocked_authorities: Math.abs(tokenBalance.unlockedAuthorities.toInteger()),
+        locked_authorities: Math.abs(tokenBalance.lockedAuthorities.toInteger()),
```
obiyankenobi: I'm trying to remember why `Math.abs` is needed here, as `toInteger` should always return a positive number. Did you try removing it? We should investigate and leave a comment so we don't have this question in the furute.

src/db.ts
obiyankenobi: Comment still has `or melting tokens`.

DATABASE.md
```diff
@@ -19,6 +19,8 @@ CREATE TABLE `address_balance` (
   `token_id` varchar(64) NOT NULL,
   `unlocked_balance` bigint unsigned NOT NULL,
   `locked_balance` bigint unsigned NOT NULL,
+  `unlocked_authorities` tinyint unsigned NOT NULL DEFAULT '0',
```
msbrogli, line 31: Why don't we treat authorities like any other utxo? We can add some flags to make it easy to filter them.

src/db.ts
msbrogli: I agree. It feels weird to have `Math.abs()` here. We could even assert that these values are `>= 0`.

src/db.ts
```diff
@@ -581,9 +622,12 @@ export const updateAddressTablesWithTx = async (
       const entry = {
         address,
         token_id: token,
-        // if they're < 0, there must be an entry already, so it will execute "ON DUPLICATE KEY UPDATE" instead of setting it to 0
-        unlocked_balance: (tokenBalance.unlocked < 0 ? 0 : tokenBalance.unlocked),
-        locked_balance: (tokenBalance.locked < 0 ? 0 : tokenBalance.locked),
+        // if it's < 0, there must be an entry already, so it will execute "ON DUPLICATE KEY UPDATE" instead of setting it to 0
```
msbrogli, line 630: In which cases it can become negative? I feel that it should never happen.

src/db.ts
```diff
@@ -581,9 +622,12 @@ export const updateAddressTablesWithTx = async (
...
+        unlocked_balance: (tokenBalance.unlockedAmount < 0 ? 0 : tokenBalance.unlockedAmount),
+        // this is never less than 0, as locked balance only changes when a tx is unlocked
+        locked_balance: tokenBalance.lockedAmount,
+        unlocked_authorities: Math.abs(tokenBalance.unlockedAuthorities.toInteger()),
```
msbrogli: Why do we need the `Math.abs()` here? It also feels weird to call the `.toInteger()` method.

src/db.ts
```diff
@@ -581,9 +622,12 @@ export const updateAddressTablesWithTx = async (
...
+        locked_authorities: Math.abs(tokenBalance.lockedAuthorities.toInteger()),
```
msbrogli: Same here.

src/db.ts
```diff
@@ -435,24 +461,38 @@ export const addUtxos = async (
...
+    (output, index) => {
+      let authorities = 0;
+      let value = output.value;
+
+      if (isAuthority(output.token_data)) {
+        authorities = value;
+        value = 0;
+      }
+
+      return [
```
msbrogli, line 482: I'm always concerned about arrays. I feel they should always be objects. It also make it impossible to the linter to find typing errors/warnings.

src/db.ts
```diff
@@ -599,16 +643,44 @@ export const updateAddressTablesWithTx = async (
...
+                                   unlocked_authorities = (unlocked_authorities | VALUES(unlocked_authorities)),
+                                   locked_authorities = locked_authorities | VALUES(locked_authorities)`,
+        [entry, tokenBalance.unlockedAmount, tokenBalance.lockedAmount, address, token],
       );
 
+      // if we're removing any of the authorities, we need to refresh the authority columns. Unlike the values,
+      // we cannot only sum/subtract, as authorities are binary: you have it or you don't. We might be spending
+      // an authority output in this tx without creating a new one, but it doesn't mean this address does not
+      // have this authority anymore, as it might have other authority outputs
+      if (tokenBalance.unlockedAuthorities.hasNegativeValue()) {
```
msbrogli, line 661: When is it expected to be negative?

src/types.ts
```diff
@@ -40,35 +43,205 @@ export interface AddressInfo {
...
+      id: this.id,
+      name: this.name,
+      symbol: this.symbol,
+    };
+  }
+}
+
+export class Authorities {
+  /**
+   * Supporting up to 8 authorities (but we only have mint and melt at the moment)
```
msbrogli, line 95: I don't get. Is it supposed to handle 8 types of authority but implements only 2? Shouldn't we raise an exception in the constructor if we get a not implemented type?

src/types.ts
```diff
@@ -40,35 +43,205 @@ export interface AddressInfo {
...
+   * Uses the array to calculate the final number. Examples:
+   * [0, 0, 0, 0, 1, 1, 0, 1] = 0b00001101 = 13
+   * [0, 0, 1, 0, 0, 0, 0, 1] = 0b00100001 = 33
+   *
+   * @returns The integer representation
+   */
+  toInteger(): number {
+    let n = 0;
+    for (let i = 0; i < this.array.length; i++) {
+      n += this.array[i] * (2 ** (this.array.length - i - 1));
```
msbrogli, line 127: It's nit-picky but we can shift bits to prevent calculating `2**x` multiple times. We can also prevent the multiplication using an `if` block.

src/types.ts
```diff
@@ -40,35 +43,205 @@ export interface AddressInfo {
...
+   * Return a new object inverting each authority value sign.
+   *
+   * @remarks
+   * If value is set to 1, it becomes -1 and vice versa. Value 0 remains unchanged.
+   *
+   * @returns A new Authority object with the values inverted
+   */
+  toNegative(): Authorities {
+    const finalAuthorities = this.array.map((value) => {
+      if (value === 0) return 0;
```
msbrogli, line 154: Why do we need the `if`?

tests/api.test.ts
```diff
@@ -99,6 +100,18 @@ test('GET /balances', async () => {
 
   await addToWalletTable(mysql, [['my-wallet', 'xpubkey', 'ready', 5, 10000, 10001]]);
 
+  // add tokens
+  const token1 = { id: 'token1', name: 'MyToken1', symbol: 'MT1' };
+  const token2 = { id: 'token2', name: 'MyToken2', symbol: 'MT2' };
+  const token3 = { id: 'token3', name: 'MyToken3', symbol: 'MT3' };
+  const token4 = { id: 'token4', name: 'MyToken4', symbol: 'MT4' };
+  await addToTokenTable(mysql, [
+    [token1.id, token1.name, token1.symbol],
```
msbrogli: Same feelings about using an array instead of an object.

tests/api.test.ts
```diff
@@ -138,36 +165,70 @@ test('GET /balances', async () => {
...
   expect(result.statusCode).toBe(200);
   expect(returnBody.success).toBe(true);
   expect(returnBody.balances).toHaveLength(1);
-  expect(returnBody.balances).toContainEqual({ tokenId: 'token3', transactions: 2, balance: { unlocked: 6, locked: 0, lockExpires: null } });
+  expect(returnBody.balances).toContainEqual({
+    token: token3,
+    transactions: 2,
+    balance: { unlocked: 6, locked: 0 },
+    lockExpires: null,
+    tokenAuthorities: { unlocked: { mint: false, melt: false }, locked: { mint: false, melt: false } },
```
msbrogli, line 225: That's great and easy to read. Cheers to objects! \o/

tests/utils.ts
```diff
@@ -488,6 +501,16 @@ export const addToAddressTxHistoryTable = async (
   [entries]);
 };
 
+export const addToTokenTable = async (
+  mysql: ServerlessMysql,
+  entries: unknown[][],
```
msbrogli: I know it is an utility function for the tests but I think we should receive a list of objects and map internally to an array.

msbrogli (changes requested):

src/db.ts
andreabadesso: I agree. I will refactor this

src/types.ts
andreabadesso: I agree, I'll add the exception

src/types.ts
andreabadesso: Because javascript uses the [IEEE_754](https://en.wikipedia.org/wiki/IEEE_754) standard and has negative and positive zeros, so `(-1) * 0` would return -0.

Apparently `-0 === 0` is `true` on most cases, so there wouldn't be a problem, but I would argue that the if is needed to be safe

tests/api.test.ts
andreabadesso: This was refactored on 0ff2f1bf5d74b28143db2eb6c3fa14516cd88cc1 from `feat/timelock` branch. Will come when we rebase

src/db.ts
andreabadesso: `toInteger` will return negative values if we are spending an authority on the tx, the way the service is designed, we are setting the authority bits to `-1` when dealing with balances of a transaction.

In this case, toInteger will return a negative value, e.g: new `Authorities([-1, -1])` will return -3.

This is expected behaviour with the current design (using "negative bits" to represent spent authority in a tx) so the `Math.abs` is required

What I suggest is creating a `toUnsignedInteger` method that will only return positive values by using Math.abs before returning. What do you think, @obiyankenobi @msbrogli ?

I also got a little confused when trying to understand what was going on on this method so I added a comment on 2675404e53233fdc07589652cd3222c432b20f0e to try to explain better what is doing.

msbrogli: I agree with the `toUnsignedInteger()` method. Anyway, I still don't understand why we have negative values (and why we have decided to use negative values to represent a specific situation).

src/types.ts
msbrogli: WTF! Really?! Ok. I feel we should add a comment explaining why this code is here (and mentioning the IEEE 754 for float numbers). Thanks!

msbrogli (changes requested):

DATABASE.md
andreabadesso: We could and probably should because of Pay-to-ScriptHash (P2SH) transactions.

If the Authority UTXO is inside a P2SH transaction, we won't be able to know if it is "unlocked" until we verify its script, making this cache invalid.

I would suggest to open a new issue to refactor the way we store and cache authorities instead of doing in this PR tho, what do you think?

src/db.ts
andreabadesso: Will use the new `.toUnsignedInteger()`

src/db.ts
andreabadesso: Will use the new `.toUnsignedInteger()`

DATABASE.md
msbrogli: I agree. Let's do it in the future. Would you please add another issue about it to add it to our backlog?

src/types.ts
andreabadesso: Done bbec6f667ed38c3213912ae7e204964e49570d47

tests/utils.ts
andreabadesso: Done in 98315b43356214ddce32c1c4367099e7eff98cf9

DATABASE.md
andreabadesso: Done. https://github.com/HathorNetwork/hathor-wallet-service/issues/15

src/types.ts
andreabadesso: I don't think we can shift bits because we could have negative values in the array (representing an authority being spent on a TX)

I agree on the if block, I will add a `if (this.array[i] === 0) continue` because we may be dealing with negative numbers

src/db.ts
andreabadesso: We can have negative values for an Authority when calculating balances of a transaction.

When calculating the tokenBalanceMap from the Inputs, we set the authorities to negative when they are being spent on the inputs as you can see on 

https://github.com/HathorNetwork/hathor-wallet-service/blob/a42578f9458a2bb51c6b73caac125412a7bbbddc/src/types.ts#L435

So when the balances for an address are **merged**, we get the correct authorities value:

https://github.com/HathorNetwork/hathor-wallet-service/blob/a42578f9458a2bb51c6b73caac125412a7bbbddc/src/txProcessor.ts#L187

This is done by using [Math.sign](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) on the merge function:

https://github.com/HathorNetwork/hathor-wallet-service/blob/a42578f9458a2bb51c6b73caac125412a7bbbddc/src/types.ts#L196

andreabadesso: Added the `toUnsignedInteger` method

src/db.ts
andreabadesso: This method is dealing with balances of a transaction, so `tokenBalance.unlocked` can be negative when we are spending inputs

To make it clearer (it gets kind of hard to understand because we are using `ON DUPLICATE`), the object `entry` will only be inserted if there are not already an address_balance row for this (address, token_id)

This will be clearer when we refactor with the ORM

src/types.ts
msbrogli: Following the docstring, I understood that the arrays must have only zeros and ones. What other values are allowed? I mean, the algorithm to convert to integer might be wrong for other values. For example, what is the expected result for `[-1, -1, -1]`?

msbrogli (commented):
Just replied some comments. It seems there are some things yet to finish.

src/db.ts
andreabadesso: @msbrogli 

This is the expected syntax for bulk inserts on the ServerlessMysql lib we are using, this will be refactored and clearer when we add the ORM

src/types.ts
andreabadesso: @msbrogli this will be removed or refactored when we close https://github.com/HathorNetwork/hathor-wallet-service/issues/15

src/db.ts
andreabadesso: @msbrogli this will be removed or refactored when we close https://github.com/HathorNetwork/hathor-wallet-service/issues/15

DATABASE.md
pedroferreira1: You need to remove this TODO, right?

pedroferreira1 (approved):

msbrogli (approved):

DATABASE.md
andreabadesso: Done

PR #9, opened 2020-09-02
feat: tx proposal

obiyankenobi:
I'm having some problems with `bitcore-lib` and `hathor-wallet-lib`. It'll be easier if I make some changes to the lib so I don't have to generate the addresses in this service and won't need to import `bitcore-lib` directly.

Since this code (address generation) was introduced on #4 , I'll go back to it and start making all the changes suggested. When that is done, I can come back to this PR to finish what's left. But in theory, most of the work is done and the tx is assembled and ready to be sent.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,77 +1,34 @@
 {
   "name": "hathor-wallet-service",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "version": "0.0.1",
-      "hasInstallScript": true,
-      "license": "MIT",
-      "dependencies": {
-        "@hathor/wallet-lib": "^0.16.5",
-        "aws-lambda": "^1.0.6",
-        "aws-sdk": "^2.700.0",
-        "bitcore-lib": "^0.16.0",
-        "joi": "^17.3.0",
-        "mysql": "^2.18.1",
-        "serverless-mysql": "^1.5.4",
-        "source-map-support": "^0.5.19",
-        "uuid": "^8.3.0"
-      },
-      "devDependencies": {
-        "@types/aws-lambda": "^8.10.17",
-        "@types/jest": "^26.0.0",
-        "@types/node": "^10.12.18",
-        "@typescript-eslint/eslint-plugin": "^3.3.0",
-        "@typescript-eslint/parser": "^3.3.0",
-        "dotenv": "^8.2.0",
-        "eslint": "^7.2.0",
-        "eslint-config-airbnb-base": "^14.2.0",
-        "eslint-import-resolver-alias": "^1.1.2",
-        "eslint-plugin-import": "^2.21.2",
-        "eslint-plugin-jest": "^23.13.2",
-        "eslint-plugin-module-resolver": "^0.16.0",
-        "fork-ts-checker-webpack-plugin": "^3.0.1",
-        "jest": "^26.0.1",
-        "npm-run-all": "^4.1.5",
-        "serverless": "^1.73.1",
-        "serverless-dotenv-plugin": "^2.4.2",
-        "serverless-offline": "^6.4.0",
-        "serverless-prune-plugin": "^1.4.3",
-        "serverless-webpack": "^5.2.0",
-        "ts-jest": "^26.1.1",
-        "ts-loader": "^5.3.3",
-        "typescript": "^3.9.5",
-        "typescript-eslint": "0.0.1-alpha.0",
-        "webpack": "^4.29.0",
-        "webpack-node-externals": "^1.7.2"
+  "dependencies": {
+    "2-thenable": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/2-thenable/-/2-thenable-1.0.0.tgz",
+      "integrity": "sha512-HqiDzaLDFCXkcCO/SwoyhRwqYtINFHF7t9BDRq4x90TOKNAJpiqUt9X5lQ08bwxYzc067HUywDjGySpebHcUpw==",
+      "dev": true,
+      "requires": {
+        "d": "1",
+        "es5-ext": "^0.10.47"
       }
     },
-    "node_modules/@babel/code-frame": {
+    "@babel/code-frame": {
       "version": "7.10.1",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.1.tgz",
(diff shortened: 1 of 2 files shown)


src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,445 @@
...
+  await addTxProposalOutputs(mysql, txProposalId, finalOutputs);
+
+  await closeDbConnection(mysql);
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({
+      success: true,
+      txProposalId,
+      inputs: inputUtxos.map((utxo) => ({ txId: utxo.txId, index: utxo.index })),
```
obiyankenobi: Also need to return the tokens array, as it's needed to assemble the tx for mining.

.env.production
```diff
@@ -1,3 +1,4 @@
 STAGE=production
-MAX_ADDRESS_GAP=20
+MAX_ADDRESS_GAP=10
```
msbrogli: I feel the default value should be `20`. Any reason to change it?

.gitignore
```diff
@@ -12,3 +12,5 @@ coverage
 
 # eslint
 .eslintcache
+
+.env
```
msbrogli: You removed the `.env` file. Shouldn't it be `.env.*`?

DATABASE.md
```diff
@@ -35,6 +42,21 @@ CREATE TABLE `address_tx_history` (
   PRIMARY KEY (`address`,`tx_id`,`token_id`)
 );
 
+CREATE TABLE `version_data` (
```
msbrogli: Why don't we store it in memory? I feel we should get it from the fullnode when the service starts.

DATABASE.md
```diff
@@ -48,19 +70,53 @@ CREATE TABLE `token` (
...
+  `id` varchar(36) NOT NULL,
+  `wallet_id` varchar(64) NOT NULL,
+  `status` enum('open','sent','send_error','cancelled') NOT NULL,
+  `created_at` int unsigned NOT NULL,
+  `updated_at` int unsigned DEFAULT NULL,
+  PRIMARY KEY (`id`)
+);
+
+
+CREATE TABLE `tx_proposal` (
```
msbrogli: Duplicate?

DATABASE.md
```diff
@@ -48,19 +70,53 @@ CREATE TABLE `token` (
   `symbol` varchar(5) NOT NULL,
   PRIMARY KEY (`id`));
 
+CREATE TABLE `tx_proposal` (
+  `id` varchar(36) NOT NULL,
+  `wallet_id` varchar(64) NOT NULL,
+  `status` enum('open','sent','send_error','cancelled') NOT NULL,
+  `created_at` int unsigned NOT NULL,
+  `updated_at` int unsigned DEFAULT NULL,
+  PRIMARY KEY (`id`)
```
msbrogli: Maybe we should index `wallet_id` for fast queries.

DATABASE.md
```diff
@@ -48,19 +70,53 @@ CREATE TABLE `token` (
...
+
+
+CREATE TABLE `tx_proposal_outputs` (
+  `tx_proposal_id` varchar(36) NOT NULL,
+  `index` tinyint unsigned NOT NULL,
+  `address` varchar(34) NOT NULL,
+  `token_id` varchar(64) NOT NULL,
+  `value` bigint DEFAULT NULL,
+  `timelock` int unsigned DEFAULT NULL,
+  PRIMARY KEY (`tx_proposal_id`,`index`)
```
msbrogli: In previous version, MySQL supported `O(1)` search using `tx_proposal_id` even though it is a multi-key hash index. Is it true for the latest version? Maybe we should include a `tx_proposal_id` hash index.

DATABASE.md
```diff
@@ -48,19 +70,53 @@ CREATE TABLE `token` (
...
+);
+
+
+CREATE TABLE `tx_proposal_outputs` (
+  `tx_proposal_id` varchar(36) NOT NULL,
+  `index` tinyint unsigned NOT NULL,
+  `address` varchar(34) NOT NULL,
+  `token_id` varchar(64) NOT NULL,
+  `value` bigint DEFAULT NULL,
+  `timelock` int unsigned DEFAULT NULL,
```
msbrogli: Why do we need a timelock here? I feel we should never create a tx_proposal with locked outputs.

src/api/errors.ts
```diff
@@ -8,9 +8,20 @@
 export enum ApiError {
   MISSING_PARAMETER = 'missing-parameter',
   INVALID_BODY = 'invalid-body',
+  INVALID_TX_WEIGHT = 'invalid-tx-weight',
+  INVALID_SELECTION_ALGO = 'invalid-selection-algo',
```
msbrogli: Why abbreviate?

src/api/wallet.ts
```diff
@@ -1,20 +1,11 @@
-/**
```
msbrogli: Removed the license header by mistake?

src/commons.ts
```diff
@@ -0,0 +1,254 @@
+import { ServerlessMysql } from 'serverless-mysql';
```
msbrogli: Missing license header.

src/commons.ts
```diff
@@ -0,0 +1,254 @@
...
+
+import {
```
msbrogli: Would you create a new PR for the refactor? It is way harder to review a mix of a big refactor with new features.

msbrogli (changes requested):

msbrogli:
@andreabadesso I haven't finished the review. I stopped when I reached a major refactor mixed with a new feature. I suggested you to split this PR in two parts: (i) a refactor, and (ii) the tx proposal feature.

DATABASE.md
andreabadesso: Because the service is runnning on lambda, it should be stateless.

DATABASE.md
andreabadesso: I agree. There is a separate issue to use an ORM and optimize the database though

DATABASE.md
andreabadesso: This commit was killed in the rebase, but I believe you were asking about `timelock` in `tx_proposal_outputs`

`tx_proposal_outputs` are the outputs that are being created, not the utxos that are being spent

DATABASE.md
andreabadesso: I agree, but there is already an issue to add indexes to tables

This will come with the ORM

DATABASE.md
msbrogli: I got it. Thanks for the explanation.

src/api/txProposalCreate.ts
andreabadesso: Done in https://github.com/HathorNetwork/hathor-wallet-service/pull/36

PR #21, opened 2021-02-25
refactor: export common methods

andreabadesso:


msbrogli (approved):

src/db.ts
```diff
@@ -1109,3 +1109,34 @@ export const getTokenInformation = async (
...
+  const entry = { key: 'version_check', value: now };
+
+  await mysql.query(
+    'INSERT INTO `metadata` SET ? ON DUPLICATE KEY UPDATE `value` = ?',
+    [entry, now],
+  );
+};
+
+/**
+ * Update latest version_check time
```
pedroferreira1, line 1129: Docstring is the same as in the method `updateLastVersionCheck`

src/db.ts
```diff
@@ -1109,3 +1109,34 @@ export const getTokenInformation = async (
...
 };
+
+/**
+ * Update latest version_check time
+ *
+ * @param mysql - Database connection
+ * @param now - The current timestamp
+ */
+export const updateLastVersionCheck = async (mysql: ServerlessMysql, now: number): Promise<void> => {
+  const entry = { key: 'version_check', value: now };
```
pedroferreira1, line 1120: Should we set this key as a constant (maybe in this file), in order to prevent writing more than once this key (`updateLastVersionCheck` and `getLastVersionCheck` at least for now.

pedroferreira1 (approved):

src/db.ts
andreabadesso: This was 100% refactored on the `tx_proposal` PR:
 
https://github.com/HathorNetwork/hathor-wallet-service/pull/22/

src/db.ts
andreabadesso: This was 100% refactored on the `tx_proposal` PR:

https://github.com/HathorNetwork/hathor-wallet-service/pull/22/

PR #22, opened 2021-02-25
feat: tx proposal (2)

andreabadesso:


.env.production
```diff
@@ -1,3 +1,4 @@
 STAGE=production
-MAX_ADDRESS_GAP=20
+MAX_ADDRESS_GAP=10
```
msbrogli: Why are you changing it?

DATABASE.md
```diff
@@ -44,6 +51,21 @@ CREATE TABLE `address_tx_history` (
   PRIMARY KEY (`address`,`tx_id`,`token_id`)
 );
 
+CREATE TABLE `version_data` (
```
msbrogli, line 56: Arght! I understand that it will run in a lambda and so we need to store these pieces of information somewhere. Should we add a `last_updated_at` field to force a query to the full node once in a while? Should we `count(*)` and guarantee there's one, and only one, row in this table?

I feel we should add a comment here explaining why it exists and that there must be exactly one row in this table.

src/api/errors.ts
```diff
@@ -8,9 +8,20 @@
 export enum ApiError {
   MISSING_PARAMETER = 'missing-parameter',
   INVALID_BODY = 'invalid-body',
-  INVALID_PARAMETER = 'invalid-parameter',
+  INVALID_TX_WEIGHT = 'invalid-tx-weight',
+  INVALID_SELECTION_ALGO = 'invalid-selection-algo',
```
msbrogli: `INVALID_SELECTION_ALGORITHM`?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+  requested: number;
+  available: number;
+}
+
+const bodySchema = Joi.object({
+  id: Joi.string()
+    .required(),
+  outputs: Joi.array()
+    .items(
+      Joi.array()
```
msbrogli: 1) Shouldn't we validate the content of the array?
2) Isn't it supposed to be an array of objects?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+
+/*
+ * Create a tx-proposal.
+ *
+ * This lambda is called by API Gateway on POST /txproposals
+ */
+export const create: APIGatewayProxyHandler = async (event) => {
+  await maybeRefreshWalletConstants(mysql);
+
+  const eventBody = (function parseBody(body) {
```
msbrogli, line 89: Missing typing?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+      };
+    }
+
+    // The error is in at least one of the parameters, we should fail with INVALID_PARAMETER
+    // while returning the first invalid parameter
+    const parameter = firstErrorDetails.path[0];
+    const apiError = firstErrorDetails.type === 'any.required' ? ApiError.MISSING_PARAMETER : ApiError.INVALID_PARAMETER;
+
+    return {
+      statusCode: 200,
```
msbrogli: 200? Isn't better to use 4xx or 5xx?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+        parameter,
+      }),
+    };
+  }
+
+  const body = value;
+  const walletId = body.id;
+
+  if (body.outputs.length > hathorLib.transaction.getMaxOutputsConstant()) {
+    await closeDbConnection(mysql);
```
msbrogli: Why don't you wrap this method with another method that call `closeDbConnection(mysql)` for you? I feel it is very error prone to do it before each return.

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.TOO_MANY_OUTPUTS, outputs: body.outputs.length }),
+    };
+  }
+
+  const outputs = parseValidateOutputs(body.outputs);
+  const inputs = body.inputs ? parseValidateInputs(body.inputs) : null;
+
+  if (!outputs || outputs.length === 0) {
+    await closeDbConnection(mysql);
```
msbrogli: Same here.

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.INVALID_PARAMETER, parameter: 'inputs' }),
+    };
+  }
+
+  const inputSelectionAlgo = (function getInputAlgoFromBody() {
+    if (!body.inputSelectionAlgo) return InputSelectionAlgo.USE_LARGER_UTXOS;
```
msbrogli: Wrong format.

```
if (...) {
  [code]
}
```

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+      body: JSON.stringify({ success: false, error: ApiError.INSUFFICIENT_INPUTS, insufficient: insufficientInputs }),
+    };
+  }
+
+  const addresses = await getUnusedAddresses(mysql, walletId);
+  const changeOutputs = getChangeOutputs(diff, addresses);
+
+  const finalOutputs = outputs.concat(changeOutputs);
+
+  // TODO get limit from the full-node (https://github.com/HathorNetwork/hathor-wallet-service/issues/12)
```
msbrogli: TODO! Or we can have some constants in this project.

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+      token: output[2],
+      timelock: output[3],
+    };
+
+    if (!hathorLib.transaction.isAddressValid(parsed.address)) {
+      // invalid address
+      return null;
+    }
+
+    if (typeof parsed.address !== 'string' || typeof parsed.value !== 'number'
```
msbrogli: `parsed.value` must also be integer. Maybe use `Number.isInteger(parsed.value)`?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+ * @returns The parsed inputs, or null if there's been an error
+ */
+export const parseValidateInputs = (inputs: unknown[]): IWalletInput[] => {
+  const parsedInputs = [];
+  for (const input of inputs) {
+    const parsed = {
+      txId: input[0],
+      index: input[1],
+    };
+    if (typeof parsed.txId !== 'string' || typeof parsed.index !== 'number') {
```
msbrogli: `parsed.index` must be an interger. Maybe use `Number.isInteger(parsed.index)`?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+export const getOutputsBalance = (outputs: IWalletOutput[], now: number): TokenBalanceMap => {
+  let outputsBalance = null;
+  for (const output of outputs) {
+    const decoded = { type: 'P2PKH', address: output.address, timelock: output.timelock };
+    // take advantage of TokenBalanceMap.fromTxOuput
+    const txOutput = {
+      decoded,
+      value: output.value,
+      token: output.token,
+      // TODO not handling authority outputs
```
msbrogli: TODO

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+    if (balance.total() < 0) {
+      changeOutputs.push({
+        address: addresses[addressToUse++],
+        value: Math.abs(balance.total()),
+        token,
+        timelock: null,
+      });
+
+      if (addressToUse >= addresses.length) {
+        // this treats an unlikely case, where we have more change outputs than addresses. In this case,
```
msbrogli, line 305: That's an interesting design decision. We should create an issue to discuss it further. For now, we can keep the current behavior.

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,497 @@
...
+export const useLargerUtxos = async (
+  _mysql: ServerlessMysql,
+  walletId: string,
+  tokenId: string,
+  balance: number,
+): Promise<Utxo[]> => {
+  const finalUtxos: Utxo[] = [];
+
+  let remainingBalance = balance;
+  const valueUtxos = await getWalletSortedValueUtxos(_mysql, walletId, tokenId);
```
msbrogli, line 348: Do we need to load all of them from database? Some wallets might have hundreds of thousands utxos.

src/api/txProposalSend.ts
```diff
@@ -0,0 +1,237 @@
...
+      statusCode: 200,
+      body: JSON.stringify({
+        success: false,
+        error: ApiError.INVALID_TX_WEIGHT,
+        message: `Expected weight >= ${calculatedTxWeight}`,
+      }),
+    };
+  }
+
+  const txHex = hathorLib.transaction.getTxHexFromData(txData);
```
msbrogli, line 155: Shouldn't we verify the PoW before moving forward?

src/api/txProposalSend.ts
```diff
@@ -0,0 +1,237 @@
...
+    return {
+      statusCode: 200,
+      body: JSON.stringify({
+        success: true,
+        txProposalId,
+        txHex,
+      }),
+    };
+  } catch (e) {
+    await updateTxProposal(
```
msbrogli, line 185: We need to rollback the transaction here, right?

src/utils.ts
```diff
@@ -81,3 +143,65 @@ export const closeDbConnection = async (mysql: ServerlessMysql): Promise<void> =
...
+
+/**
+ * Check if the received weight is in the acceptable weight range.
+ * This is calculated using the calculateTxWeight from the wallet-lib.
+ *
+ * @returns A boolean with the result
+ */
+export const validateWeight = (calculated: number, received: number): boolean => (
+  (received >= calculated)
+  && (received < (calculated * ACCEPTABLE_WEIGHT_RANGE))
```
msbrogli, line 161: Why do we have a superior limit? If the nonce matches, I feel we should accept it.

src/commons.ts
```diff
@@ -217,16 +218,37 @@ export const getWalletBalances = async (
...
+      rewardSpendMinBlocks: apiResponse.reward_spend_min_blocks,
+      maxNumberInputs: apiResponse.max_number_inputs,
+      maxNumberOutputs: apiResponse.max_number_outputs,
+    };
 
-    // TODO: Maybe when checkApiVersion fails we could use the current values on
-    // the wallet instead of propagating the error
+    await updateVersionData(mysql, versionData);
+  } else {
+    hathorLib.transaction.updateTransactionWeightConstants(
```
msbrogli, line 251: Shouldn't we do it every call, even when fetching data from the full node?

src/api/wallet.ts
```diff
@@ -58,61 +52,66 @@ export const get: APIGatewayProxyHandler = async (event) => {
...
     InvocationType: 'Event',
     Payload: JSON.stringify({ xpubkey, maxGap }),
   };
   try {
     // TODO setup lambda error handling. It's not an error on lambda.invoke, but an error during lambda execution
     await lambda.invoke(params).promise();
   } catch (e) {
-    await updateWalletStatus(mysql, walletId, WalletStatus.ERROR);
-    return closeDbAndGetError(mysql, ApiError.UNKNOWN_ERROR, { message: e.message });
+    // TODO handle
```
msbrogli, line 131: TODO

src/db.ts
```diff
@@ -1055,6 +1131,65 @@ export const maybeUpdateLatestHeight = async (mysql: ServerlessMysql, height: nu
...
+    min_tx_weight: data.minTxWeight,
+    min_tx_weight_coefficient: data.minTxWeightCoefficient,
+    min_tx_weight_k: data.minTxWeightK,
+    token_deposit_percentage: data.tokenDepositPercentage,
+    reward_spend_min_blocks: data.rewardSpendMinBlocks,
+    max_number_inputs: data.maxNumberInputs,
+    max_number_outputs: data.maxNumberOutputs,
+  };
+
+  await mysql.query(
```
msbrogli, line 1169: Should we assert there's one and only one record in this table?

msbrogli (changes requested):
I reviewed all files except the tests.

DATABASE.md
andreabadesso: https://github.com/HathorNetwork/hathor-wallet-service/blob/36c82d14b84e4a8b97759fbc7efe280b5ec365a8/src/commons.ts#L221

We are already checking "expiration time" on the method above, that's why we have a `timestamp` field. Currently the expiration time (`VERSION_CHECK_MAX_DIFF`) is set to 1 hour

I agree with the comment explaining why it exists.

I just noticed that Iḿ using `version` as the primary key, this will be an issue when we change the connected fullnode or update the current fullnode to another version as it will add a second row on the database.

I will refactor this to only allow for exactly one row

src/api/errors.ts
andreabadesso: Done in 3bbcb468296f173bc2f82cf0f5da5e36cdac934f

.env.production
andreabadesso: Can't remember why I changed it.

Changed it back to 20

src/api/txProposalSend.ts
andreabadesso: I agree, but I suggest leaving a comment there for now with a TODO to handle the transaction rollback because we have a task to handle transactions and rollbacks 

What do you think?

src/api/wallet.ts
andreabadesso: I think we can just `console.error` for now to have the error on cloudwatch/serverless dashboard and decide how to handle wallet load failures on another PR as this has nothing to do with `tx_proposal` and was done previously to it, what do you think?

src/commons.ts
andreabadesso: `hathorLib.version.checkApiVersion` already updates the weight, calling exactly the methods I'm using on the `else` code block.

We should be safe as long as the api on hathorLib don't change.

@pedroferreira1 I suggest refactoring `checkApiVersion` method on the hathor-lib to make it do _less_, it is currently:

* requesting the fullnode for version data
* updating the lib internally with the version data
* returning it

What do you think?

src/utils.ts
pedroferreira1: I discussed this with @andreabadesso and the idea was to add a protection, in order to prevent a malicious user to send a tx with a higher weight and we would overload the tx-mining-service. Not sure if we could prevent this using  other strategy.

src/api/txProposalCreate.ts
andreabadesso: I will remove the TODO as this is being done in the `tx_proposal token actions`

src/api/txProposalCreate.ts
andreabadesso: 1. We should.
2. I agree, the API is currently expecting an array of arrays on inputs and outputs. I will refactor it to expect an array of objects and validate it properly

Created issue:
https://github.com/HathorNetwork/hathor-wallet-service/issues/23

src/api/txProposalCreate.ts
andreabadesso: I feel that we should leave it without enforcing type on user input before validating it with Joi

src/api/txProposalCreate.ts
andreabadesso: I agree!! Will refactor all status codes for errors

src/api/txProposalCreate.ts
andreabadesso: I agree, but I think this refactor can be done when we add an ORM. We already have an task for that on the Epic.

I can create an issue so we won't forget

What do you think?

src/api/txProposalCreate.ts
andreabadesso: Done

src/api/txProposalCreate.ts
andreabadesso: Done!

src/api/txProposalCreate.ts
andreabadesso: I agree. Done!

src/api/txProposalCreate.ts
andreabadesso: Done. Also did it for timelock as it should also be an integer

src/api/txProposalCreate.ts
andreabadesso: Done: https://github.com/HathorNetwork/hathor-wallet-service/issues/25

src/api/txProposalCreate.ts
andreabadesso: I agree that we shouldn't need to load all of them in memory from the database.

We can try to do it on MySQL by refactoring this method to select utxos until their sum is higher than the necessary balance, ordering by value

I would suggest creating an issue for this optimization instead of handling it on tx-proposal. What do you think?

src/api/txProposalSend.ts
andreabadesso: I agree.

I think that this validation should be done on the wallet-lib tho and we don't have a method to do that yet. Created the issue at https://github.com/HathorNetwork/hathor-wallet-lib/issues/152

When it's done, we can validate the PoW

Created the issue: https://github.com/HathorNetwork/hathor-wallet-service/issues/26

src/api/balances.ts
```diff
@@ -44,27 +40,14 @@ export const get: APIGatewayProxyHandler = async (event) => {
     return closeDbAndGetError(mysql, ApiError.WALLET_NOT_READY);
   }
 
-  let tokenId: string = null;
+  const tokenIds: string[] = [];
   if (params && params.token_id) {
-    tokenId = params.token_id;
+    const tokenId = params.token_id;
     // TODO validate tokenId
```
pedroferreira1: What's this validation? Only if it's a valid hex?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,499 @@
...
+  available: number;
+}
+
+const bodySchema = Joi.object({
+  id: Joi.string()
+    .required(),
+  outputs: Joi.array()
+    .items(
+      Joi.array()
+        .min(4)
```
pedroferreira1: I didn't understand this min/max. What's the meaning?

src/api/txProposalCreate.ts
pedroferreira1: Important to say in the issue about the content validation of the array also. This was the main problem when I was using `express-validator`, to validate the content of an optional object.

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,499 @@
...
+      // There was an error parsing the body, we should fail with INVALID_PAYLOAD
+      return {
+        statusCode: 200,
+        body: JSON.stringify({ success: false, error: ApiError.INVALID_PAYLOAD }),
+      };
+    }
+
+    // The error is in at least one of the parameters, we should fail with INVALID_PARAMETER
+    // while returning the first invalid parameter
+    const parameter = firstErrorDetails.path[0];
```
pedroferreira1: Why do we return only the first invalid parameter and not all of them?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,499 @@
...
+    const missing = checkMissingUtxos(inputs, inputUtxos);
+
+    if (missing.length > 0) {
+      await closeDbConnection(mysql);
+      return {
+        statusCode: 200,
+        body: JSON.stringify({ success: false, error: ApiError.INPUTS_NOT_FOUND, missing }),
+      };
+    }
+    // check if inputs sent by user are not part of another tx proposal
```
pedroferreira1, line 163: Is this a `TODO`?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,499 @@
...
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.TOO_MANY_INPUTS, inputs: inputUtxos.length }),
+    };
+  }
+
+  // the difference between inputs and outputs will be the change
+  const inputsBalance = getInputsBalance(inputUtxos);
+  const diff = TokenBalanceMap.merge(outputsBalance, inputsBalance);
+
+  // Make sure diff is 0 or lower, which means inputs sum is grater than (or equal to) outputs sum.
```
pedroferreira1: Typo: greater

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,499 @@
...
+  }
+  if (insufficientInputs.length > 0) {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.INSUFFICIENT_INPUTS, insufficient: insufficientInputs }),
+    };
+  }
+
+  const addresses = await getUnusedAddresses(mysql, walletId);
```
pedroferreira1, line 192: Are we adding the `changeAddress` parameter in this API in another PR?

src/api/txProposalCreate.ts
```diff
@@ -0,0 +1,499 @@
...
+  if (finalOutputs.length > hathorLib.transaction.getMaxOutputsConstant()) {
+    // we also need to do this check here, as we may have added change outputs
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ success: false, error: ApiError.TOO_MANY_OUTPUTS, outputs: finalOutputs.length }),
+    };
+  }
+
+  // shuffle array
```
pedroferreira1: Maybe we should explain here that we shuffle in order to give some more privacy to the user, given that the last output won't always be the change.

src/utils.ts
```diff
@@ -81,3 +143,65 @@ export const closeDbConnection = async (mysql: ServerlessMysql): Promise<void> =
...
+ *   'address4': 4,
+ * }
+ * ```
+ *
+ * @param xpubkey - The xpubkey
+ * @param startIndex - Generate addresses starting from this index
+ * @param quantity - Amount of addresses to generate
+ * @returns A list with the generated addresses and corresponding index
+ */
+export const getHathorAddresses = (xpubkey: string, startIndex: number, quantity: number): StringMap<number> => {
```
pedroferreira1: The wallet lib has a method exactly like this. Check the method `getAddresses` in `src/helpers.js`.

src/api/txProposalDestroy.ts
```diff
@@ -0,0 +1,76 @@
...
+  updateTxProposal,
+  removeTxProposalOutputs,
+} from '@src/db';
+import { TxProposalStatus } from '@src/types';
+import { closeDbConnection, getDbConnection, getUnixTimestamp } from '@src/utils';
+
+const mysql = getDbConnection();
+
+/*
+ * Send a transaction.
```
pedroferreira1: This comment should be another thing, right? Something related to destroy a tx proposal.

src/api/txProposalSend.ts
```diff
@@ -0,0 +1,237 @@
...
+
+const bodySchema = Joi.object({
+  timestamp: Joi.number()
+    .required(),
+  parents: Joi.array()
+    .required()
+    .length(2),
+  weight: Joi.number()
+    .required(),
+  nonce: Joi.number()
```
pedroferreira1, line 51: Can we specify with `Joi` that we want an integer for `nonce`? In order to throw error if the user sends a float?

truncate.sql
```diff
@@ -0,0 +1,14 @@
+SELECT Concat('TRUNCATE TABLE ',table_schema,'.',TABLE_NAME, ';') FROM INFORMATION_SCHEMA.TABLES where  table_schema in ('hathor');
```
pedroferreira1: I think this file should not be commited.

src/db.ts
```diff
@@ -1109,3 +1244,202 @@ export const getTokenInformation = async (
...
+ * Get the unused addresses for a wallet.
+ *
+ * @remarks
+ * An unsued address is an address with 0 transactions. Addresses are ordered by index, ascending.
+ *
+ * @param mysql - Database connection
+ * @param walletId - The wallet's id
+ * @returns List of unused addresses
+ */
+export const getUnusedAddresses = async (mysql: ServerlessMysql, walletId: string): Promise<string[]> => {
```
pedroferreira1, line 1272: This method is being used to create a change output but it might not be the best. The ideal scenario would be to have the change addresses coming from a different path, however this would require a bigger refactor. The problem with this is that we expect the change address to be an address that was never shared with the user but an address with 0 transactions doesn't mean that was never shared with the user.

pedroferreira1 (changes requested):

src/api/txProposalCreate.ts
andreabadesso: Created a new PR for this refactor:

https://github.com/HathorNetwork/hathor-wallet-service/pull/27

src/api/txProposalCreate.ts
andreabadesso: @msbrogli 

Refactored this on the status code PR:
https://github.com/HathorNetwork/hathor-wallet-service/pull/27

src/api/txProposalCreate.ts
andreabadesso: Refactored this on the status code PR:

https://github.com/HathorNetwork/hathor-wallet-service/pull/27

src/utils.ts
andreabadesso: Oops. This is a leftover from the rebase I did with `dev`, this was already refactored to use the wallet-lib's method.

Removed the method and the test case for it.

aac782848a779a14d2b223a1d02edfed734379a6

src/api/txProposalDestroy.ts
andreabadesso: Fixed

src/api/txProposalSend.ts
andreabadesso: Yes we can, done in 17318c61ed681e27e887e305152770f6daa2a635

truncate.sql
andreabadesso: I agree. Done.

src/api/txProposalCreate.ts
andreabadesso: Done

src/api/txProposalCreate.ts
andreabadesso: I suggest doing it in another PR. 

Created the issue here: https://github.com/HathorNetwork/hathor-wallet-service/issues/28

src/api/txProposalCreate.ts
andreabadesso: I agree. Done

src/api/txProposalCreate.ts
andreabadesso: We are currently expecting inputs and outputs to be an array instead of an object. This will be refactored, we've got an issue for that: https://github.com/HathorNetwork/hathor-wallet-service/issues/23

src/api/txProposalCreate.ts
andreabadesso: Done in 1d50f33323168003a5bb2d969a7268693a6f8ac4

src/api/txProposalCreate.ts
andreabadesso: Done in PR https://github.com/HathorNetwork/hathor-wallet-service/pull/29

src/api/txProposalCreate.ts
andreabadesso: Done in PR https://github.com/HathorNetwork/hathor-wallet-service/pull/29

src/api/txProposalCreate.ts
andreabadesso: Done in 594a2f4d3a1975d1736f40352e355307f9b8ab30

src/api/balances.ts
andreabadesso: Done in PR: https://github.com/HathorNetwork/hathor-wallet-service/pull/30

src/db.ts
pedroferreira1: An issue was created to handle this later: https://github.com/HathorNetwork/hathor-wallet-service/issues/31

pedroferreira1 (approved):

DATABASE.md
andreabadesso: Done in 1fb1fd22fb67107d6dd36bbc6606cc6d6a7a1434

src/db.ts
andreabadesso: Done in 1fb1fd22fb67107d6dd36bbc6606cc6d6a7a1434

src/api/txProposalCreate.ts
msbrogli: I agree. Thanks!

src/api/txProposalCreate.ts
msbrogli: I agree. Let's do it separately. We can get the db cursor and iterate over it. This way we can control what keeps in memory and what will be collected by the garbage collector.

src/api/txProposalSend.ts
msbrogli: Great! Thanks!

src/api/txProposalSend.ts
msbrogli: I agree. Thanks!

src/utils.ts
msbrogli: But I feel this protection must be in the tx-mining-service. The wallet is not affected by it at all.

src/commons.ts
msbrogli: @andreabadesso Seems like a good suggestion. But, for now, I feel you can leave a note and move forward. And also double check that we have a test that will fail if the hathorLib changes in the future.

src/api/wallet.ts
msbrogli: I agree with your suggestion. But I think we should have an issue to keep track of it.

DATABASE.md
msbrogli: Great! Thanks!

msbrogli (approved):

pedroferreira1 (approved):

src/api/wallet.ts
andreabadesso: Done https://github.com/HathorNetwork/hathor-wallet-service/issues/32

src/utils.ts
andreabadesso: I created an issue so we can discuss this further: https://github.com/HathorNetwork/hathor-wallet-service/issues/33

src/api/txProposalCreate.ts
andreabadesso: I agree. Added the issue at https://github.com/HathorNetwork/hathor-wallet-service/issues/34

PR #27, opened 2021-03-01
refactor: status codes and error wrapper

### LINKED ISSUE #24: Refactor status codes on errors
I agree!! Will refactor all status codes for errors

_Originally posted by @andreabadesso in https://github.com/HathorNetwork/hathor-wallet-service/pull/22#discussion_r583738976_

andreabadesso:
Refactored status codes to something other than 200 and wrapped errors to a common method


src/api/utils.ts
```diff
@@ -12,6 +12,27 @@ import { ApiError } from '@src/api/errors';
 import { StringMap } from '@src/types';
 import { closeDbConnection } from '@src/utils';
 
+export const STATUS_CODE_TABLE = {
+  [ApiError.MISSING_PARAMETER]: 422,
```
pedroferreira1: What's the idea behind the differences between the errors 422 and 400?

I was reading that 422 means that it's semantically wrong but the syntax is correct. I feel all of our errors are like that, no? But I was also reading that 400 is also correct for all of them, it's just that 422 is more specific.

In my opinion we could use 400 for all parameter errors.

andreabadesso: The idea was to use 422 when the request has a syntactically correct body but has invalid data (semantically incorrect) as per the [RFC](https://tools.ietf.org/html/rfc4918#section-11.2)

And to use 400 when the request has valid body and valid data but failed to complete due to client error, e.g. valid input array but inputs were already spent

pedroferreira1: Sure, so the idea of what I was reading is correct.

What consists an invalid syntax? Because I believe most our errors are semantically invalid, e.g. `TOO_MANY_INPUTS`. This has nothing to do with syntax, right? It's our definition of a valid transaction.

andreabadesso: The rationale was:
```
Invalid JSON (sintaxically incorrect)? 400
Valid JSON (sintaxically correct)?
   JSON parameters failed validation? 422
   JSON paramteres passed validation?
     TX Proposal was created using those parameters? 201
     TX Proposal failed using those parameters? 400
```

msbrogli: Great comment to add to the docs.

msbrogli (approved):

src/api/utils.ts
andreabadesso: @msbrogli I discussed this further with @pedroferreira1 on slack and we came to the conclusion that it would be easier to understand if we use 400 for all client-errors. What do you think?

jansegre: I feel like 422 is too specific to WebDAV when we're not implementing anything else from WebDAV. For me status code 400 with at least a minimal response message that helps understand why the request is malformed is enough.

jansegre: I feel like 422 is too specific to WebDAV when we're not implementing anything else from WebDAV. For me status code 400 with at least a minimal response message that helps understand why the request is malformed is enough.

msbrogli: I'm fine either way.

andreabadesso: Updated all 422 status codes to 400

pedroferreira1 (approved):

PR #29, opened 2021-03-03
refactor: expecting an array of objects on inputs and outputs instead of array of arrays

### LINKED ISSUE #23: Refactor txProposalCreate to receive objects in inputs and outputs instead of array of arrays
_Originally posted by @andreabadesso in https://github.com/HathorNetwork/hathor-wallet-service/pull/22#discussion_r583729783_

andreabadesso:


tests/txProposal.test.ts
```diff
@@ -277,18 +217,6 @@ const _checkTxProposalTables = async (txProposalId, inputs, outputs): Promise<vo
   expect(await getTxProposalOutputs(mysql, txProposalId)).toStrictEqual(outputs);
 };
 
-test('POST /txproposals with null as param should fail with ApiError.INVALID_PAYLOAD', async () => {
```
andreabadesso, line 280: This test case was already covered on `tests/api.test.ts`, that's why I removed it

src/api/txProposalCreate.ts
```diff
@@ -36,16 +43,33 @@ const bodySchema = Joi.object({
...
     .items(
-      Joi.array()
-        .min(2)
-        .max(2),
+      Joi.object({
+        txId: Joi.string()
+          .alphanum()
+          .required(),
+        index: Joi.number()
+          .integer()
```
pedroferreira1, line 72: Can we set here that the `minValue` is `0`? Just to make sure the index won't be negative.

src/api/txProposalCreate.ts
```diff
@@ -36,16 +43,33 @@ const bodySchema = Joi.object({
...
     .items(
-      Joi.array()
-        .min(4)
-        .max(4),
+      Joi.object({
+        address: Joi.string()
+          .alphanum()
+          .required(),
+        value: Joi.number()
+          .integer()
```
pedroferreira1, line 51: Can we set here that the `minValue` is `1`? Just to make sure the user won't send zero/negative value.

src/api/txProposalCreate.ts
```diff
@@ -36,16 +43,33 @@ const bodySchema = Joi.object({
...
+          .required(),
+        token: Joi.string()
+          .alphanum(),
+        timelock: Joi.number()
+          .integer()
```
pedroferreira1, line 57: The same as the other. Negative timestamp is invalid.

src/api/txProposalCreate.ts
```diff
@@ -66,31 +90,24 @@ export const create: APIGatewayProxyHandler = async (event) => {
     }
   }(event.body));
 
-  const { value, error } = bodySchema.validate(eventBody);
+  const { value, error } = bodySchema.validate(eventBody, {
```
pedroferreira1, line 96: Maybe add a comment here to explain why we are using 

```
abortEarly: false,
convert: false,
```

pedroferreira1 (changes requested):

src/api/txProposalCreate.ts
andreabadesso: Done in 501c5ce4c3ac14a309de3452edbd143ac786747b

src/api/txProposalCreate.ts
andreabadesso: Done in 501c5ce4c3ac14a309de3452edbd143ac786747b

src/api/txProposalCreate.ts
andreabadesso: Done in 501c5ce4c3ac14a309de3452edbd143ac786747b

src/api/txProposalCreate.ts
andreabadesso: Done in 412018c1f866bf4c742bceed3259bcc4b4f5df96

pedroferreira1 (approved):

tests/txProposal.test.ts
```diff
@@ -40,64 +38,6 @@ afterAll(async () => {
...
-  // less than 4 elements
-  outputs = [[ADDRESSES[0], 10, 'token1']];
-  expect(parseValidateOutputs(outputs)).toBeNull();
-
-  // wrong address type
-  outputs = [[10, 10, 'token1', 20]];
-  expect(parseValidateOutputs(outputs)).toBeNull();
-
-  // wrong value type
-  outputs = [[ADDRESSES[0], '10', 'token1', 20]];
```
msbrogli, line 58: Have we included these removed error cases to the direct calls to the API?

msbrogli (approved):

tests/txProposal.test.ts
andreabadesso: Yes, `parseValidateOutputs` and `parseValidateInputs` were validating the data types, we don't need them anymore because we are using `Joi` to validate user input.

I only removed their specific unit tests, but param validation tests are still on `tests/api.test.ts`

PR #30, opened 2021-03-03
feat: added params validation to missing routes

andreabadesso:
…h joi

### DIFF
--- a/src/api/balances.ts
+++ b/src/api/balances.ts
@@ -28,7 +28,6 @@ const paramsSchema = Joi.object({
     .required(),
   token_id: Joi.string()
     .alphanum()
-    .required()
     .optional(),
 });
 
--- a/src/api/balances.ts
+++ b/src/api/balances.ts
@@ -35,6 +35,10 @@ const paramsSchema = Joi.object({
  * Get the balances of a wallet
  *
  * This lambda is called by API Gateway on GET /balances
+ *
+ * XXX: If token_id is not sent as a filter, we return all token balances
+ * Maybe we should limit the amount of tokens to query the balance to prevent an user
+ * with a lot of different tokens in his wallet from doing an expensive query
  */
 export const get: APIGatewayProxyHandler = async (event) => {
   const params = event.queryStringParameters;
--- a/src/api/balances.ts
+++ b/src/api/balances.ts
@@ -19,8 +19,18 @@ import {
   getDbConnection,
   getUnixTimestamp,
 } from '@src/utils';
+import Joi from 'joi';
 
 const mysql = getDbConnection();
+
+const paramsSchema = Joi.object({
+  id: Joi.string()
+    .required(),
+  token_id: Joi.string()
+    .alphanum()
+    .required()
+    .optional(),
+});
 
 /*
  * Get the balances of a wallet
@@ -29,14 +39,24 @@ const mysql = getDbConnection();
  */
 export const get: APIGatewayProxyHandler = async (event) => {
   const params = event.queryStringParameters;
-  let walletId: string;
-  if (params && params.id) {
-    walletId = params.id;
-  } else {
-    return closeDbAndGetError(mysql, ApiError.MISSING_PARAMETER, { parameter: 'id' });
+
+  const { value, error } = paramsSchema.validate(params, {
+    abortEarly: false,
+    convert: false,
+  });
+
+  if (error) {
+    const details = error.details.map((err) => ({
+      message: err.message,
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
   }
 
+  const walletId = value.id;
   const status = await getWallet(mysql, walletId);
+
   if (!status) {
     return closeDbAndGetError(mysql, ApiError.WALLET_NOT_FOUND);
   }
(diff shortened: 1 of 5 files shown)


src/api/balances.ts
```diff
@@ -14,25 +14,45 @@ import { getWalletBalances } from '@src/commons';
...
 import { closeDbConnection, getDbConnection, getUnixTimestamp } from '@src/utils';
 
 const mysql = getDbConnection();
 
+const paramsSchema = Joi.object({
+  id: Joi.string()
+    .required(),
+  token_id: Joi.string()
+    .alphanum()
+    .required()
```
pedroferreira1: Why is it required and optional?

src/api/txhistory.ts
```diff
@@ -16,14 +16,32 @@ import {
...
 const MAX_COUNT = 15;
-
 const htrToken = hathorLib.constants.HATHOR_TOKEN_CONFIG.uid;
 
+const paramsSchema = Joi.object({
+  id: Joi.string()
+    .required(),
+  token_id: Joi.string()
+    .alphanum()
+    .required()
```
pedroferreira1: It shouldn't be required, right?

src/api/balances.ts
```diff
@@ -14,25 +14,45 @@ import { getWalletBalances } from '@src/commons';
...
+    .optional(),
```
pedroferreira1, line 31: Should we have the default as HTR? Or the API returns all token balances if no token_id is used?

src/api/txhistory.ts
```diff
@@ -16,14 +16,32 @@ import {
...
+    .default(htrToken)
+    .optional(),
+  skip: Joi.number()
```
pedroferreira1, line 32: Should we add here `min(0)`?

src/api/txhistory.ts
```diff
@@ -16,14 +16,32 @@ import {
...
+    .integer()
+    .default(0)
+    .optional(),
+  count: Joi.number()
+    .integer()
+    .default(MAX_COUNT)
+    .optional(),
```
pedroferreira1, line 42: Maybe add `.positive()` and `max(MAX_COUNT)`

src/api/wallet.ts
```diff
@@ -15,26 +15,39 @@ import {
...
-    walletId = params.id;
-  } else {
-    return {
-      statusCode: 200,
-      body: JSON.stringify({ success: false, error: ApiError.MISSING_PARAMETER, parameter: 'id' }),
-    };
+
+  const { value, error } = getParamsSchema.validate(params, {
+    abortEarly: false,
+    convert: true, // Skip and count will come as query params as strings
```
pedroferreira1: I think this comment is not from here, right?

pedroferreira1 (changes requested):

src/api/balances.ts
andreabadesso: Typo. Fixed

src/api/txhistory.ts
andreabadesso: Typo. Fixed

src/api/txhistory.ts
andreabadesso: I agree. Done

src/api/txhistory.ts
andreabadesso: I agree. Done

src/api/wallet.ts
andreabadesso: Typo. Fixed

src/api/balances.ts
msbrogli: Maybe we should add a parameter for that (e.g. `all_tokens`).

src/api/txhistory.ts
```diff
@@ -32,41 +52,27 @@ const htrToken = hathorLib.constants.HATHOR_TOKEN_CONFIG.uid;
 export const get: APIGatewayProxyHandler = async (event) => {
   const params = event.queryStringParameters;
 
-  let walletId: string;
-  if (params && params.id) {
-    walletId = params.id;
-  } else {
-    return closeDbAndGetError(mysql, ApiError.MISSING_PARAMETER, { parameter: 'id' });
-  }
+  const { value, error } = paramsSchema.validate(params, {
```
msbrogli, line 55: This code repeats itself many times. Should we move the options to a constant?

msbrogli (approved):

src/api/balances.ts
andreabadesso: By default we return all balances from all tokens if token_id is not passed

I like `token_id` parameter being a filter and suggest keeping all tokens as a default

pedroferreira1: I agree to keep returning all tokens by default and only filter when `token_uid` is passed.

pedroferreira1 (approved):

src/api/txhistory.ts
andreabadesso: The options are not always the same, I want `convert` to be false when the params are coming from the query string

msbrogli: Ok. I haven't noticed that.

src/api/balances.ts
msbrogli: Ok. Just add a note that we expect the number of tokens in a wallet to be small.

msbrogli (approved):

src/api/balances.ts
andreabadesso: Done in 9834750f23bf9a75db15d8b0a5d0ab2ce4e6df52

PR #36, opened 2021-03-10
feat: returning token list on txProposalCreate

andreabadesso:
We need to return the tokens array as it's needed to assemble the tx for mining

### DIFF
--- a/src/api/txProposalCreate.ts
+++ b/src/api/txProposalCreate.ts
@@ -228,6 +228,9 @@ export const create: APIGatewayProxyHandler = async (event) => {
 
   const retInputs = await Promise.all(inputPromises);
 
+  // We need to return the tokens array as it's needed to assemble the tx for mining
+  const tokens = Array.from(new Set([...outputsBalance.getTokens(), ...inputsBalance.getTokens()]));
+
   return {
     statusCode: 201,
     body: JSON.stringify({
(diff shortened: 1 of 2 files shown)


src/api/txProposalCreate.ts
```diff
@@ -215,13 +215,17 @@ export const create: APIGatewayProxyHandler = async (event) => {
 
   await closeDbConnection(mysql);
 
+  // We need to return the tokens array as it's needed to assemble the tx for mining
+  const tokens = Array.from(new Set([...outputsBalance.getTokens(), ...inputsBalance.getTokens()]));
```
pedroferreira1: If this is needed only for assembling the tx for mining, we don't need to add the tokens in the inputs. If we have a tx that is a destroy authority tx (which has only one input and no outputs), this tx has an empty tokens array.

pedroferreira1 (changes requested):

src/api/txProposalCreate.ts
andreabadesso: Done in 4c65abea9c2b550925ccca4aee49d7aab37e394f

andreabadesso: Done in 4c65abea9c2b550925ccca4aee49d7aab37e394f

msbrogli (approved):

pedroferreira1 (approved):

PR #44, opened 2021-03-15
feat: return address path in tx proposal create

### LINKED ISSUE #43: Return addressIndex in create tx proposal POST
This change was suggested here https://github.com/HathorNetwork/hathor-wallet-service/issues/10#issuecomment-763236110 and it's needed for the lib integration.

pedroferreira1:


tests/integration.test.ts
```diff
@@ -132,7 +132,7 @@ test('receive blocks and txs and then start wallet', async () => {
   await loadWallet({ xpubkey: XPUBKEY, maxGap }, null, null);
 
   await checkAfterReceivingTx2(true);
-}, 30000);
+}, 60000);
```
msbrogli, line 135: Why did you increase the timeout?

src/db.ts
```diff
@@ -261,6 +261,37 @@ export const updateExistingAddresses = async (mysql: ServerlessMysql, walletId:
   }
 };
 
+/**
+ * Get a wallet's address detail.
+ *
+ * @param mysql - Database connection
+ * @param walletId - Wallet id
+ * @param address - Address to get the detail
+ * @returns The details of the address {address, index, transactions}
```
msbrogli: `... or null if not found`.

DATABASE.md
```diff
@@ -35,13 +35,6 @@ CREATE TABLE `address_balance` (
   PRIMARY KEY (`address`,`token_id`)
 );
 
-CREATE TABLE `token` (
```
msbrogli, line 38: Why did you delete it?

msbrogli (changes requested):

tests/integration.test.ts
pedroferreira1: I increased just to pass the tests. Before the changes this test was almost getting timeout in my machine. With the changes, I believe because of the db access, it was getting timeout eventually. We already have an issue to add indexes in the db and understand why it's a bit slow to execute queries.

DATABASE.md
pedroferreira1: It was duplicated in this file.

andreabadesso (approved):

tests/integration.test.ts
msbrogli: Is it close to timeout with 30 seconds? Even 30 seconds seems a lot to me.

DATABASE.md
msbrogli: Ok. Thanks!

tests/integration.test.ts
pedroferreira1: Yes, I agree. That's why we have an issue to understand some queries slowness. @andreabadesso already said that sharing the mysql connection might be a problem but still need to investigate.

msbrogli: Ok. We can do it now but I think we should open an issue to reduce these timeouts to just a few seconds (or even less).

msbrogli (approved):

PR #47, opened 2021-03-24
feat: derive xpub to change path before getting addresses from the lib

### LINKED ISSUE #46: fix: Generate wallet using a derived xpub instead of a root xpub due to limitations on hardened derivation.
The wallet service was expecting the root xpub, which was wrong because we can't derive a hardened path with a xpub.

We must change to receive the xpub in the last hardened path (account).

pedroferreira1:
TODO: still need to update the wallet lib with the xpub derive method

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -55,7 +55,8 @@ export const generateAddresses = async (mysql: ServerlessMysql, xpubkey: string,
   const allAddresses: string[] = [];
 
   do {
-    const addrMap = hathorLib.helpers.getAddresses(xpubkey, highestCheckedIndex + 1, maxGap, 'mainnet');
+    // We currently generate only addresses from changeDerivation = 0
+    const addrMap = hathorLib.helpers.getAddresses(xpubkey, 0, highestCheckedIndex + 1, maxGap, 'mainnet');
     allAddresses.push(...Object.keys(addrMap));
 
     const results: DbSelectResult = await mysql.query(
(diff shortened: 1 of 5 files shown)


src/db.ts
```diff
@@ -54,8 +54,13 @@ export const generateAddresses = async (mysql: ServerlessMysql, xpubkey: string,
   const newAddresses: AddressIndexMap = {};
   const allAddresses: string[] = [];
 
+  // We currently generate only addresses in change derivation path 0
+  // so we derive our xpub to this path and use it to get the addresses
```
msbrogli, line 59: Add a comment about the change level in BIP-0044.

msbrogli (approved):

src/db.ts
```diff
@@ -54,8 +54,13 @@ export const generateAddresses = async (mysql: ServerlessMysql, xpubkey: string,
...
+  const derivedXpub = walletUtils.xpubDeriveChild(xpubkey, 0);
+
   do {
-    const addrMap = hathorLib.helpers.getAddresses(xpubkey, highestCheckedIndex + 1, maxGap, 'mainnet');
+
```
andreabadesso: Is this white line passing eslint?

andreabadesso (approved):

src/db.ts
pedroferreira1: It was not. Fixed, thanks

PR #49, opened 2021-03-24
feat: add validation for wallet start to check first wallet address

### LINKED ISSUE #48: Add first address validation when starting the wallet
Question: should this be required or optional?

Lib PR: https://github.com/HathorNetwork/hathor-wallet-lib/pull/179

pedroferreira1:


src/api/wallet.ts
```diff
@@ -71,6 +72,8 @@ export const get: APIGatewayProxyHandler = async (event) => {
 const loadBodySchema = Joi.object({
   xpubkey: Joi.string()
     .required(),
+  address0: Joi.string()
```
msbrogli: I called it `firstAddress` in hathor-wallet-headless. Should we use the same name here?

src/api/wallet.ts
```diff
@@ -111,6 +114,15 @@ export const load: APIGatewayProxyHandler = async (event) => {
     return closeDbAndGetError(mysql, ApiError.WALLET_ALREADY_LOADED, { status });
   }
 
+  const expectedAddress0 = value.address0;
+
+  const address0 = hathorLib.walletUtils.getAddressAtIndex(xpubkey, 0, 0, 'mainnet');
```
msbrogli: We should get the network from config, right?

msbrogli: Maybe we should make it optional in the config file. What do you think?

msbrogli (changes requested):

src/api/wallet.ts
pedroferreira1: Yes, done

src/api/wallet.ts
pedroferreira1: Makes sense, changed.

src/api/wallet.ts
pedroferreira1: Makes sense, done

msbrogli (approved):

msbrogli (approved):

src/api/wallet.ts
```diff
@@ -71,6 +72,7 @@ export const get: APIGatewayProxyHandler = async (event) => {
 const loadBodySchema = Joi.object({
   xpubkey: Joi.string()
     .required(),
+  firstAddress: Joi.string(),
```
andreabadesso: When `process.env.CONFIRM_FIRST_ADDRESS` is `'true'`, shouldn't `firstAddress` be required?

andreabadesso (changes requested):

src/api/wallet.ts
pedroferreira1: Done. Can you see if it's good?

andreabadesso: LGTM

andreabadesso (approved):

PR #58, opened 2021-04-08
feat: Wallet Service Authentication

### LINKED ISSUE #59: [Design-OLD] Wallet Service Authentication
# Motivation
Hathor Wallet Service needs an authentication mechanism. This authentication mechanism should support or at least consider session management for future improvements.

# Solution
## POST /login to generate a token
image

The purpose of this endpoint is to generate a token to be used later on every request to the API. 

It receives the following body parameters:
```ts
{
    "xpubkey": string,
    "timestamp": number,
    "signature": string,
}
```
To be valid, the signature should have been done to a message with `timestamp` and `xpubkey` concatenated using an `xprivkey` derivated on the path `m/1/1`.
The function that does this is:
```js
import bitcoreLib from "bitcore-lib";
import Message from "bitcore-message";

const verifySignature = (
  signature: string,
  xpubkey: string,
  timestamp: number
): boolean => {
  const message = String(timestamp).concat(xpubkey);
  const derivedXpubkey = new bitcoreLib.HDPublicKey(xpubkey).derive("m/1/1");
  const address = derivedXpubkey.publicKey.toAddress().toString();
  return new Message(message).verify(address, signature);
};
```

Currently, it generates a JWT token carrying:
```ts
{
    "xpubkey": string,
    "timestamp": number,
    "signature": string,
}
```

To generate such body params something like below can be done:
```js
const bitcoreLib = require("bitcore-lib");
const Message = require("bitcore-message");

const timestamp = Math.floor(Date.now() / 1000);
const xprivkey = new bitcoreLib.HDPrivateKey(
  "xprv9s21ZrQH143K3QTDL4LXw2F7HEK3wJUD2nW2nRk4stbPy6cq3jPPqjiChkVvvNKmPGJxWUtg6LnF5kejMRNNU3TGtRBeJgk33yuGBxrMPHi"
);
const derivedPrivateKey = xprivkey.derive("m/1/1");
const xpubkey = xprivkey.hdPublicKey;
const message = String(timestamp).concat(xpubkey);
const signature = new Message(message).sign(derivedPrivateKey.privateKey);

console.info(
  JSON.stringify(
    {
      timestamp,
      xpub: xpubkey.toString(),
      signature: signature.toString(),
    },
    null,
    4
  )
);
```

## Authorizer (intercepts all requests to the API)
image

Basically, API Gateway calls the authorizer lambda to generate a policy allowing or not the request to the protected endpoints.
Currently, the authorizer decodes the JWT (carried by the Header "Authorization") to check the signature and the expiration time.

# Open discussions
1. **Does the AWS Authorizer add another lambda call which implies a higher cost and response delay?**
Yes, but I'm investigating if the token can be cached on the API Gateway itself to mitigate this issue. Some info about it [here](https://www.alexdebrie.com/posts/lambda-custom-authorizers/#caching-your-custom-authorizers).

2. **How can we make the authentication stateful to allow more sophisticated features that depend on session management?**
Under investigation.

3. **New dependency `bitcore-message` and the multiples instances of `bitcore-lib`.**
To solve this, package.json must declare a resolution to a `bitcore-lib` version (the same as the `hathor-wallet-lib`):
```json
...
  "resolutions": {
    "bitcore-lib": "0.16.0"
  }
...
```
Besides that, to enforce this resolution I'm using `npx npm-force-resolutions`, which is not the best solution. Better ways to do it are under investigation.

natancamargo:
Design: #59

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2758,6 +2758,14 @@
       "integrity": "sha1-7ihweulOEdK4J7y+UnC86n8+ce4=",
       "dev": true
     },
+    "@types/jsonwebtoken": {
+      "version": "8.5.1",
+      "resolved": "https://registry.npmjs.org/@types/jsonwebtoken/-/jsonwebtoken-8.5.1.tgz",
+      "integrity": "sha512-rNAPdomlIUX0i0cg2+I+Q1wOUr531zHBQ+cV/28PJ39bSPKjahatZZ2LMuhiguETkCgLVzfruw/ZvNMNkKoSzw==",
+      "requires": {
+        "@types/node": "*"
+      }
+    },
     "@types/lodash": {
       "version": "4.14.155",
       "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.155.tgz",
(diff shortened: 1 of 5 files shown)


r4mmer:
These were already made in #69.

PR #61, opened 2021-04-20
feat: token actions implementation

andreabadesso:
This PR was abandoned, replaced by https://github.com/HathorNetwork/hathor-wallet-service/pull/82


DATABASE.md
```diff
@@ -83,18 +83,27 @@ CREATE TABLE `token` (
...
+  `symbol` varchar(5) NOT NULL,
+  PRIMARY KEY (`tx_proposal_id`) -- We only accept one token creation per tx_proposal for now.
+);
+
 CREATE TABLE `tx_proposal_outputs` (
   `tx_proposal_id` varchar(36) NOT NULL,
   `index` tinyint unsigned NOT NULL,
   `address` varchar(34) NOT NULL,
-  `token_id` varchar(64) NOT NULL,
+  `token_id` varchar(64) DEFAULT NULL,
```
pedroferreira1, line 104: Why this is default null now? We always have a `token_id`, right? Or the token creation outputs you are not setting the token id?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+    TokenActionType.REGULAR_TRANSACTION,
+    TokenActionType.CREATE_TOKEN,
+    TokenActionType.MINT_TOKEN,
+    TokenActionType.MELT_TOKEN,
+    TokenActionType.DELEGATE_MINT,
+    TokenActionType.DELEGATE_MELT,
+    TokenActionType.DESTROY_MINT,
+    TokenActionType.DESTROY_MELT,
+  ).default(TokenActionType.REGULAR_TRANSACTION),
+}).unknown(true);
```
pedroferreira1, line 26: What's this unknown?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export const createTxSchema = Joi.object({
```
pedroferreira1, line 28: We should have `changeAddress` here as well, right?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+      Joi.object({
+        address: Joi.string()
+          .alphanum()
+          .required(),
+        value: Joi.number()
+          .integer()
+          .positive()
+          .required(),
+        token: Joi.string()
+          .alphanum(),
```
pedroferreira1, line 41: The default is optional or required?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+          .min(0),
+      }),
+    ),
+  inputSelectionAlgo: Joi.string(),
+}).unknown(true);
+
+export const createTokenSchema = Joi.object({
+  name: Joi.string().required().max(30),
+  symbol: Joi.string().required().max(5).min(2),
+  amount: Joi.number().required(),
```
pedroferreira1, line 69: Add `.posivite()` and `.integer()`?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+        index: Joi.number()
+          .integer()
+          .required()
+          .min(0),
+      }),
+    ),
+}).unknown(true);
+
+export const mintTokenSchema = Joi.object({
+  amount: Joi.number().required().min(1),
```
pedroferreira1, line 94: Maybe we should have all amount fields with the same definition. I feel it's better to have as you defined the first output amount (`.integer().positive()`). What do you think?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+        index: Joi.number()
+          .integer()
+          .required()
+          .min(0),
+      }),
+    ),
+}).unknown(true);
+
+export const meltTokenSchema = Joi.object({
+  amount: Joi.number().required().min(1),
```
pedroferreira1, line 118: Same comment about amount.

src/api/txProposalSend.ts
```diff
@@ -143,8 +151,20 @@ export const send: APIGatewayProxyHandler = async (event) => {
     outputs,
   };
 
-  await maybeRefreshWalletConstants(mysql);
+  if (txProposal.type === TokenActionType.CREATE_TOKEN) {
```
pedroferreira1, line 154: Maybe add a comment here explaining that a create token transaction does not have the tokens array and has a specific version and name/symbol as custom fields.

src/db.ts
```diff
@@ -1509,3 +1578,43 @@ export const getTxProposalInputs = async (
   }
   return inputs;
 };
+
+export const getAuthorityUtxosForToken = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+  tokenId: string,
+  authority: number,
+  quantity = 1,
```
pedroferreira1, line 1587: Missing `:number` or typescript infers from the default value?

src/types.ts
```diff
@@ -550,3 +571,99 @@ export interface ApiResponse {
...
+  requested: number;
+  available: number;
+}
+
+export interface BaseTxBody {
+  version?: number;
+  inputSelectionAlgo?: string;
+}
+
+export interface CreateTxBody extends BaseTxBody {
```
pedroferreira1, line 605: Should we add `changeAddress`?

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+    return body.destinationAddress;
+  }());
+
+  const inputSelectionAlgo = (function getInputAlgoFromBody() {
+    if (!body.inputSelectionAlgo) {
+      return InputSelectionAlgo.USE_LARGER_UTXOS;
+    }
+
+    return InputSelectionAlgo[body.inputSelectionAlgo];
+  }());
```
pedroferreira1, line 233: Shouldn't we add below here this validation?

```
  if (!inputSelectionAlgo) {
    return closeDbAndGetError(mysql, ApiError.INVALID_SELECTION_ALGORITHM);
  }
```

I think at least all validations could be refactored in a single place.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+  const htrInputBalance = inputsBalance.get(uid);
+
+  if ((htrInputBalance.unlockedAmount * -1) < necessaryDepositAmount) {
+    return closeDbAndGetError(mysql, ApiError.INSUFFICIENT_INPUTS, { insufficient: [uid] });
+  }
+
+  // Get unused addresses to be used on change and authority outputs
+  const addresses = await getUnusedAddresses(mysql, walletId);
+
+  // Create the deposit change without creating an deposit output
```
pedroferreira1, line 278: a* deposit

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+  const now = getUnixTimestamp();
+
+  await createTxProposal(mysql, txProposalId, walletId, now, TokenActionType.CREATE_TOKEN);
+  await addTxProposalOutputs(mysql, txProposalId, finalOutputs);
+  await addTxProposalTokenInfo(mysql, txProposalId, body.name, body.symbol);
+
+  // DB operations are over, close the mysql conn:
+  await closeDbConnection(mysql);
+
+  const inputPromises = inputUtxos.map(async (utxo) => {
```
pedroferreira1, line 358: This also could be a separate method.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+    return closeDbAndGetError(mysql, ApiError.MISSING_AUTHORITY_INPUT, {
+      authority: hathorLib.constants.TOKEN_MINT_MASK,
+    });
+  }
+
+  // check if we selected enough hathor for the deposit
+  const inputsBalance = getInputsBalance(inputUtxos);
+  const htrInputBalance = inputsBalance.get(uid);
+
+  if ((htrInputBalance.unlockedAmount * -1) < necessaryDepositAmount) {
```
pedroferreira1, line 473: Why do you need to do it? `unlockedAmount` is negative?

Isn't it enough just `if necessaryDepositAmount  > htrInputBalance.unlockedAmount`?

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+  }
+
+  // check if we selected enough token to melt
+  const inputsBalance = getInputsBalance(inputUtxos);
+  const tokenInputBalance = inputsBalance.get(body.token);
 
-enum InputSelectionAlgo {
-  USE_LARGER_UTXOS = 'use-larger-utxos',
-}
+  if ((tokenInputBalance.unlockedAmount * -1) < body.amount) {
```
pedroferreira1, line 653: Same question as in the mint.

src/api/txProposalCreate.ts
```diff
@@ -167,54 +775,65 @@ export const create: APIGatewayProxyHandler = async (event) => {
...
-  const changeOutputs = getChangeOutputs(diff, addresses);
+  // get unused addresses to be used on outputs
+  const tokenIndex = 1; // We only allow the delegation of 1 token per api call, so this can be fixed.
 
-  const finalOutputs = outputs.concat(changeOutputs);
+  const finalOutputs: IWalletOutput[] = [];
 
-  // we also need to do this check here, as we may have added change outputs
+  // one output for each authority
+  for (let i = 0; i < body.amount; i++) {
```
pedroferreira1, line 812: I think I've missed this `amount` parameter in delegate authorities in the design. I am not sure we should have it, I believe should always be 1. Maybe we could have the parameter but as optional and the default could be 1

src/api/txProposalCreate.ts
```diff
@@ -228,8 +847,106 @@ export const create: APIGatewayProxyHandler = async (event) => {
...
+  }
+
+  // check if we have exactly one authority
+  const authorities = inputUtxos.filter((utxo) => (
+    utxo.authorities === authority
+    && utxo.tokenId === body.token
+  ));
+
+  // more than one authority was sent from the user
+  if (authorities.length > 1) {
```
pedroferreira1, line 913: The destroy we have the quantity, then it's possible to destroy more than one in a request, doesn't it?

pedroferreira1 (changes requested):

DATABASE.md
andreabadesso: Exactly. Token create actions will not have the token id yet

src/api/schemas.ts
andreabadesso: `unknown` will allow for more attributes other than the ones defined in the schema without failing

src/api/schemas.ts
andreabadesso: Default is optional

src/db.ts
andreabadesso: Typescript infers from the default value and will not allow me to set `:number` if I'm providing a default value

src/api/schemas.ts
msbrogli: Why would we accept an unknown type?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+    TokenActionType.DESTROY_MINT,
+    TokenActionType.DESTROY_MELT,
+  ).default(TokenActionType.REGULAR_TRANSACTION),
+}).unknown(true);
+
+export const createTxSchema = Joi.object({
+  destinationAddress: Joi.string().alphanum().max(34),
+  outputs: Joi.array()
+    .items(
+      Joi.object({
```
msbrogli, line 32: Should we support non-standard scripts? Maybe in a future release?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+  name: Joi.string().required().max(30),
+  symbol: Joi.string().required().max(5).min(2),
+  amount: Joi.number().required(),
+  destinationAddress: Joi.string().alphanum().max(34),
+  changeAddress: Joi.string().alphanum().max(34),
+  createMint: Joi.boolean().default(true),
+  createMelt: Joi.boolean().default(true),
+  meltDestination: Joi.string().alphanum().max(34),
+  mintDestination: Joi.string().alphanum().max(34),
+  version: Joi.number()
```
msbrogli, line 76: Why would we allow user to change the version?

src/api/schemas.ts
msbrogli: Can't we create a `const JoiTxInputAmount = Joi.number().required().integer().min(1)`?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+        index: Joi.number()
+          .integer()
+          .required()
+          .min(0),
+      }),
+    ),
+  inputSelectionAlgo: Joi.string(),
+}).unknown(true);
+
+export const createTokenSchema = Joi.object({
```
msbrogli, line 66: Brainstorm: Isn't it better to receive a serialized transaction and just sign the known inputs? I feel it would simplify the whole API and be much more flexible. We can create validation rules after parsing the serialized tx if we need to.

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+}).unknown(true);
+
+export const mintTokenSchema = Joi.object({
+  amount: Joi.number().required().min(1),
+  token: Joi.string().alphanum().required(),
+  createAnotherAuthority: Joi.boolean().default(true),
+  destinationAddress: Joi.string().alphanum().max(34),
+  changeAddress: Joi.string().alphanum().max(34),
+  authorityAddress: Joi.string().alphanum().max(34),
+  version: Joi.number()
```
msbrogli, line 100: Why would we allow user to change the version?

src/api/schemas.ts
```diff
@@ -0,0 +1,180 @@
...
+    ),
+}).unknown(true);
+
+export const meltTokenSchema = Joi.object({
+  amount: Joi.number().required().min(1),
+  token: Joi.string().alphanum().required(),
+  createAnotherAuthority: Joi.boolean().default(true),
+  destinationAddress: Joi.string().alphanum().max(34),
+  authorityAddress: Joi.string().alphanum().max(34),
+  version: Joi.number()
```
msbrogli, line 123: Why would we allow user to change the version?

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+
+    const missing = checkMissingUtxos(inputs, inputUtxos);
+
+    if (missing.length > 0) {
+      return closeDbAndGetError(mysql, ApiError.INPUTS_NOT_FOUND, { missing });
+    }
+
+    // check if inputs sent by user are not part of another tx proposal
+    if (checkUsedUtxos(inputUtxos)) {
+      return closeDbAndGetError(mysql, ApiError.INPUTS_ALREADY_USED);
```
msbrogli, line 113: Maybe `INPUTS_ALREADY_PROVISIONED`? The word used gives the feeling that it has already been spent so it would be a double spending.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+    }
+  } else {
+    for (const [tokenId, tokenBalance] of outputsBalance.iterator()) {
```
msbrogli, line 116: What if the inputs sent by the user are not enough? Should we generate an error or complete with more inputs?

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+  for (const [token, tokenBalance] of diff.iterator()) {
+    if (tokenBalance.total() > 0) insufficientInputs.push(token);
+  }
+  if (insufficientInputs.length > 0) {
+    return closeDbAndGetError(mysql, ApiError.INSUFFICIENT_INPUTS, { insufficient: insufficientInputs });
+  }
+
+  const addresses: string[] = await (async function getChangeAddresses() {
+    if (!body.destinationAddress) {
+      const destinationAddresses = await getUnusedAddresses(mysql, walletId);
```
msbrogli, line 143: Maybe create a `getChangeAddress()`? In the initial implementation, it can be a direct call to `getUnusedAddresses()` just like here.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+  await addTxProposalOutputs(mysql, txProposalId, finalOutputs);
+
+  await closeDbConnection(mysql);
+
+  const inputPromises = inputUtxos.map(async (utxo) => {
+    const addressDetail: AddressInfo = await getWalletAddressDetail(mysql, walletId, utxo.address);
+    // XXX We should store in address table the path of the address, not the index
+    // For now we return the hardcoded path with only the address index as variable
+    // The client will be prepared to receive any path when we add this in the service in the future
+    const addressPath = `m/44'/${hathorLib.constants.HATHOR_BIP44_CODE}'/0'/0/${addressDetail.index}`;
```
msbrogli, line 180: I agree it should be in the address table. For now, we should at least move to a separate method (e.g. `generateAddressPath()`).

src/api/txProposalCreate.ts
msbrogli: I agree with the refactor. It seems this piece of code has been duplicated several times.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+
+  const { uid } = hathorLib.constants.HATHOR_TOKEN_CONFIG;
+
+  const necessaryDepositAmount = hathorLib.tokensUtils.getDepositAmount(body.amount);
+
+  // we create a fake output to get available utxos from the database to fill it
+  const necessaryHtrBalance = new TokenBalanceMap();
+  necessaryHtrBalance.set(uid, new Balance(necessaryDepositAmount, 0)); // 1% deposit
+
+  // If inputs were sent, we must validate if they are enough for the amount of tokens we want to create.
```
msbrogli, line 243: This code seems equal to the code above. Maybe we should find a better way. What about a `fillTxInputs()` or `verifyManualInputs()` method?

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+    timelock: null,
+  }];
+
+  // Create the outputs with the mint and melt authorities if necessary
+
+  // Mint
+  if (body.createMint) {
+    const mintDestination: string = (function getMintDestinationAddress() {
+      if (!body.mintDestination) {
+        // TODO: Make sure we are not already using this address on change outputs
```
msbrogli, line 298: Pending TODO item.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
+    };
+
+    tokensOutputs.push(mintOutput);
+  }
+
+  // Melt
+  if (body.createMelt) {
+    const meltDestination: string = (function getMeltDestinationAddress() {
+      if (!body.meltDestination) {
+        // TODO: Make sure we are not already using this address on change outputs
```
msbrogli, line 320: Pending TODO item.

src/api/txProposalCreate.ts
```diff
@@ -24,115 +33,721 @@ import {
...
 } from '@src/types';
+import { ValidationResult } from 'joi';
 import { closeDbAndGetError } from '@src/api/utils';
 import { arrayShuffle, closeDbConnection, getDbConnection, getUnixTimestamp } from '@src/utils';
 import hathorLib from '@hathor/wallet-lib';
 
-const mysql = getDbConnection();
+const mysql = getDbConnection();
+
+enum InputSelectionAlgo {
```
msbrogli, line 54: **This comment is not related to the place I put it.**

Shouldn't we simplify and receive a serialized transaction with some instructions such as "fill inputs"? I feel the code is too complicated because there are too many possible cases. @andreabadesso @pedroferreira1

msbrogli (changes requested):

src/api/txProposalCreate.ts
pedroferreira1: I am thinking about how this would make the code simpler. We would still need to have the code to select the inputs, change output and tx validation, which are the main reasons for code complexity, isn't it?

PR #62, opened 2021-04-27
fix: added a 'ignore list' of transactions including the genesis txs

andreabadesso:
We could extend this to ignore other transactions, maybe have a list of transactions to ignore on an env variable but I think that this should be enough for the transactions on the `testnet` and `mainnet` as they won't ever change

### DIFF
--- a/src/txProcessor.ts
+++ b/src/txProcessor.ts
@@ -38,6 +38,19 @@ import { closeDbConnection, getDbConnection, getUnixTimestamp } from '@src/utils
 
 const mysql = getDbConnection();
 
+export const IGNORE_TXS = {
+  mainnet: [
+    '000006cb93385b8b87a545a1cbb6197e6caff600c12cc12fc54250d39c8088fc',
+    '0002d4d2a15def7604688e1878ab681142a7b155cbe52a6b4e031250ae96db0a',
+    '0002ad8d1519daaddc8e1a37b14aac0b045129c01832281fb1c02d873c7abbf9',
+  ],
+  testnet: [
+    '0000033139d08176d1051fb3a272c3610457f0c7f686afbe0afe3d37f966db85',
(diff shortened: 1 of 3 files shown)


src/txProcessor.ts
```diff
@@ -83,6 +96,14 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
   // TODO mysql error treatment
 
   const txId = tx.tx_id;
+  const network = process.env.NETWORK;
+
+  // we should ignore genesis transactions as they have no parents, inputs and outputs
```
pedroferreira1: The genesis blocks have the premine outputs. If you are adding this output in the database in another way we should at least make it clear here.

pedroferreira1 (approved):

src/txProcessor.ts
andreabadesso: Improved the commend to explain that genesis txs and pre-mined outputs should be already in the database when the service starts in 4706c8e20354ed60f88aaaf1b1298a1f28f3468d

src/txProcessor.ts
```diff
@@ -38,6 +38,19 @@ import { closeDbConnection, getDbConnection, getUnixTimestamp } from '@src/utils
 
 const mysql = getDbConnection();
 
+export const IGNORE_TXS = {
```
msbrogli, line 41: I think we should use this PR to update the README with the explanation about the genesis and also add the scripts to add the genesis transactions to the database.

msbrogli (changes requested):

src/txProcessor.ts
andreabadesso: Done in 5a06691b84ba7e99aeccce7f39e1d6f08da9a342

I added the inserts to add the genesis block to the database and the UTXO with the pre-mined tokens

msbrogli: Ok. Thanks!

msbrogli (approved):

PR #69, opened 2021-05-12
feature: wallet service authorization

### LINKED ISSUE #66: [Design] Wallet Service Authentication
# Summary
[summary]: #summary

When calling a function from the wallet service the user can/must (depending on the function) provide some sort of authentication,
proving he has the private key of some resource the function is trying to access.
Eventually this could support session management for future improvements.
This design is for the service that will authenticate the user and authorize his access to the wallet service resources.

# Motivation
[motivation]: #motivation

Hathor Wallet Service needs an authentication (authn) and authorization (authz) mechanisms.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

To make an authenticated call you need to provide a way for the service to verify that you have the private key associated with the wallet being accessed.
A way to authenticate is provide:

- A message
- The xpubkey
- The message signed by the private key

This way you could get the address from the xpubkey and validate that the message was signed with the private key of the address
Now, we can choose the message as a concatenation of available information, this way, anyone can build the original message and we can omit it from the required information.

```js
const xpubkey = "xxx";
const timestamp = 123;

const message = String(timestamp).concat(xpubkey);
```
To create the token we will be using the [jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) lib.
```js
const jwt = require('jsonwebtoken');

// secret to sign jwt
const secret = 'shhhhh';
// The info needed
const data = {
    ts: 123,
    xpub: "xxx",
    sign: "sss",
}

const token = jwt.sign(secret, secret);
```
Now that we have a way to create a token, we need to understand what will be signed.
```js
const jwt = require('jsonwebtoken');
const bitcoreLib = require("bitcore-lib");
const Message = require("./bitcore-message");

const hdprivkey = "xprv9s21ZrQH143K3e6DUVMPzeKHiiQPiyRNy2QTpvCTB8XYTL7vWAj6BRXmp8NW4ovQvDTemNnNyArVeJTKetWFfcTLgPQJangAt3pkp7wLGNS"
const secret = 'shhhhh';

const derivationPath = "m/44'/280'/0'/0" // chosen by the client

const timestamp = Math.floor(Date.now() / 1000);
const xprivkey = new bitcoreLib.HDPrivateKey(hdprivkey);
const derivedPrivateKey = xprivkey.deriveChild(derivationPath);
const xpubkey = derivedPrivateKey.hdPublicKey;
const address = xpubkey.publicKey.toAddress().toString();
const message = String(timestamp).concat(address);
const signature = new Message(message).sign(derivedPrivateKey.privateKey);

const data = {
    ts: timestamp,
    xpub: xpubkey.toString(),
    sign: signature.toString(),
};
const token = jwt.sign(data, secret);
```
Now to verify the generated token:
```js
const Message = require("./bitcore-message");
const jwt = require('jwt-simple');

const token = '...'
const data = jwt.verify(token, secret);
const xpubkey = new bitcoreLib.HDPublicKey(data.xpub);
const address = xpubkey.publicKey.toAddress().toString()
const message = String(data.ts).concat(address);
const valid = new Message(message).verify(address, data.sign);
```
The endpoints needed are:
- /token
    * post data to generate token with secret
- /verify
    * post token to open the token with the secret and verify signature

Since current deployment of the wallet service is made in aws lambda with AWS API Gateway we will use a lambda authorizer as described [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html).
We will use response format version 2.0 described in the [lambda authorizer doc](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html) for simplicity.

The jsonwebtoken is a complete implementation of the [rfc7519](https://tools.ietf.org/html/rfc7519) for JSON Web Tokens (JWT), so we could expand some functionalities like token expiration time (exp), asymmetric signing, etc.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

### POST /token

This is a public endpoint that validates and exchanges a set of data to a token
Input JSON body:
```ts
{
    "ts": number,
    "xpub": string,
    "sign": string,
}
```
Return JSON body:
```ts
{
    "token": string
}
```

### Lambda authorizer (/verify)

This function will serve to validate the token given to the client proving he owns the wallet identified by `xpub`.
Input JSON body:
```ts
{
    "token": string
}
```
Return JSON body:
```ts
{
    "isAuthorized": boolean,
    "context": object,
}
```

### API Gateway

The routes that need authorization will be protected with the `verify` lambda authorizer.
If the header `Authorization` is not present, the API Gateway returns a HTTP 401 response,
The client then should create a token using the `token` endpoint and retry the call.

Caching the authorizer response is desirable because if would reduce number of calls to the lambda authorizer.

r4mmer:


AUTHENTICATION.md
```diff
@@ -0,0 +1,48 @@
+# Authentication
+
+### Protected functions
+
+- wallet
+- proposeTx
+// TODO: list all protected views
```
andreabadesso: TODO

src/api/auth.ts
```diff
@@ -0,0 +1,214 @@
...
+import hathorLib from '@hathor/wallet-lib';
+// const hathorLib = require('@hathor/wallet-lib');
+
+const EXPIRATION_TIME_IN_SECONDS = 30 * 60;
+const MAX_TIMESTAMP_SHIFT_IN_SECONDS = 30;
+
+hathorLib.network.setNetwork(process.env.NETWORK);
+
+const bodySchema = Joi.object({
+  ts: Joi.number().required(),
```
andreabadesso: Maybe use https://joi.dev/api/?v=17.4.0#datetimestamptype

src/api/auth.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  const signature = value.sign;
+  const timestamp = value.ts;
+  const xpubkeyStr = value.xpub;
+
+  const timestampShiftInSeconds = Math.abs(
+    Math.floor(Date.now() / 1000) - timestamp,
+  );
+  if (timestampShiftInSeconds >= MAX_TIMESTAMP_SHIFT_IN_SECONDS) {
+    const details = {
```
andreabadesso: Maybe use:

```
const details = [{
    message: `...`
}]
```

To be consistent with the txProposalCreate API that returns a list of details on error

src/api/auth.ts
```diff
@@ -0,0 +1,214 @@
...
+    },
+  );
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ success: true, token }),
+  };
+};
+
+type Dictionary<T> = { [s: string]: T };
```
andreabadesso: Move this to the `types.ts` file?

src/api/errors.ts
```diff
@@ -16,6 +16,7 @@ export enum ApiError {
   INSUFFICIENT_FUNDS = 'insufficient-funds',
   INSUFFICIENT_INPUTS = 'insufficient-inputs',
   INVALID_PARAMETER = 'invalid-parameter',
+  INVALID_SIGNATURE = 'invalid-signature',
```
andreabadesso: Maybe use `AUTH_INVALID_SIGNATURE = 'invalid-auth-signature'` to distinguish it from a invalid input signature on `txProposalSend`

src/api/auth.ts
```diff
@@ -0,0 +1,214 @@
...
+      addr: address.toString(),
+      wid: walletId,
+    },
+    process.env.AUTH_SECRET,
+    {
+      expiresIn: EXPIRATION_TIME_IN_SECONDS,
+      // issuer:
+      // notBefore:
+      // audience:
+      // jwtId: // generate or accept?
```
andreabadesso: I think this is necessary to be able to blacklist JWT tokens in the future and I think we should generate it

Maybe for now, just generate a random UUID

I don't think we have an use for the other options yet

andreabadesso (changes requested):

src/api/auth.ts
r4mmer: This converts the timestamp value to a `Date`, here we are looking for the actual number value for the message.

src/api/auth.ts
r4mmer: Done.

src/api/auth.ts
r4mmer: This was to prevent a bug on serverless offline (if the context wasn't typed str to str dict it wasn't passed to the function), but i double checked and it was fixed, so i'm removing this.

src/api/errors.ts
r4mmer: Done.

src/api/auth.ts
r4mmer: Ok

andreabadesso (approved):

package.json
```diff
@@ -17,27 +18,28 @@
     "url": "https://github.com/HathorNetwork/hathor-wallet-service.git"
   },
   "dependencies": {
-    "@hathor/wallet-lib": "^0.18.0",
+    "@hathor/wallet-lib": "^0.19.0",
```
pedroferreira1: You will need the next version, right? The one with the open lib PR.

src/api/auth.ts
```diff
@@ -0,0 +1,205 @@
...
+// import { closeDbAndGetError } from '@src/api/utils';
+import hathorLib from '@hathor/wallet-lib';
+
+const EXPIRATION_TIME_IN_SECONDS = 30 * 60;
+const MAX_TIMESTAMP_SHIFT_IN_SECONDS = 30;
+
+hathorLib.network.setNetwork(process.env.NETWORK);
+
+const bodySchema = Joi.object({
+  ts: Joi.number().required(),
```
pedroferreira1: Maybe a bit too much but we can add `.number().positive().required()`

src/api/auth.ts
```diff
@@ -0,0 +1,205 @@
...
+import { v4 as uuid4 } from 'uuid';
+import {
+  getWalletId,
+  // getDbConnection,
+} from '@src/utils';
+import Joi from 'joi';
+import jwt from 'jsonwebtoken';
+import bitcore from 'bitcore-lib';
+import { ApiError } from '@src/api/errors';
+// import { closeDbAndGetError } from '@src/api/utils';
```
pedroferreira1: Remove?

src/api/auth.ts
```diff
@@ -0,0 +1,205 @@
...
+  APIGatewayProxyHandler,
+  APIGatewayTokenAuthorizerHandler,
+  CustomAuthorizerResult,
+  PolicyDocument,
+  Statement,
+} from 'aws-lambda';
+import { v4 as uuid4 } from 'uuid';
+import {
+  getWalletId,
+  // getDbConnection,
```
pedroferreira1: Remove?

src/api/addresses.ts
```diff
@@ -5,49 +5,27 @@
...
   getWallet,
   getWalletAddresses,
 } from '@src/db';
-import Joi from 'joi';
 import { closeDbConnection, getDbConnection } from '@src/utils';
 
-const mysql = getDbConnection();
+import { walletIdProxyHandler } from '@src/commons';
 
-const paramsSchema = Joi.object({
```
pedroferreira1, line 22: With this authorization change we are not sending the walletId as query parameters anymore, right? We are getting the id from the auth header?

src/api/auth.ts
```diff
@@ -0,0 +1,205 @@
...
+  const context = { walletId: principalId };
+  authResponse.context = context;
+
+  return authResponse;
+};
+
+export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = async (event) => {
+  const { authorizationToken } = event;
+  if (!authorizationToken) {
+    throw new Error('Unauthorized'); // returns a 401
```
pedroferreira1, line 180: Where do you handle this exception to return 401? Or is it automatically?

pedroferreira1 (changes requested):

package.json
r4mmer: Yes, this update was made by a tool i was testing to check for updates in dependencies, so i ended up updating this, but the authentication will need the version with the bitcore-lib updated to 8.25.10

src/api/addresses.ts
r4mmer: Yes, the token has the walletId and the authorizer extracts and embeds it in the `event.requestContext.authorizer`. This way the function already has access to the walletId the user owns (and is viewing right now). The walletId only is necessary if the api allows users to request info of other wallets.

src/api/auth.ts
r4mmer: According to aws documentation if you return the string `"Unauthorized"` or throw an error containing this string, it will return a 401 default response. (Also, the validation is case-sensitive).

src/api/auth.ts
r4mmer: Done.

src/api/auth.ts
r4mmer: Done.

src/api/auth.ts
r4mmer: Done.

src/api/auth.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+/**
+ * Verify a signature for a given timestamp and xpubkey
+ *
+ * @param signature - The signature done by the xpriv of the wallet
+ * @param xpub - The xpubkey of the wallet as a string
+ * @param timestamp - Unix Timestamp of the signature
+ * @returns true if the timestamp, the xpubkey and signature matches
```
luislhl: I think there is a mismatch between this docs and the params in the function

r4mmer: You're right, adding the new parameters now

andreabadesso (approved):

PR #70, opened 2021-05-13
feat: create new API to return addresses to use of a wallet

pedroferreira1:


andreabadesso (approved):

src/types.ts
```diff
@@ -73,6 +73,11 @@ export interface AddressInfo {
   transactions: number;
 }
 
+export interface ShortAddressInfo {
+  address: string;
+  index: number;
```
msbrogli, line 85: Should we add the derivation path?

src/db.ts
```diff
@@ -980,6 +981,40 @@ export const getWalletAddresses = async (mysql: ServerlessMysql, walletId: strin
...
+     WHERE \`wallet_id\` = ?
+       AND \`transactions\` = 0
+       AND \`index\` > (
+         SELECT MAX(\`index\`)
+           FROM \`address\`
+           WHERE \`wallet_id\` = ?
+           AND \`transactions\` > 0
+       )
+  ORDER BY \`index\`
+       ASC`, [walletId, walletId]);
```
msbrogli: Should we limit the max number of results? I understand the gap limit should automatically do it but it would prevent problems from eventual bugs or later changes in the way addresses are handled.

serverless.yml
```diff
@@ -60,6 +60,13 @@ functions:
           path: addresses
           method: get
           cors: true
+  getAddressesToUseApi:
```
msbrogli: Maybe add a description of what this API does? I'm not sure where to add it but I feel we should have it. In the OpenAPI description file seems a good place if there is one.

msbrogli (changes requested):

jansegre (commented):
Looks OK. I only think that "addresses to use" is an awkward expression. I don't immediately understand what is it supposed to mean, is it supposed to be "new addresses + potentially already used addresses". Either way, here are some alternatives (which IMO are better but I'm obviously biased here):

- new addresses
- next addresses
- available addresses

src/types.ts
pedroferreira1: Done

pedroferreira1:
> Looks OK. I only think that "addresses to use" is an awkward expression. I don't immediately understand what is it supposed to mean, is it supposed to be "new addresses + potentially already used addresses". Either way, here are some alternatives (which IMO are better but I'm obviously biased here):
> 
> * new addresses
> * next addresses
> * available addresses

Changed the name to newAddresses and the API name to /addresses/new

src/db.ts
pedroferreira1: I agree it's important. Is it good enough to add a LIMIT as the gap limit of the wallet?

This raised another concern for me that we are not limiting the gap limit the wallets are requesting to use. We must limit them at what? 100? I know some use cases use much more than that but not sure we should allow use cases to use our public wallet service.

serverless.yml
pedroferreira1: Done.

tests/api.test.ts
```diff
@@ -96,6 +97,46 @@ test('GET /addresses', async () => {
...
+    { address: ADDRESSES[5], index: 5, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[6], index: 6, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[7], index: 7, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[8], index: 8, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[9], index: 0, walletId: null, transactions: 0 },
+    { address: ADDRESSES[10], index: 0, walletId: 'test', transactions: 0 },
+  ]);
+
+  // missing param
+  await _testInvalidPayload(newAddressesGet, ['"id" is required']);
```
andreabadesso: This will never be reached, id is not a parameter anymore, it comes from the authentication, right?

tests/api.test.ts
```diff
@@ -96,6 +97,46 @@ test('GET /addresses', async () => {
...
+
+  // missing wallet
+  await _testMissingWallet(newAddressesGet, 'some-wallet');
+
+  // wallet not ready
+  await _testWalletNotReady(newAddressesGet);
+
+  // success case
+  const event = makeGatewayEvent({ id: 'my-wallet' });
```
andreabadesso: You should use `makeGatewayEventWithAuthorizer`

andreabadesso (changes requested):

tests/api.test.ts
pedroferreira1: Fixed

tests/api.test.ts
pedroferreira1: Fixed, thanks

src/db.ts
msbrogli: I feel our public wallet service should allow at most 40 as GAP Limit.

msbrogli (approved):

andreabadesso (approved):

PR #71, opened 2021-05-13
feat: handle reorg

andreabadesso:
# Summary
[summary]: #summary

When a reorg has happened in the fullnode, we need to detect it and re-calculate our balances in the database

This PR intends to describe the changes that are necessary to be able to handle reorgs and propose an implementation with those changes.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

On the original design, the wallet-service didn't store the transactions, making it very hard to handle a re-org as there were not enough information to re-calculate the balance.

The proposed solution is to not only store all transactions that the service receives but to refactor the `utxo` table to become a `tx_output` table, storing all transactions outputs and what tx_id spent it (if it's spent).

This way, we can easily detect our last valid block and re-calculate by removing all `tx_outputs` that are voided and then recalculating the `address_balance`.

## Re-org algorithm

The algorithm we use to handle the re-org is the following:

1. Store all block `tx_id`s where `height` > `reorg_height` in a variable A
2. Delete from the `blocks` table all blocks where height > `reorg_height`
3. Delete from the `transaction` table all `tx_id`s in A
4. Set all `wallet_tx_history` records that have `tx_id` in A as `dirty`
5. Set all `address_tx_history` records that have `tx_id` in A as `dirty`
6. Store all transactions that spend `tx_outputs` from A in B (using the `spent_by` column)
7. Delete all `tx_outputs` that belong to transactions in A
8. Update all `tx_outputs` that are spent by transactions in B to have `spent_by = NULL`
9. Set `A = B` and jump to step 3 until `len(A) = 0`
10. Remove height from all transactions on the `transaction` table that have `height` > `reorg_height` (they are back to the mempool)
11. Recalculate balances for all addresses that were affected by the re-org using the `tx_output` table

## Detecting the latest valid block

To detect the wallet-service's last valid block on the database, we do a binary search on the `blocks` database, checking if our stored block is still valid on the full-node.

As we need to store all transactions that the wallet-service accepts, we need a table to represent them:

```sql
CREATE TABLE `transaction` (
  `tx_id` varchar(64) NOT NULL,
  `timestamp` int unsigned NOT NULL,
  `version` tinyint unsigned NOT NULL,
  `voided` boolean NOT NULL DEFAULT false,
  `height` int unsigned DEFAULT NULL,
  PRIMARY KEY (`tx_id`)
);
```

The `utxo` table was renamed to `tx_output` because we now store all `tx_outputs`, not only the unspent ones.

```sql
CREATE TABLE `tx_output` (
  `tx_id` varchar(64) NOT NULL,
  `token_id` varchar(64) NOT NULL,
  `locked` tinyint unsigned NOT NULL DEFAULT '0',
  `tx_proposal` varchar(36) DEFAULT NULL,
  `tx_proposal_index` tinyint unsigned DEFAULT NULL,
+   `spent_by` varchar(64) DEFAULT NULL, -- `spent_by` should be set every time a received transaction uses this `tx_ouput`
+   `dirty` tinyint unsigned NOT NULL DEFAULT '0', -- When a transaction is voided, we set it to `dirty` so the wallet frontend won't lose information
  PRIMARY KEY (`tx_id`,`index`)
);
```

We also removed the `metadata` table because we can get the wallet-service's best block from the `blocks` database.

The new implementation of `getLatestHeight` query is:

```sql
SELECT MAX(`height`) as value FROM `blocks`;
```

## New re-org handler (onHandleReorgRequest)

There is a new handler, available only via aws-sdk invoke that is responsible for calling the handleReorg method and return if it was sucessfully or not.

This handler will be called by the wallet-service daemon when a re-org is detected.

## How to test using the simulator

We are using the hathor-core simulator to test re-org scenarios, this is done by using a custom simulation script, that simulate a healhty blockchain than waits before simulating a split-brain condition, all this while leaving all apis available for the daemon to consume.

Here is the simulation code:

```python
import os
os.environ['HATHOR_CONFIG_FILE'] = 'hathor.conf.testnet'

import random
from typing import Optional

from hathor.simulator import FakeConnection, MinerSimulator, RandomTransactionGenerator, Simulator

class AndreSimulation:
    seed_config: Optional[int] = None

    def __init__(self):
        if self.seed_config is None:
            self.seed_config = random.randint(0, 2**32 - 1)

        self.simulator = Simulator()
        self.simulator.set_seed(self.seed_config)
        self.simulator.start()

        print('-'*30)
        print('Simulation seed config:', self.simulator.seed)
        print('-'*30)

    def wait_signal(self, signal) -> None:
        import os
        import time
        path = '/tmp/{}'.format(signal)
        print('Waiting for a signal... path={}'.format(path))
        while not os.path.isfile(path):
            time.sleep(1)
        os.remove(path)

    def run_http_api(self, manager: 'HathorManager') -> None:
        print('Running HTTP API...')
        from hathor.cli.run_node import RunNode
        class MyRunNode(RunNode):
            def __init__(self, *args, **kwargs):
                pass
        run = MyRunNode()
        run.parser = run.create_parser()
        args = run.parse_args(argv=['--status', '8080', '--wallet-index'])
        run.manager = manager
        run.wallet = manager.wallet
        run.register_resources(args)

    def test_andre_reorg(self):
        from hathor.cli.util import setup_logging
        setup_logging()

        # Create first peer that has an HTTP API.
        peer11 = self.simulator.create_peer()
        self.run_http_api(peer11)

        from twisted.internet import reactor
        reactor.callInThread(self._msbrogli_run, peer11)
        reactor.run()

    def _msbrogli_run(self, peer11):
        from hathor.graphviz import GraphvizVisualizer
        from twisted.python.failure import Failure

        # Create other peers.
        peer12 = self.simulator.create_peer()
        conn = FakeConnection(peer11, peer12, latency=0.001)
        miner1 = MinerSimulator(peer11, hashpower=10e6)
        miner1.start()
        self.simulator.add_connection(conn)

        peer21 = self.simulator.create_peer()
        peer22 = self.simulator.create_peer()
        conn = FakeConnection(peer21, peer22, latency=0.001)
        self.simulator.add_connection(conn)

        # Connect both networks.
        special_conn = FakeConnection(peer12, peer22, latency=0.001)
        self.simulator.add_connection(special_conn)

        gen_tx1 = RandomTransactionGenerator(peer11, rate=1/30., hashpower=10e6, ignore_no_funds=True)
        gen_tx1.start()

        gen_tx2 = RandomTransactionGenerator(peer21, rate=1/30., hashpower=10e6, ignore_no_funds=True)
        gen_tx2.start()

        # Run for 10 minutes to get a higher mining difficulty.
        self.simulator.run(600)

        # Add another miner.
        miner2 = MinerSimulator(peer21, hashpower=50e6)
        miner2.start()

        # Run for 1h.
        self.simulator.run(3600)

        graphviz = GraphvizVisualizer(peer11.tx_storage, include_verifications=True)
        dot = graphviz.dot()
        dot.render('peer11-pre')

        self.wait_signal('x')
        print('Signal received. Continuing..');

        # Disconnect networks causing a split brain.
        special_conn.disconnect(Failure(Exception('split-brain')))
        self.simulator.connections.remove(special_conn)
        # Replace peer11 wallet by a new one.
        self.simulator.create_wallet(peer11)

        # Run for 1h.
        self.simulator.run(3600)

        # Connect the networks again causing a reorg.
        conn = FakeConnection(peer11, peer21, latency=0.001)
        self.simulator.add_connection(conn)

        self.simulator.run(1800)

        miner1.stop()
        miner2.stop()
        gen_tx1.stop()
        self.simulator.run(600)

        graphviz = GraphvizVisualizer(peer11.tx_storage, include_verifications=True)
        dot = graphviz.dot()
        dot.render('peer11-post')


if __name__ == '__main__':
    x = AndreSimulation()
    x.test_andre_reorg()
```

The steps I'm taking to simulate the reorg are:

1. Run the wallet-service locally using with a testnet database (including the genesis transactions on DATABASE.md):
`npm run offline`

2. Run `simulate.py`:
`python simulate.py`

This will simulate a blockchain with transactions and wait for a "signal" (a file on /tmp/x)

3. Run the `wallet-service-daemon` locally, pointing to the simulator and the local wallet-service:

```bash
export STAGE=local
export AWS_REGION=us-east-1
export MAX_ADDRESS_GAP=20
export NETWORK=testnet
export SERVICE_NAME=hathor-wallet-service
export DEFAULT_SERVER=http://localhost:8080/v1a/
export CONSOLE_LEVEL=debug

cd dist && node index.js
```

This will sync the local wallet service until the latest block from the full-node and keep it running, waiting for new blocks.

4. Signal the simulator to continue:

`touch /tmp/x`

This will cause the simulator to simulate a split-brain condition, causing the blockchain to re-org.

The daemon will notice it and transition to a `reorg` state.

5. Call manually the `onHandleReorgRequest` handler

I created a script to make it easier to do this (you will need AWS-CLI credentials configured already):

```javascript
/*
 * This script starts a connection with a local full node and listens to new transactions. Upon
 * receiving a tx, it invokes the wallet-service lambda to handle it.
 *
 * This code is to be used on dev environments only.
 */

require('dotenv').config();

const AWS = require('aws-sdk');

// we need to set a region even if we don't make any calls
AWS.config.update({region:'us-east-1'});


const lambda = new AWS.Lambda({
  apiVersion: '2015-03-31',
  endpoint: process.env.STAGE === 'local'
    ? process.env.WALLET_SERVICE_LOCAL_URL || 'http://localhost:3002'
    : `https://lambda.${process.env.AWS_REGION}.amazonaws.com`,
});

const sendEvent = () => {
const params = {
  FunctionName: `${process.env.SERVICE_NAME}-${process.env.STAGE}-onHandleReorgRequest`,
};

  lambda.invoke(params, (err, data) => {
    if (err) {
      console.error('ERROR', err);
      return process.exit(1);
    }

    console.log('data: ', data);
  });
};

sendEvent();
```

This will handle the reorg and re-calculate the balance. We can manually check if the balances are correct, but the reorg handler already assert if balances are correct by checking if the `address_tx_history` SUM matches the `address_balance` table


src/commons.ts
```diff
@@ -259,3 +283,112 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+      start = midHeight + 1;
+    } else {
+      end = midHeight - 1;
+    }
+  }
+
+  return latestValidBlock;
+};
+
+export const handleReorg = async (mysql: ServerlessMysql): Promise<void> => {
```
msbrogli: Missing docstring.

src/commons.ts
```diff
@@ -259,3 +283,112 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+  const allTxsAfterHeight = await getTxsAfterHeight(mysql, height);
+  let txs: Tx[] = allTxsAfterHeight.filter((tx) => [
+    hathorLib.constants.BLOCK_VERSION,
+    hathorLib.constants.MERGED_MINED_BLOCK_VERSION,
+  ].indexOf(tx.version) > -1);
+
+  let removedUtxoList: DbTxOutput[] = [];
+
+  while (txs.length > 0) {
+    await removeTxs(mysql, txs);
```
msbrogli: Why would we remove the transactions? If they have not been voided, they will still be valid but without any confirmation.

msbrogli: Oh! I just noticed that these are blocks. Maybe rename the variable to `blocks`? Or `blocksAfterHeight`?

src/commons.ts
msbrogli: Should we log each step of the reorg? I feel it might be slow to run and logging each step might help debugging.

src/fullnode.ts
```diff
@@ -0,0 +1,32 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import axios from 'axios';
+
+export const BASE_URL = process.env.DEFAULT_SERVER;
+export const TIMEOUT = 10000;
+
+export const create = (baseURL = BASE_URL): any => {
```
msbrogli, line 18: Missing docstring.

src/txProcessor.ts
```diff
@@ -106,6 +148,23 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
...
+  if (dbTx) {
+    // ignore tx if we already have it confirmed on our database
+    if (dbTx.height) {
+      return;
+    }
+
+    // set height and break out because it was already on the mempool
+    // so we can consider that our balances have already been calculated
+    // and the utxos were already inserted
+    await addTx(mysql, txId, tx.height, tx.timestamp, tx.version);
```
msbrogli: Maybe rename to `updateTx(...)`? It might be a direct call to `addTx(...)` in the current implementation. I just feel they might differ in the future, and it is more intuitive to call updateTx instead of addTx here.

src/txProcessor.ts
```diff
@@ -85,14 +91,50 @@ export const onNewTxEvent = async (event: SQSEvent): Promise<APIGatewayProxyResu
...
+  // @ts-ignore
+  await handleReorg(mysql);
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ success: true }),
+  };
+};
+
+export const onSearchForLatestValidBlockRequest: APIGatewayProxyHandler = async () => {
```
msbrogli, line 147: Missing docstring.

src/txProcessor.ts
```diff
@@ -85,14 +91,50 @@ export const onNewTxEvent = async (event: SQSEvent): Promise<APIGatewayProxyResu
...
+  // @ts-ignore
+  await addNewTx(event.body, now, blockRewardLock);
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ success: true }),
+  };
+};
+
+export const onHandleReorgRequest: APIGatewayProxyHandler = async () => {
```
msbrogli, line 120: Missing docstring.

src/txProcessor.ts
```diff
@@ -85,14 +91,50 @@ export const onNewTxEvent = async (event: SQSEvent): Promise<APIGatewayProxyResu
   };
 };
 
+export const onNewTxRequest: APIGatewayProxyHandler = async (event) => {
```
msbrogli, line 100: Missing docstring.

src/db.ts
```diff
@@ -534,32 +540,63 @@ export const addUtxos = async (
...
+ *
+ * @remarks
+ * This method adds a transaction to the transaction table
+ *
+ * @param mysql - Database connection
+ * @param txId - Transaction id
+ * @param timestamp - The transaction timestamp
+ * @param version - The transaction version
+ */
+export const addTx = async (
```
msbrogli: Maybe rename to `addOrUpdateTx()`?

DATABASE.md
```diff
@@ -99,7 +100,7 @@ CREATE TABLE `tx_proposal_outputs` (
   PRIMARY KEY (`tx_proposal_id`,`index`)
 );
 
-CREATE TABLE `utxo` (
+CREATE TABLE `tx_output` (
   `tx_id` varchar(64) NOT NULL,
```
msbrogli: `tx_id` might point to a block, right? If yes, I think we should add a comment here to clarify this point.

msbrogli (approved):

src/commons.ts
andreabadesso: Done in 4226ced035e28859eae4ad9c33ec9a9b8d688707

src/commons.ts
andreabadesso: This variable is used to remove transactions as well, on the loop

Maybe I could rename it to `hashes`, what do you think?

src/commons.ts
andreabadesso: I agree!

Done in 6a0a50d0b627ef911ad6f997eac756acc7d94c0b

src/fullnode.ts
andreabadesso: Done in b1bceda1f66a75436b92370c0a3273d1d523bcd5

src/txProcessor.ts
andreabadesso: I agree.

Done in cafcd5697f43008ac3e63d88583993b03f4f3c20

src/db.ts
andreabadesso: Done in cafcd5697f43008ac3e63d88583993b03f4f3c20

src/txProcessor.ts
andreabadesso: Done in bee212d80ccf8578feab5935f1e8983def225b16

src/txProcessor.ts
andreabadesso: Done in bee212d80ccf8578feab5935f1e8983def225b16

src/txProcessor.ts
andreabadesso: Done in bee212d80ccf8578feab5935f1e8983def225b16

DATABASE.md
andreabadesso: I agree

Done in 122f158abf756125aab83845a62fcf0aa338a3fb

DATABASE.md
```diff
@@ -142,8 +145,18 @@ CREATE TABLE `wallet_tx_history` (
...
+  `dirty` tinyint unsigned NOT NULL DEFAULT '0',
   PRIMARY KEY (`wallet_id`,`token_id`,`tx_id`)
 );
+
+CREATE TABLE `transaction` (
+  `tx_id` varchar(64) NOT NULL,
+  `timestamp` int unsigned NOT NULL,
+  `version` tinyint unsigned NOT NULL,
+  `voided` boolean NOT NULL DEFAULT false,
+  `height` int unsigned DEFAULT NULL,
```
pedroferreira1, line 152: Are we going to store all blocks duplicated in `blocks` and `transaction` tables?

src/utils.ts
```diff
@@ -160,3 +161,19 @@ export const validateWeight = (calculated: number, received: number): boolean =>
...
   && (received < (calculated * ACCEPTABLE_WEIGHT_RANGE))
 );
+
+/**
+ * Requests the fullnode for the requested block information and returns
+ * if it is voided or not
+ *
+ * @returns A boolean with the result
+ */
+export const checkBlockForVoided = async (txId: string): Promise<boolean> => {
```
pedroferreira1: I prefer `isBlockVoided` but I am ok with it. Or `isTxVoided` because we are not checking if it's a block, so it's a more generic method.

src/fullnode.ts
```diff
@@ -0,0 +1,37 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import axios from 'axios';
+
+export const BASE_URL = process.env.DEFAULT_SERVER;
+export const TIMEOUT = 10000;
+
+/**
+ * Creates an handler for requesting data from the fullnode
```
pedroferreira1: Typo: a handler

src/commons.ts
```diff
@@ -259,3 +283,135 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+ *
+ * @returns A Block instance with the last block that is not voided.
+ */
+export const searchForLatestValidBlock = async (mysql: ServerlessMysql): Promise<Block> => {
+  // Get our current best block.
+  const latestHeight: number = await getLatestHeight(mysql);
+  const bestBlock: Block = await getBlockByHeight(mysql, latestHeight);
+
+  let start = 0;
+  let end = bestBlock.height; // 63
```
pedroferreira1: What's 63? (on the line below as well)

src/commons.ts
```diff
@@ -259,3 +283,135 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+  let latestValidBlock = bestBlock; // 63
+
+  while (start <= end) {
```
pedroferreira1, line 303: I was thinking about this algorithm here and maybe I am exaggerating in the optmization but I think we've noticed that reorgs of size 1 happen regularly in our network, so every time this reorg happens we will need to do this binary search in all blocks of the network, right now this would be ~20 iterations, what's is not too much but if 99% of our reorgs are of size < 2, what do you think about getting the latest 3 blocks of the network to see if any of them are valid before starting the binary search? Then, if the assumption of 99% of reorgs are small is correct, we will prevent this longer binary search 99% of the time.

@msbrogli @andreabadesso

src/db.ts
```diff
@@ -534,32 +540,82 @@ export const addUtxos = async (
...
+
 /**
  * Remove a tx inputs from the utxo table.
  *
  * @param mysql - Database connection
  * @param inputs - The transaction inputs
+ * @param txId - The transaction that spent these utxos
  */
-export const removeUtxos = async (mysql: ServerlessMysql, inputs: TxInput[]): Promise<void> => {
+export const removeUtxos = async (mysql: ServerlessMysql, inputs: TxInput[], txId: string): Promise<void> => {
```
pedroferreira1: Should we update this method name? It should be called `updateOutputSpentBy` imo, what do you think?

src/txProcessor.ts
```diff
@@ -132,10 +214,11 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
 
   // add outputs to utxo table
   markLockedOutputs(tx.outputs, now, heightlock !== null);
+  await addOrUpdateTx(mysql, txId, tx.height, tx.timestamp, tx.version);
   await addUtxos(mysql, txId, tx.outputs, heightlock);
 
   // remove inputs from utxo table
-  await removeUtxos(mysql, tx.inputs);
+  await removeUtxos(mysql, tx.inputs, txId);
```
pedroferreira1: Are we still removing the utxos?

pedroferreira1: Nevermind, I just noticed that the method name is still `removeUtxos` but we are only updating the rows right now. We should rename it (I've added the comment there).

DATABASE.md
pedroferreira1: Maybe we are saving only blocks in `blocks` table and only transactions in `transaction`? Is that right?

If that's the case, don't we need the timestamp of blocks? And maybe add a comment that the height of this table is the tx `first_block`, I thought it was to store the blocks height.

src/db.ts
```diff
@@ -1247,14 +1309,35 @@ export const getVersionData = async (mysql: ServerlessMysql): Promise<FullNodeVe
...
 
+/**
+ * Get block by height
+ *
+ * @param mysql - Database connection
+ * @param height - The height to query
+ *
+ * @returns The latest height
+ */
+export const getBlockByHeight = async (mysql: ServerlessMysql, height: number): Promise<Block> => {
```
pedroferreira1, line 1338: This return type `Promise<Block>` includes returning `null` in the case we call this method with a height we don't have in the db?

src/db.ts
```diff
@@ -1509,3 +1596,488 @@ export const getTxProposalInputs = async (
...
+  const results: DbSelectResult = await mysql.query(
+    `SELECT *
+       FROM \`tx_output\`
+      WHERE \`spent_by\` IN (?)`,
+    [txIds],
+  );
+
+  const utxos = [];
+  for (const result of results) {
+    const utxo: DbTxOutput = {
```
pedroferreira1, line 1755: We do this map at least two times, should we create a method for that in order to make it easier for future changes?

src/db.ts
```diff
@@ -1509,3 +1596,488 @@ export const getTxProposalInputs = async (
...
+ * Set a list of tx_outputs as unspent
+ *
+ * @param mysql - Database connection
+ * @param txOutputs - The list of tx_outputs to unspend
+ */
+export const unspendUtxos = async (
+  mysql: ServerlessMysql,
+  txOutputs: DbTxOutput[],
+): Promise<void> => {
+  const txIds = txOutputs.map((txOutput) => txOutput.txId);
```
pedroferreira1: I am not sure this is right, this method is used here: https://github.com/HathorNetwork/hathor-wallet-service/pull/71/files?file-filters%5B%5D=.json&file-filters%5B%5D=.ts#diff-f2a2f842c9891cbd8bd1777480db992084e7be4cd6abbc0f58d0f156bc6c0968R375

The `txOutputs` in the parameter are the ones that were spent but should not be spent anymore, right? Then we must search here by `tx_id` and `index` because we could have an output that must be unspend and another output that is still correct and both are from the same transaction, right?

src/db.ts
```diff
@@ -1509,3 +1596,488 @@ export const getTxProposalInputs = async (
...
+      \`unlocked_balance\`,
+      \`locked_balance\`,
+      \`unlocked_authorities\`,
+      \`locked_authorities\`,
+      \`timelock_expires\`,
+      \`transactions\`
+    )
+     SELECT address,
+            token_id,
+            SUM(\`value\`), -- unlocked_balance
```
pedroferreira1, line 1946: For an authority output the value is always 0?

src/db.ts
```diff
@@ -1509,3 +1596,488 @@ export const getTxProposalInputs = async (
...
+ * Retrieves a list of `AddressTotalBalance`s from a list of addresses
+ *
+ * @param mysql - Database connection
+ * @param addresses - The addresses to query
+ */
+export const fetchAddressTxHistorySum = async (
+  mysql: ServerlessMysql,
+  addresses: string[],
+): Promise<AddressTotalBalance[]> => {
+  const results: DbSelectResult = await mysql.query(
```
pedroferreira1, line 2069: Why don't you get this info from `address_balance` table?

src/commons.ts
pedroferreira1: I am thinking again about this and, each reorg call should be every ~30s if we are unlucky and this full binary search should be really fast even without optmization, not sure if it's valid.

src/commons.ts
```diff
@@ -259,3 +283,135 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+  const { height } = await searchForLatestValidBlock(mysql);
+
+  console.log(`Handling reorg. Our latest best block is ${height}`);
+
+  // remove blocks where height > latestValidBlock
+  await deleteBlocksAfterHeight(mysql, height);
+
+  // fetch all block transactions where height > latestValidBlock
+  const allTxsAfterHeight = await getTxsAfterHeight(mysql, height);
+  let txs: Tx[] = allTxsAfterHeight.filter((tx) => [
```
pedroferreira1, line 399: With this filter I notice that the `transaction` table also has all the blocks, just like the `blocks` table. Is it necessary?

src/commons.ts
pedroferreira1: @msbrogli they can be transactions if the transactions are spending outputs that were from reorged blocks, then they become voided as well.

src/commons.ts
```diff
@@ -259,3 +283,135 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+    console.log(`Removing WalletTxHistory from ${txs.length} transactions.`);
+    await removeWalletTxHistory(mysql, txs);
+    console.log(`Removing AddressTxHistory from ${txs.length} transactions.`);
+    await removeAddressTxHistory(mysql, txs);
+
+    const txOutputs: DbTxOutput[] = await getTxOutputs(mysql, txs); // "A" Outputs
+
+    // txOutputs might contain more than one output that spend from the same tx
+    const txIds = txOutputs.reduce((acc: Set<string>, utxo: DbTxOutput) => {
+      acc.add(utxo.spentBy);
```
pedroferreira1: Should we add only if `utxo.spentBy !== null`? Or is it safe to do like this?

src/commons.ts
```diff
@@ -259,3 +283,135 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+    }, new Set<string>());
+
+    console.log(`Setting ${txOutputs.length} tx_outputs as dirty.`);
+    // delete tx outputs:
+    await deleteUtxos(mysql, txOutputs);
+
+    removedUtxoList = [...removedUtxoList, ...txOutputs];
+
+    // get outputs that were spent in txOutputs
+    const spentOutputs: DbTxOutput[] = await getTxOutputsBySpent(mysql, [...txIds]);
```
pedroferreira1: Maybe I am misunderstanding this part but you are getting these outputs to mark them as `spent_by = null`, right? Right now you are getting the outputs from `txIds`, which are the ids of the transactions that are spending the removed outputs, i.e. you are getting the outputs that will be removed in the next iteration.

I think you should be getting the outputs that are being spent by txs:

```
const txsBeingDeletedIds = txs.map((tx) => tx.tx_id);
const spentOutputs: DbTxOutput[] = await getTxOutputsBySpent(mysql, [...txsBeingDeletedIds]);   // Now you are using outputs that might be deleted, are being spent by a tx that will be deleted, then should be marked as not used

... 

// And the unspendUtxos method has a relevant comment as well imo
```

Is that right?

src/commons.ts
```diff
@@ -259,3 +283,135 @@ export const maybeRefreshWalletConstants = async (mysql: ServerlessMysql): Promi
...
+
+  // get all remaining txs and set height = null (mempool)
+  const remainingTxs: Tx[] = await getTxsAfterHeight(mysql, height);
+  if (remainingTxs.length > 0) {
+    console.log(`Setting ${remainingTxs.length} unconfirmed transactions to the mempool (height = NULL).`);
+    await removeTxsHeight(mysql, remainingTxs);
+  }
+
+  // fetch all addresses affected by the reorg
+  const affectedAddresses = removedUtxoList.reduce((acc: Set<string>, utxo: DbTxOutput) => acc.add(utxo.address), new Set<string>());
```
pedroferreira1: I think we should add the utxos that were marked as `spent_by = null` in the `removedUtxoList` array as well, right? Maybe rename to `affectedUtxoList`. Because those utxos were not removed but they were spent but are not anymore, then the address balance changes also.

pedroferreira1 (approved):

pedroferreira1 (changes requested):
I miss clicked in "Approve" last time, just fixing this

src/db.ts
andreabadesso: I agree.

Done in 5d52f0d7f2e1c9caf915606954ce6e032a04c326

src/commons.ts
andreabadesso: I agree.

Done in f4dfc72e1b7a093a9b396f34060163227cd95f0f

src/commons.ts
andreabadesso: Yes, you are right

Fixed in f4dfc72e1b7a093a9b396f34060163227cd95f0f

src/commons.ts
andreabadesso: You are right.

Fixed in f4dfc72e1b7a093a9b396f34060163227cd95f0f

src/commons.ts
andreabadesso: Oops

Done in 7a81e7348869a003e6554578d02c8b7ef30d4843

src/commons.ts
andreabadesso: Done in another PR:
https://github.com/HathorNetwork/hathor-wallet-service/pull/76

src/db.ts
andreabadesso: Yes we should

Done in e411037c4ef31f42e6b4e57086c0250df2032001

src/db.ts
andreabadesso: Done in bd56d1f5c0b448167e7c8928135ef6454c9f157c

src/db.ts
andreabadesso: This is used to assert if the sum of `address_tx_history` is exactly equal the balance on the `address_balance` table

src/fullnode.ts
andreabadesso: Done in 76d3d66a2ccda0031e3c82182626d37e035ba990

src/utils.ts
andreabadesso: I agree, done in 0e3a602f3b2635343558948edac5578834634fe0

src/db.ts
andreabadesso: Yes

DATABASE.md
pedroferreira1: This was done in another PR

src/txProcessor.ts
```diff
@@ -132,10 +214,11 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
 
   // add outputs to utxo table
   markLockedOutputs(tx.outputs, now, heightlock !== null);
+  await addOrUpdateTx(mysql, txId, tx.height, tx.timestamp, tx.version);
   await addUtxos(mysql, txId, tx.outputs, heightlock);
 
   // remove inputs from utxo table
```
pedroferreira1: Update comment after method rename?

pedroferreira1 (approved):

src/txProcessor.ts
andreabadesso: Done in 73af9092af09e3a966ba715d19d7765d11850854

pedroferreira1: Still makes no sense to me the comment. 

`// remove inputs from tx_outputs table` should be `// mark tx_outputs used in the new tx as spent by`. What do you think?

andreabadesso: I agree with you, done in 7eb739884c9463e86771395c5bd217680b01d2e7

src/commons.ts
msbrogli: Another approach is to run an exponential search before running the binary search. Sync-v1 does exactly this for the reasons already pointed out by @pedroferreira1. I feel we can open an issue to do it in the future. What do you think?

msbrogli (approved):

PR #72, opened 2021-05-17
feature: update real time

r4mmer:


src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+});
+
+const validateAdminToken = (authHeader: string): boolean => {
+  const sanitizedToken = authHeader.replace(/Bearer /gi, '');
+  return sanitizedToken === process.env.ADMINTOKEN;
+};
+
+export const broadcast: APIGatewayProxyHandler = async (event) => {
+  if (process.env.IS_OFFLINE) {
+    console.log(event); // eslint-disable-line no-console
```
justgigio: Is this for serverless running offline? If so, what do you think of using some High-Order function for ApiGateway handlers? (I'm using annotations in python).

So you can have something like:
```javascript
const ApiGatewayHandler = (handler) => {
  return (event, context) => {
    if (process.env.IS_OFFLINE) { console.log(event); }
    // ...
    // You can even parse headers and params here. it's extreme useful when we need some special header 
    //   or we want to normalize query strings and body params
    // ...
    return handler(event, context);
  }
}
```

And export you function like that:
`export const broadcast = ApiGatewayHandler(realBroadcastFunction);`

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  const authHeader = event.headers && event.headers.Authorization;
+  if (!(authHeader && validateAdminToken(authHeader))) {
```
justgigio: this kind of validation can be easily done in the example i gave in the previous comment.

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+      statusCode: 403,
+      body: JSON.stringify({
+        success: false,
+        error: ApiError.FORBIDDEN,
+      }),
+    };
+  }
+
+  const payload = parseBody(event.body);
+  if (!payload) {
```
justgigio: And this validation as well.

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ message: 'sent' }),
+  };
+};
+
+// maybe allow disconnecting a single connection by id and not all wallet connections
+export const disconnect: APIGatewayProxyHandler = async (event) => {
+  const authHeader = event.headers && event.headers.Authorization;
+  if (!(authHeader && validateAdminToken(authHeader))) {
```
justgigio: i saw a lot of code like this here. Just reinforcing the first comment about High-Order Functions

src/ws/txNotify.ts
```diff
@@ -0,0 +1,123 @@
...
+  for (const evt of event.Records) {
+    const { value, error } = newTxbodySchema.validate(evt.body, {
+      abortEarly: false,
+      convert: true,
+    });
+
+    if (error) {
+      // extract error message
+      await closeRedisClient(redisClient);
+      return {
```
justgigio: This lambda will consume SQS, right? So what will happen when `400` is returned? Does it make sense to have a dead letter queue to reprocess? Or maybe a log for us to know that something is going wrong?

README.md
```diff
@@ -126,6 +134,57 @@ curl --header "Content-Type: application/json" \
...
+
+This action is idempotent, the lambda just responds with a `PONG` message.
+
+#### WebSocket Action: Join Wallet
+- Trigger: Client initiated
+- body: `{"action":"joinWallet", "wallet":"my-wallet-id"}`
+
+This action will subscribe the client to any updates of the wallet identified by the id on the body.
+
+#### WebSocker Action: New TX
```
pedroferreira1: Typo: Websocke**t**

README.md
```diff
@@ -126,6 +134,57 @@ curl --header "Content-Type: application/json" \
...
+- Trigger: SQS Event
+- When: A new tx is processed by the wallet-service
+- To: All wallets affected by the tx
+- body: The tx in JSON format
+
+#### WebSocker Action: New TX
```
pedroferreira1: Typo: Websocke**t**

Typo: **Update TX**

src/api/errors.ts
```diff
@@ -25,4 +25,6 @@ export enum ApiError {
   WALLET_NOT_FOUND = 'wallet-not-found',
   WALLET_NOT_READY = 'wallet-not-ready',
   WALLET_ALREADY_LOADED = 'wallet-already-loaded',
+  FORBIDDEN = 'forbidden',
```
pedroferreira1: Is this related to the admin token in the env file, right? Why do we have this? Shouldn't we integrate the wallet authorization here after is finished for the APIs?

src/txProcessor.ts
```diff
@@ -163,4 +163,6 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
   // update wallet_balance and wallet_tx_history tables
   const walletBalanceMap: StringMap<TokenBalanceMap> = getWalletBalanceMap(addressWalletMap, addressBalanceMap);
   await updateWalletTablesWithTx(mysql, txId, tx.timestamp, walletBalanceMap);
+
+  // TODO send new_tx (ws)event for all walletIds on addressWalletMap
```
pedroferreira1: This is the event to be sent to SQS? I think this is one of the most important parts of the PR, right?

src/redis.ts
```diff
@@ -0,0 +1,171 @@
...
+import redis from 'redis';
+import { promisify } from 'util';
+
+const redisConfig: RedisConfig = {
+  host: process.env.REDIS_HOST,
+  port: parseInt(process.env.REDIS_PORT, 10),
+  password: process.env.REDIS_PASSWORD,
+};
+
+export default redis;
```
pedroferreira1: Why do you export what you are importing?

r4mmer:
### Open discussions
- wallets: should we have a `joinWallet` action or send the `walletId` on the connection call?
    - the `joinWallet` lets unvalidated clients connect to the API GW.
    - `joinWallet` lets the client use a single connection to listen to multiple wallet events
    - Implementing auth will not be a problem on either one.
- wallets: should we validate the walletId? Look in the wallets table to check if exists?
- txNotify: The trigger should be from SQS or lambda invoke?
    - the caller would be a lambda for `onNewTxEvent` that is called by an invoke, so if the caller invokes as well the execution time of `onNewTx` would be doubled (as 2 lambdas are running at the same time until completion)
    - SQS lets the lambda be called asynchronously but introduces another resource, and who would provision the SQS? (The SQS arn must be on the serverless.yml, so it must exist before lambda deployment).
- channels: should we let the `join channel` action enter the PR? There is no application for it now, but it opens up for future applications.
- ws admin: should it be a http trigger, or no trigger and we invoke from the cli or sdk if needed
- ws admin: should we make multicast by address or tx_id? (currently only walletId is supported)
- ws admin: should we have a disconnect by `connectionId` or only by `walletId` (by walletId disconnects all connections of that wallet)

Did i forget anything?

src/redis.ts
```diff
@@ -0,0 +1,171 @@
...
+
+// export const client = redis.createClient(redisConfig);
```
pedroferreira1: Remove?

src/redis.ts
```diff
@@ -0,0 +1,171 @@
...
+ *  - same parameters: TTL is refreshed
+ *  - other params: updates connection keys
+ */
+export const initWsConnection = async (
+  // walletID: string,
+  client: redis.RedisClient,
+  connInfo: WsConnectionInfo,
+): Promise<string> => {
+  const setAsync = promisify(client.set).bind(client);
+  return setAsync(`walletsvc:conn:${connInfo.id}`, connInfo.url);
```
pedroferreira1: Maybe we should have a var with the prefix `walletsvc`

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// import AWS from 'aws-sdk';
```
pedroferreira1: Remove import?

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+const parseBody = (body: string) => {
+  try {
+    return JSON.parse(body);
+  } catch (e) {
+    return null;
+  }
+};
+
+const multicastSchema = Joi.object({
+  // addresses: Joi.array()
```
pedroferreira1: Remove?

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+  //   .items(Joi.string()),
+  wallets: Joi.array()
+    .items(Joi.string())
+    .required(),
+  payload: Joi.object().required(),
+});
+
+const disconnectSchema = Joi.object({
+  // addresses: Joi.array()
```
pedroferreira1: Remove?

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+    proms.push(wsGetWalletConnections(redisClient, walletId).then((connections) => {
+      const p = [];
+      connections.forEach((connInfo) => {
+        p.push(sendMessageToClient(connInfo, redisClient, payload));
+      });
+      return Promise.all(p);
+    }));
+  }
+
+  // maybe Promise.all().then(() => return)?
```
pedroferreira1: I prefer async/await syntax

src/ws/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+  const connID = event.requestContext.connectionId;
+  if (process.env.IS_OFFLINE === 'true') {
+    return {
+      id: connID,
+      url: 'http://localhost:3001',
+    };
+  }
+  const domain = event.requestContext.domainName;
+  const stage = event.requestContext.stage;
+  // return util.format(util.format('https://%s/%s', domain, stage));
```
pedroferreira1: Remove?

tests/utils.ts
```diff
@@ -635,3 +637,30 @@ export const checkVersionDataTable = async (mysql: ServerlessMysql, versionData:
...
+    multi.set(k, v);
+  }
+  multi.exec();
+};
+
+export const redisCleanup = (
+  client: RedisClient,
+): void => {
+  client.flushdb();
+  // client.keys('*', (err, keys) => {
```
pedroferreira1: Remove?

src/ws/txNotify.ts
```diff
@@ -0,0 +1,123 @@
...
+    .min(1)
+    .required(),
+  update: Joi.object({
+    tx_id: Joi.string().required(),
+    is_voided: Joi.boolean(),
+  })
+    .required(),
+});
+
+// const parseBody = (body: string) => {
```
pedroferreira1: Remove

src/ws/txNotify.ts
pedroferreira1: For what I understood this 400 error will be returned when a message with wrong body schema is received. In this case I believe we should discard those messages (we can log but not reprocess).

src/ws/txNotify.ts
```diff
@@ -0,0 +1,123 @@
...
+      await closeRedisClient(redisClient);
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ error: true, message: 'error' }),
+      };
+    }
+
+    const wallets = value.wallets;
+    const updateBody = value.update;
+    await Promise.all(wallets.map((wallet) => notifyWallet(redisClient, wallet, updateBody)));
```
pedroferreira1: Should we have a `type` key in the notify body message? For what I am seeing we are just sending the data in the ws message, right?

src/ws/joinChannel.ts
```diff
@@ -0,0 +1,150 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { APIGatewayProxyEvent } from 'aws-lambda';
+// import { APIGatewayProxyHandler } from 'aws-lambda';
```
pedroferreira1: Remove?

src/ws/connection.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// import AWS from 'aws-sdk';
```
pedroferreira1: Remove? and the one below

src/ws/connection.ts
```diff
@@ -0,0 +1,65 @@
...
+    await endWsConnection(redisClient, connInfo.id);
+    await closeRedisClient(redisClient);
+    return { statusCode: 200 };
+  }
+
+  if (routeKey === 'ping') {
+    return sendAndReturn(connInfo, 200, { message: 'PONG' }, redisClient);
+  }
+
+  // if (routeKey === '$default') {
```
pedroferreira1: We should remove all these commented codes

pedroferreira1 (changes requested):

pedroferreira1:
> wallets: should we have a joinWallet action or send the walletId on the connection call?

I prefer starting with the walletId on the connection call, even though it prevents out connection from listening to multiple wallets (we don't have it for now)

> wallets: should we validate the walletId? Look in the wallets table to check if exists?

Yes, I think so.

> txNotify: The trigger should be from SQS or lambda invoke?

I think using the SQS would be the best approach. The only downside is to have a new resource? If yes, I would go with that.

> channels: should we let the join channel action enter the PR? There is no application for it now, but it opens up for future applications.

I like it but I am afraid someone might explore this flooding our redis with fake connections. At least for the wallets we can validate if the wallet id exists.

> ws admin: should it be a http trigger, or no trigger and we invoke from the cli or sdk if needed

I am ok with triggering it with the cli for now. We don't have a regular use for that, I just think this should be clearly documented somewhere.

> ws admin: should we make multicast by address or tx_id? (currently only walletId is supported)

I think walletId is enough for now.

> ws admin: should we have a disconnect by connectionId or only by walletId (by walletId disconnects all connections of that wallet)

I think we should disconnect only by connectionId by default. Disconnect by walletId would disconnect all devices at the same time (mobile, desktop, headless, ...). We could support this but the default must be to disconnect by connectionId.

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+      }),
+    };
+  }
+
+  const wallets = value.wallets;
+  const payload = value.payload;
+
+  // for each wallet, get connections and send payload to each connection of each wallet
+  const redisClient = getRedisClient();
+  const proms = [];
```
andreabadesso: Suggestion:

Refactor to something like (untested):

```
const proms = wallets.reduce(async (acc, walletId) => {
    const connections = await wsGetWalletConnections(redisClient, walletId);
    const disconnectPromises = connections.map((connection) => disconnectClient(connection, redisClient));

    return [...acc, ...disconnectPromises];
}, []);
```

src/ws/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+        reject(err);
+      }
+      resolve(data);
+    },
+  );
+});
+
+export const disconnectClient = async (
+  connInfo: WsConnectionInfo,
```
andreabadesso: Maybe change the parameters order to have the `RedisClient` as the first argument to be consistent with `@src/redis` method signatures

src/ws/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+  client: RedisClient,
+): Promise<any> => new Promise((resolve, reject) => { // eslint-disable-line @typescript-eslint/no-explicit-any
+  const apigatewaymanagementapi = new AWS.ApiGatewayManagementApi({
```
andreabadesso: camelCase?

src/types.ts
```diff
@@ -550,3 +550,14 @@ export interface ApiResponse {
   success: boolean;
   message: string;
 }
+
```
andreabadesso, line 553: I think we should create a `types` directory and separate types into different files, those types here should be in `types/redis.ts`

What do you think?

src/ws/admin.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  const body = parseBody(event.body);
+  const { value, error } = multicastSchema.validate(body, {
+    abortEarly: false,
+    convert: false,
+  });
+
+  if (error) {
+    const details = error.details.map((err) => ({
```
andreabadesso: I know you are just being consistent with the other APIs that are doing this map to get the error details but I think it would be optimal to have a `getErrorDetails` method on utils and use it everywhere

Maybe create an issue to refactor this?

src/ws/joinChannel.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  await closeDbConnection(mysql);
+  await closeRedisClient(redisClient);
+  return { statusCode: 200 };
+};
+
+const joinWallet = async (
+  event: APIGatewayProxyEvent,
+  connInfo: WsConnectionInfo,
+  _mysql: ServerlessMysql,
```
andreabadesso: Why the underscore on `mysql` and `client`?

andreabadesso (changes requested):

src/ws/txNotify.ts
r4mmer: You're right, just checked the type for SQSHandler, the return type is void, so no return message is needed.

src/ws/admin.ts
r4mmer: For this one i think it's not needed, i prefer not having an http trigger and make the admin calls only by invoke (sdk, cli or aws console). So i'm cleaning this validation out of the handler.

src/api/errors.ts
r4mmer: You're right, i guess i didn't see this change when changing branches. I'm cleaning this out.

src/redis.ts
r4mmer: You're right, i'm cleaning this out.

src/ws/admin.ts
r4mmer: I think it's a good use for some validations, but we should see if we want to have this offline log first.

src/types.ts
r4mmer: I guess, but is there any other need for this, or redis is just a special case? if redis is a special case, i could put these types on `src/redis.ts` instead.

src/ws/joinChannel.ts
r4mmer: To not have conflicts with global mysql (and in the first version i had a global client as well)

src/txProcessor.ts
```diff
@@ -163,4 +164,19 @@ const addNewTx = async (tx: Transaction, now: number, blockRewardLock: number) =
   // update wallet_balance and wallet_tx_history tables
   const walletBalanceMap: StringMap<TokenBalanceMap> = getWalletBalanceMap(addressWalletMap, addressBalanceMap);
   await updateWalletTablesWithTx(mysql, txId, tx.timestamp, walletBalanceMap);
+
+  const queueUrl = process.env.NEW_TX_SQS;
+  if (!queueUrl) {
```
pedroferreira1: Shouldn't this be `if (queueUrl)`?

pedroferreira1 (changes requested):

src/txProcessor.ts
r4mmer: Done

src/ws/admin.ts
r4mmer: Changed to something along those lines, but there were a few modifications since the arrow function cannot be async (we cannot 'await' inside the arrow, but we can return promises).

src/ws/utils.ts
r4mmer: Done.

src/ws/utils.ts
r4mmer: Done.

src/ws/admin.ts
r4mmer: Since this call is not called from ApiGateway it doesn't need to respond with details of the error, so i removed this part, but maybe we should refactor this on another issue.

src/txProcessor.ts
r4mmer: Yes, i was thinking of `if (!queueUrl) return;` but ended up wrapping the sqs call on the it.

src/redis.ts
```diff
@@ -0,0 +1,160 @@
...
+ *  - broadcast
+ *  - send to wallet connections
+ *  - send to channel? (maybe wallet-`walletId` should be a channel)
+ *
+ *  If this function is called when the fields already exists:
+ *  - same parameters: TTL is refreshed
+ *  - other params: updates connection keys
+ */
+export const initWsConnection = async (
+  // walletID: string,
```
pedroferreira1: Remove?

src/redis.ts
```diff
@@ -0,0 +1,160 @@
...
+  do {
+    const reply = await scanAsync(cursor, 'MATCH', pattern);
+    cursor = reply[0];
+    found.push(...reply[1]);
+  } while (cursor !== '0');
+
+  return found;
+};
+
+// XXX: AsyncGenerator implementations of scanAll, maybe use these?
```
pedroferreira1: Remove?

src/ws/admin.ts
```diff
@@ -0,0 +1,125 @@
...
+  payload: Joi.object().required(),
+});
+
+const disconnectSchema = Joi.object({
+  connections: Joi.array()
+    .items(Joi.string())
+    .required(),
+});
+
+export const broadcast: Handler = async (event) => {
```
pedroferreira1, line 38: I don't think broadcast and multicast should be open APIs. We could have basic auth or just something to protect from a bad usage.

src/ws/join.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const handler = async (
+  event: APIGatewayProxyEvent,
+): Promise<void> => {
+  const redisClient = getRedisClient();
+  const routeKey = event.requestContext.routeKey;
+  const connInfo = connectionInfoFromEvent(event);
+
+  if (routeKey === 'join') {
```
pedroferreira1: Is there any other possible `routeKey`? Should we return any error if it's different than join?

src/ws/join.ts
```diff
@@ -0,0 +1,99 @@
...
+      success: false,
+      message: 'Invalid parameters',
+    });
+    return;
+  }
+
+  const walletId = value.id;
+
+  // validate walletID
+  // verify ownership of wallet
```
pedroferreira1: This is a TODO, right? We are verifying if the wallet exists, now we must verify if your are the wallet owner after integrating with auth, right?

src/ws/admin.ts
r4mmer: I just saw that i didn't remove the triggers from serverless, it should not have any triggers, so it would be invoke only, a caller must have an invoke permission (like `src.txProcessor.addNewTx` that is called by the daemon).

src/ws/join.ts
r4mmer: The API Gateway takes care of routing actions to their lambdas, this would be good if we had more `routeKey`'s pointing to this lambda, but we just have the "join".

src/ws/join.ts
r4mmer: Yes, for now we just verify that the wallet exists on the database.

src/redis.ts
r4mmer: Done

src/redis.ts
r4mmer: Done

src/ws/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+export const sendMessageToClient = async (
+  client: RedisClient,
+  connInfo: WsConnectionInfo,
+  payload: any, // eslint-disable-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
+): Promise<any> => { // eslint-disable-line @typescript-eslint/no-explicit-any
+  const apiGwClient = new AWS.ApiGatewayManagementApi({
+    apiVersion: '2018-11-29',
+    endpoint: connInfo.url,
+  });
+  // AWS.Request.promise() will make the request and return a thenable with the response
```
pedroferreira1, line 38: Didn't understand this "AWS.Request.promise() will make the request and return a thenable with the response". Is 'thenable' a typo?

src/ws/join.ts
pedroferreira1: You've removed both comments. I think you can keep the comment that we validate that the wallet exists on the db and another comment with TODO to verify ownership

pedroferreira1 (approved):

src/ws/utils.ts
r4mmer: 'thenable' is a way to refer to `Promise` or other objects that have the `.then()` function. A way to see this is that all promises are thenables but not all thenables are promises. I use this nomenclature because it's what the [aws-sdk docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#promise-property) use.

src/ws/join.ts
r4mmer: Done.

andreabadesso (approved):

PR #74, opened 2021-06-01
feat: handle voided transactions

andreabadesso:


src/commons.ts
```diff
@@ -378,6 +378,55 @@ export const handleVoidedTxList = async (mysql: ServerlessMysql, txs: Tx[]): Pro
   return [newTxs, affectedUtxoList];
 };
 
+/**
+ * Handles a reorg by finding the last valid block on the service's database and
```
pedroferreira1: The docstring (including parameters and return) is wrong

src/commons.ts
```diff
@@ -378,6 +378,55 @@ export const handleVoidedTxList = async (mysql: ServerlessMysql, txs: Tx[]): Pro
...
+    txs = newTxs;
+    affectedUtxoList = [...affectedUtxoList, ...newAffectedUtxoList];
+  }
+
+  // fetch all addresses affected by the voided tx
+  const affectedAddresses = affectedUtxoList.reduce((acc: Set<string>, utxo: DbTxOutput) => acc.add(utxo.address), new Set<string>());
+
+  console.log(`Rebuilding balances for ${affectedAddresses.size} addresses.`);
+  if (affectedAddresses.size > 0) {
+    const addresses = [...affectedAddresses];
```
pedroferreira1, line 406: Why do you recreate the list again here?

tests/txProcessor.test.ts
```diff
@@ -266,3 +267,69 @@ test('receive token creation tx', async () => {
...
+  // void the transaction
+  await txProcessor.handleVoidedTx(tx);
+
+  // both utxos should be voided
+  await expect(checkUtxoTable(mysql, 3, txId2, 0, token, addr, 2000, 0, null, null, false, null, true)).resolves.toBe(true);
+  await expect(checkUtxoTable(mysql, 3, txId2, 1, token, 'other', 500, 0, null, null, false, null, true)).resolves.toBe(true);
+  // the original utxo should not be voided and should not have been spent
+  await expect(checkUtxoTable(mysql, 3, txId1, 0, token, addr, 2500, 0, null, null, false, null, false)).resolves.toBe(true);
+
+  await expect(checkAddressBalanceTable(mysql, 1, addr, token, 2500, 0, 0, 1)).resolves.toBe(true);
```
pedroferreira1, line 350: Maybe we should add a tx3 in this test that spends tx2. With tx2 as voided, tx3 should become voided as well and tx1 unspent. What do you think?

pedroferreira1 (changes requested):

src/commons.ts
andreabadesso: `affectedAddresses` is a `Set`, I need to convert it to an `array`

msbrogli, line 406: Maybe log which address is being recalculated?

msbrogli (approved):

src/commons.ts
andreabadesso: I agree, the list probably won't be large so we can safely log

Done in d38c66bf0d8304c4d694e4bf499c4b17300dc994

src/commons.ts
andreabadesso: Done in c6fa85e97944dde34ef931a0d8d50cb516b825df

tests/txProcessor.test.ts
andreabadesso: I agree!

Done in 27d47abafca129b3c2b22d03757262eb18f44208

pedroferreira1 (approved):

PR #76, opened 2021-06-07
refactor: using the transactions table to handle blocks

andreabadesso:


src/db.ts
```diff
@@ -1309,7 +1309,17 @@ export const getVersionData = async (mysql: ServerlessMysql): Promise<FullNodeVe
...
-  const results: DbSelectResult = await mysql.query('SELECT MAX(`height`) as value FROM `blocks`');
+  const blockVersion = [
+    constants.BLOCK_VERSION,
+    constants.MERGED_MINED_BLOCK_VERSION,
+  ];
+  const results: DbSelectResult = await mysql.query(
+    `SELECT MAX(\`height\`) AS value
+       FROM \`transaction\`
+      WHERE version
+         IN (?)`, [blockVersion],
```
msbrogli: Double check that there is a hash index for the `version` field.

msbrogli (approved):
I've been wondering whether we should store blocks in a separate table or not for a while. It seems you reached the conclusion that a single table is easier to maintain, right? You should've suffered from handling two tables and decided to move to a single table approach. What pain made you decide to change?

pedroferreira1:
> I've been wondering whether we should store blocks in a separate table or not for a while. It seems you reached the conclusion that a single table is easier to maintain, right? You should've suffered from handling two tables and decided to move to a single table approach. What pain made you decide to change?

Before this refactor we were storing the blocks duplicated, in the blocks and transactions table. I suggested saving only once because it was not useful having the same information twice. We discussed about the queries in the future if we end up having many tps in the network but I still believe it won't be bad for us.

src/db.ts
andreabadesso: Added the index on `DATABASE.md` on commit 68ca86212fdc0ec35f71656b77bd678ba3ae69d0

andreabadesso: What made me decide to remove the `blocks` database was the fact that we already have this information on the `transactions` table, so we would be storing the same information twice.

With a properly indexed table, I see no reason to separate it on another table for now

DATABASE.md
```diff
@@ -156,7 +150,10 @@ CREATE TABLE `transaction` (
   `voided` boolean NOT NULL DEFAULT false,
   `height` int unsigned DEFAULT NULL,
```
pedroferreira1, line 152: Can we add a comment in this column to say that if it's a block, it's the height of the block and if it's a transaction, it's the height of the first_block?

src/db.ts
```diff
@@ -1308,7 +1308,17 @@ export const getVersionData = async (mysql: ServerlessMysql): Promise<FullNodeVe
  * @returns The latest height
  */
 export const getLatestHeight = async (mysql: ServerlessMysql): Promise<number> => {
-  const results: DbSelectResult = await mysql.query('SELECT MAX(`height`) as value FROM `blocks`');
+  const blockVersion = [
```
pedroferreira1: We have this array three times in this same file, can we create a single variable that is used in the three methods?

pedroferreira1 (changes requested):

src/db.ts
andreabadesso: Done in 54b7fd7d6ddb50e47b013c2da2ae242658304d1e

DATABASE.md
andreabadesso: Done in 7a4388591656421b396b5c710fc89d27efe1bae3

pedroferreira1 (approved):

PR #78, opened 2021-06-17
chore: create a SQS table automatically with an iam role on deploy using serverless.yml

andreabadesso:


luislhl (approved):

serverless.yml
```diff
@@ -31,10 +38,17 @@ provider:
   environment:
     AWS_NODEJS_CONNECTION_REUSE_ENABLED: 1
     SERVICE_NAME: ${self:service}
-    AWS_REGION: ${self:provider.region}
+    # AWS_REGION: ${self:provider.region}
```
pedroferreira1: Why comment this? If it's not needed we could remove it.

pedroferreira1 (approved):

msbrogli (approved):

serverless.yml
andreabadesso: I agree, done in bd85c1378230e4ee41ebba7862802ca53256c2ef

serverless.yml
```diff
@@ -129,11 +143,19 @@ functions:
           route: join
   wsTxNotifyNew:
     handler: src/ws/txNotify.onNewTx
+    events:
+      - sqs:
+          arn:
+            Fn::GetAtt:
+              - WalletServiceNewTxQueue
```
luislhl, line 150: I think we should be adding the stage suffix here too

serverless.yml
```diff
@@ -31,10 +39,16 @@ provider:
...
     DB_ENDPOINT: ${env:DB_ENDPOINT, 'rds-endpoint-TODO'}
     DB_NAME: ${self:service}-${env:STAGE}
     DB_USER: ${env:STAGE}-user
+    NEW_TX_SQS: { Ref: WalletServiceNewTxQueue }
+  iamRoleStatements:
+    - Effect: Allow
+      Action:
+        - sqs:*
+      Resource:
+        Fn::GetAtt: [ WalletServiceNewTxQueue, Arn ]
```
luislhl, line 51: I think we should be adding the stage suffix here too

luislhl (changes requested):
See previous comments

serverless.yml
andreabadesso: I tested it, this is the resource name and not the queue name, so it will get the queue name (with the stage) from the resource

serverless.yml
andreabadesso: Same as above

msbrogli (approved):

luislhl (approved):

PR #79, opened 2021-06-17
fix: returning correct status code on txprocessor error

andreabadesso:


r4mmer (approved):

src/txProcessor.ts
```diff
@@ -98,17 +98,29 @@ export const onNewTxEvent = async (event: SQSEvent): Promise<APIGatewayProxyResu
  * This is a lambda function that should be invoked using the aws-sdk.
  */
 export const onNewTxRequest: APIGatewayProxyHandler = async (event) => {
-  const now = getUnixTimestamp();
-  const blockRewardLock = parseInt(process.env.BLOCK_REWARD_LOCK, 10);
+  try {
```
pedroferreira1: This `try` is to catch errors inside `addNewTx`, right? Maybe we could have the `try...catch` block only for this method. What do you think?

andreabadesso: I agree, done in e8609a1984fba384831efbc78ecb0d1480226dfe

pedroferreira1 (approved):

PR #81, opened 2021-06-21
feat: Mempool sync and "out-of-order" txs

r4mmer:
Prepare for mempool sync and make any txs that spend utxos which we don't have to wait for those utxos.
Make tx processing from SQS instead of invoke

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -2086,3 +2086,98 @@ export const fetchAddressTxHistorySum = async (
     transactions: result.transactions as number,
   }));
 };
+
+/**
+ * Find all transactions being spent that we do not have.
+ *
+ * @param mysql - Database connection
+ * @param txInputs - List of inputs spent by the tx
+ * @returns A list with the missing UTXOs, if any
+ */
+export const checkForMissingTxs = async (mysql: ServerlessMysql, txInputs: TxInput[]): Promise<string[]> => {
+  const inputs = [];
+  for (const txIn of txInputs) {
+    const input: IWalletInput = {
+      txId: txIn.tx_id as string,
+      index: txIn.index as number,
+    };
+    inputs.push(input);
+  }
+
+  const utxos: DbTxOutput[] = await getUtxos(mysql, inputs);
+
+  if (inputs.length === utxos.length) return [];
+
+  // XXX: Should we check if tx input index match utxo as well?
+
+  const missing = new Set(inputs.map((input) => input.txId));
+  for (const utxo of utxos) {
+    missing.delete(utxo.txId);
+  }
+  return Array.from(missing);
+};
+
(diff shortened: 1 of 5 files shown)


src/db.ts
```diff
@@ -2086,3 +2086,98 @@ export const fetchAddressTxHistorySum = async (
...
+    inputs.push(input);
+  }
+
+  const utxos: DbTxOutput[] = await getUtxos(mysql, inputs);
+
+  if (inputs.length === utxos.length) return [];
+
+  // XXX: Should we check if tx input index match utxo as well?
+
+  const missing = new Set(inputs.map((input) => input.txId));
```
andreabadesso, line 2111: I think we need to use `[input.txId, input.index]`, right?

src/db.ts
```diff
@@ -2086,3 +2086,98 @@ export const fetchAddressTxHistorySum = async (
...
+              LEFT OUTER JOIN \`unconfirmed_tx_missing\`
+              ON \`unconfirmed_tx\`.\`tx_id\` = \`unconfirmed_tx_missing\`.\`spent_by\`
+              WHERE \`unconfirmed_tx_missing\`.\`spent_by\` IS NULL`,
+    },
+  );
+  // No need to nest tables because we only return the fields on one of the tables
+  // https://github.com/mysqljs/mysql#joins-with-overlapping-column-names
+
+  // ...then delete and return them
+  for (let i = 0; i < doneWaiting.length; i++) {
```
andreabadesso: Maybe delete in bulk using `DELETE ... WHERE tx_id IN doneWaiting.map((tx) => tx.tx_id);` ?

src/txProcessor.ts
```diff
@@ -261,16 +272,29 @@ export const addNewTx = async (tx: Transaction, now: number, blockRewardLock: nu
...
-
   const sqs = new AWS.SQS({ apiVersion: '2012-11-05' });
+
+  // XXX: maybe put this after WS?
+  // This tx has been processed, check if any txs are waiting for it!
+  const txReprocess = await clearMissingTx(mysql, txId);
+  const promises = [];
+  txReprocess.forEach((readyTx) => {
+    const params = {
+      MessageBody: readyTx.data, // already in json string format, maybe use JSON.parse?
```
andreabadesso: I think it is safe to just pass it along (without JSON.stringify(JSON.parse))) because we are already validating the schema before inserting it into the database

andreabadesso (changes requested):

src/db.ts
r4mmer: I don't think we need the index, since processing the tx with the waited id is all we need to reprocess the current "out-of-order" tx.

In the event of the "out-of-order" tx is waiting for index 2 but when processing the waited tx we find out if has only 1 output, it should be up to the `addNewTx` method to verify that the TX is invalid.

src/db.ts
r4mmer: Done.

Issue #75, opened 2021-06-03
[Design] New token actions API format


andreabadesso:
# Summary
[summary]: #summary

The idea is that the wallet-lib will be responsible for querying the wallet-service for utxos and compose the transaction locally before sending it again to the wallet-service to lock the utxos using the `txProposalCreate` lambda function.

To be able to do this, we need to develop an API to properly filter the utxos on the wallet-service database.

We also need to refactor the `txProposalCreate` API to accept a txHex instead of a partial transaction body (with inputs and outputs).

This is currently implemented in https://github.com/HathorNetwork/hathor-wallet-service/pull/82

# Motivation
[motivation]: #motivation

The motivation for this new token actions format is to make the API more generic to be able to, in the future, handle multi-sig transactions and have "draft" transactions that will be updated by multiple users before being sent.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

This implementation will consist of two parts, an API with multiple filters to query UTXOs and a refactor on the `txProposalCreate` to accept a `txHex` with a full transaction instead of a partial transaction body with only outputs or partial inputs.

The order of a tx creation request will be:

1. The wallet-lib will query the wallet-service for UTXOs matching the user needs
2. The wallet-lib will assemble a transaction and send it serialized in a `txHex`
3. The wallet service will decode de-serialize the transaction and validate its contents
4. The wallet service will lock the transaction inputs and store a new `txProposal` on the database
5. The wallet-lib will call the `txProposalSend` function to send the stored `txProposal` using its UUID
6. The wallet-service will validate if the transaction is the same as the one sent in `txProposalCreate` and push it to the full-node

In step 3, these are the validations we are doing on `txProposalCreate`:

* Too many outputs
* Too many inputs
* User not authenticated
* Wallet not ready
* Wallet not found
* Utxos not found
* Utxos not in the user's wallet
* Utxos already used in another tx proposal

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

## Filter Utxos API

`GET /utxos`

Possible query parameters:
```js
{
  addresses: Joi.array()
    .items(Joi.string().alphanum())
    .min(1)
    .required(),
  tokenId: Joi.string().default('00'),
  authority: Joi.number().default(0).integer().positive(),
  ignoreLocked: Joi.boolean().optional(),
  biggerThan: Joi.number().integer().positive(),
  smallerThan: Joi.number().integer().positive(),
[...]

PR #82, opened 2021-06-21
feat: token actions utxo api implementation

andreabadesso:
Design at https://github.com/HathorNetwork/hathor-wallet-service/issues/75


src/db.ts
```diff
@@ -2087,3 +2092,109 @@ export const fetchAddressTxHistorySum = async (
...
+  const results: DbSelectResult = await mysql.query(
+    `SELECT *
+       FROM \`tx_output\`
+      WHERE \`address\`
+         IN (?)
+        AND \`token_id\` = ?
+        AND \`authorities\` = ?
+        ${finalFilters.ignoreLocked ? 'AND `locked` = FALSE' : ''}
+        ${finalFilters.authority === 0 ? 'AND value < ?' : ''}
+        ${finalFilters.authority === 0 ? 'AND value > ?' : ''}
```
r4mmer, line 2133: `finalFilters.authority` twice. I think it should be `smallerThan` and `biggerThan`, right?

src/db.ts
```diff
@@ -2087,3 +2092,109 @@ export const fetchAddressTxHistorySum = async (
...
+        AND \`tx_proposal\` IS NULL
+        AND \`voided\` = FALSE
+        AND \`spent_by\` IS NULL
+   ORDER BY \`value\` DESC
+        ${finalFilters.maxUtxos ? 'LIMIT ?' : ''}
```
r4mmer, line 2138: Won't this throw an error? if `maxUtxos` exist and its not "falsy" it will be ok, but if not, there will be 5 `?` and 6 parameters.

src/db.ts
```diff
@@ -2087,3 +2092,109 @@ export const fetchAddressTxHistorySum = async (
...
+  txProposalId: string,
+  name: string,
+  symbol: string,
+): Promise<void> => {
+  const entry = [
+    [txProposalId, name, symbol],
+  ];
+
+  await mysql.query(
+    'INSERT INTO `tx_proposal_token_info` VALUES ?',
```
r4mmer: I think that if entry is a list of values the columns should be specified on the query, so we should either add the columns to the query or use a list of objects.

src/db.ts
```diff
@@ -2087,3 +2092,109 @@ export const fetchAddressTxHistorySum = async (
...
+ * @param name - The token name
+ * @param symbol - The token symbol
+ */
+export const addTxProposalTokenInfo = async (
+  mysql: ServerlessMysql,
+  txProposalId: string,
+  name: string,
+  symbol: string,
+): Promise<void> => {
+  const entry = [
```
r4mmer: The query below uses `[ entry ]`, so this shouldn't be a list of entries, right?

src/db.ts
```diff
@@ -2087,3 +2092,109 @@ export const fetchAddressTxHistorySum = async (
...
+ * @param txProposalId - The transaction proposal id
+ * @returns Information about the new token
+ */
+export const getTxProposalTokenInfo = async (
+  mysql: ServerlessMysql,
+  txProposalId: string,
+): Promise<TxProposalTokenInfo> => {
+  const results: DbSelectResult = await mysql.query(
+    'SELECT * FROM `tx_proposal_token_info` WHERE `tx_proposal_id` = ?',
+    [txProposalId],
```
r4mmer, line 2185: For a single parameter we don't need the list.

from https://github.com/mysqljs/mysql#performing-queries 

> If the query only has a single replacement character (?), and the value is not null, undefined, or an array, it can be passed directly as the second argument to .query

r4mmer (changes requested):

src/db.ts
andreabadesso: No, that is correct, we are only checking value if we are not filtering for authority utxos as authority utxos will always have value = 0

src/db.ts
andreabadesso: No, it will just ignore the 6th parameter, I think it is safe to do that

src/db.ts
andreabadesso: This query was removed entirely in the latest refactor

src/db.ts
andreabadesso: This query was removed entirely in the latest refactor

src/db.ts
andreabadesso: This query was entirely removed in the latest refactor

DATABASE.md
```diff
@@ -73,6 +73,7 @@ CREATE TABLE `tx_proposal` (
   `id` varchar(36) NOT NULL,
   `wallet_id` varchar(64) NOT NULL,
   `status` enum('open','sent','send_error','cancelled') NOT NULL,
+  `version` int unsigned NOT NULL DEFAULT 1,
```
pedroferreira1: Do we still need this field?

DATABASE.md
```diff
@@ -148,6 +157,9 @@ CREATE TABLE `transaction` (
 );
 
 CREATE INDEX transaction_version_idx USING HASH ON `transaction`(`version`);
+CREATE INDEX tx_output_heightlock_idx USING HASH ON `tx_output`(`heightlock`);
```
pedroferreira1, line 141: We've already created those indexes in dev, or not?

DATABASE.md
```diff
@@ -82,12 +83,20 @@ CREATE TABLE `tx_proposal_outputs` (
   `tx_proposal_id` varchar(36) NOT NULL,
   `index` tinyint unsigned NOT NULL,
   `address` varchar(34) NOT NULL,
-  `token_id` varchar(64) NOT NULL,
+  `token_id` varchar(64), -- Can be null because it might be a token creation action
```
pedroferreira1: The first file is a migration that deletes this table. Why are we adding those fields here?

db/migrations/20210718180539-add_tx_proposal_version.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('tx_proposal', 'version', {
```
pedroferreira1: Why do we need it?

serverless.yml
```diff
@@ -123,6 +122,24 @@ functions:
...
+          request:
+            parameters:
+              querystrings:
+                addresses: true
+                tokenId: true
+                authority: true
+                ignoreLocked: true
+                biggerThan: true
+                smallerThan: true
+                maxUtxos: true
```
pedroferreira1: We also have txId and index as optional parameters, right? How is it working if they are not here? Why do we need this configuration?

DATABASE.md
```diff
@@ -82,12 +83,20 @@ CREATE TABLE `tx_proposal_outputs` (
...
+  `token_data` int unsigned NOT NULL,
   `value` bigint DEFAULT NULL,
   `timelock` int unsigned DEFAULT NULL,
   PRIMARY KEY (`tx_proposal_id`,`index`)
 );
 
+CREATE TABLE `tx_proposal_token_info` (
```
pedroferreira1: Why do we need this table?

src/types.ts
```diff
@@ -614,3 +616,29 @@ export type WalletProxyHandler = (
...
+  USE_LARGER_UTXOS = 'use-larger-utxos',
+}
+
+export interface IWalletInsufficientFunds {
+  tokenId: string;
+  requested: number;
+  available: number;
+}
+
+export interface TxProposalTokenInfo {
```
pedroferreira1: I haven't seen this being used.

src/types.ts
```diff
@@ -614,3 +616,29 @@ export type WalletProxyHandler = (
...
+  biggerThan?: number;
+  smallerThan?: number;
+  maxUtxos?: number;
+}
+
+export enum InputSelectionAlgo {
+  USE_LARGER_UTXOS = 'use-larger-utxos',
+}
+
+export interface IWalletInsufficientFunds {
```
pedroferreira1: Where's it being used?

src/types.ts
```diff
@@ -614,3 +616,29 @@ export type WalletProxyHandler = (
...
+  addresses: string[];
+  tokenId?: string;
+  authority?: number;
+  ignoreLocked?: boolean;
+  biggerThan?: number;
+  smallerThan?: number;
+  maxUtxos?: number;
+}
+
+export enum InputSelectionAlgo {
```
pedroferreira1: Where's it being used?

src/types.ts
```diff
@@ -614,3 +616,29 @@ export type WalletProxyHandler = (
   context?: Context,
   callback?: Callback<APIGatewayProxyResult>
 ) => Promise<APIGatewayProxyResult>;
+
+export interface IFilterUtxo {
```
pedroferreira1, line 621: We also have `txId` and `index` as optional parameters here, right?

tests/db.test.ts
```diff
@@ -967,35 +969,76 @@ test('updateWalletLockedBalance', async () => {
   await expect(checkWalletBalanceTable(mysql, 3, wallet1, tokenId, 25, 5, now, 5, 0b11, 0b01)).resolves.toBe(true);
 });
 
-test('getLatestHeight and deleteBlocksAfterHeight', async () => {
+test('updateTx should add height to a tx', async () => {
```
pedroferreira1, line 996: I didn't understand this test. Where are we checking the height?

tests/db.test.ts
```diff
@@ -1457,3 +1482,80 @@ test('markAddressTxHistoryAsVoided', async () => {
...
+  // limit to 2 utxos, should return the largest 2 ordered by value
+  utxos = await filterUtxos(mysql, { addresses: [addr1, addr2], tokenId, maxUtxos: 2 });
+  expect(utxos).toHaveLength(2);
+  expect(utxos[0]).toStrictEqual({
+    txId: txId2, index: 2, tokenId, address: addr2, value: 1500, authorities: 0, timelock: null, heightlock: null, locked: true,
+  });
+  expect(utxos[1]).toStrictEqual({
+    txId: txId2, index: 1, tokenId, address: addr1, value: 1000, authorities: 0, timelock: null, heightlock: null, locked: false,
+  });
+});
```
pedroferreira1, line 1627: Should we add a test for `filterUtxos` that filters by `txId` and `index`?

pedroferreira1: Nevermind. It's already being tested in the other test file

src/api/utxos.ts
```diff
@@ -0,0 +1,162 @@
...
+    .min(1)
+    .optional(),
+  tokenId: Joi.string().default('00'),
+  authority: Joi.number().default(0).integer().positive(),
+  ignoreLocked: Joi.boolean().optional(),
+  biggerThan: Joi.number().integer().positive().default(-1),
+  smallerThan: Joi.number().integer().positive().default(constants.MAX_OUTPUT_VALUE + 1),
+  maxUtxos: Joi.number().integer().positive().default(constants.MAX_OUTPUTS),
+  txId: Joi.string().optional(),
+  index: Joi.number().optional(),
```
pedroferreira1: Add `min(0)`?

src/api/utxos.ts
```diff
@@ -0,0 +1,162 @@
...
+});
+
+/*
+ * Filter utxos
+ *
+ * This lambda is called by API Gateway on POST /filter_utxos
```
pedroferreira1: It's a GET /wallet/utxos

src/api/utxos.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    const denied = await validateAddresses(walletAddresses, [utxo.address]);
+
+    if (denied.length > 0) {
+      // the requested utxo does not belong to the user's wallet.
+      return closeDbAndGetError(mysql, ApiError.FORBIDDEN);
+    }
+
+    // check if the utxo is a member of the user's wallet
```
pedroferreira1: This comment should be above, right? Here you map each utxo to add the address path to it.

src/api/utxos.ts
```diff
@@ -0,0 +1,162 @@
...
+});
+
+/**
+ * Returns a new list of utxos with the addressPaths for each utxo
+ *
+ * @param walletAddress - A list of addresses for the user's wallet
+ * @param utxos - A list of utxos to map
+ * @returns A list with the mapped utxos
+ */
+export const mapUtxosWithPath = (walletAddresses: AddressInfo[], utxos: DbTxOutput[]) => utxos.map((utxo) => {
```
pedroferreira1: Missing return type

src/api/utxos.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Confirm that the requested addresses belongs to the user's wallet
+ *
+ * @param walletId - The user wallet id
+ * @param addresses - List of addresses to validate
+ * @returns A list with the denied addresses, if any
+ */
+export const validateAddresses = async (walletAddresses: AddressInfo[], addresses: string[]): Promise<string[]> => {
+  const flatAddresses = walletAddresses.map((walletAddress) => walletAddress.address);
```
pedroferreira1, line 150: I am thinking if it's worth it to have another type `AddressString` (or something like that), which is only the address. Everywhere you are using this `walletAddresses` you need to map them to the `.address`.

src/api/utxos.ts
```diff
@@ -0,0 +1,162 @@
...
+    throw new Error('Utxo address not in user\'s wallet');
+  }
+  const addressPath = `m/44'/${constants.HATHOR_BIP44_CODE}'/0'/0/${addressDetail.index}`;
+  return { ...utxo, addressPath };
+});
+
+/**
+ * Confirm that the requested addresses belongs to the user's wallet
+ *
+ * @param walletId - The user wallet id
```
pedroferreira1: This is wrong docstring, right?

src/db.ts
```diff
@@ -1363,7 +1409,7 @@ export const getLatestHeight = async (mysql: ServerlessMysql): Promise<number> =
          IN (?)`, [BLOCK_VERSION],
   );
 
-  if (results.length > 0) {
+  if (results.length > 0 && results[0].value !== null) {
```
pedroferreira1, line 1418: Why did you add this? What's this `value` in transaction table?

src/db.ts
```diff
@@ -1492,9 +1538,10 @@ export const createTxProposal = async (
   mysql: ServerlessMysql,
   txProposalId: string,
   walletId: string,
+  version: number,
```
pedroferreira1: Do we still need this?

src/db.ts
```diff
@@ -2129,3 +2083,91 @@ export const fetchAddressTxHistorySum = async (
...
+        AND \`voided\` = FALSE
+        AND \`spent_by\` IS NULL
+   ORDER BY \`value\` DESC
+        ${finalFilters.maxUtxos ? 'LIMIT ?' : ''}
+       `,
+    [
+      finalFilters.addresses,
+      finalFilters.tokenId,
+      finalFilters.authority,
+      finalFilters.smallerThan,
```
pedroferreira1: If `authority` is !== 0 and `maxUtxos` need to be filtered, this query will still work? I think the array of values should change dynamically for it to work, or am I missing something?

src/db.ts
```diff
@@ -2129,3 +2083,91 @@ export const fetchAddressTxHistorySum = async (
...
+      finalFilters.biggerThan,
+      finalFilters.maxUtxos,
+    ],
+  );
+
+  const utxos: DbTxOutput[] = results.map((utxo) => ({
```
pedroferreira1: I think you are missing 2 fields here, comparing to the `getUtxo` method:

```
    txProposalId: result.tx_proposal as string,
    txProposalIndex: result.tx_proposal_index as number,
```

Because it's a big mapping and it's used more than once, maybe it's worth it creating a method for that.

pedroferreira1 (changes requested):

DATABASE.md
andreabadesso: We don't. This is a leftover from the old `txProposalCreate` API. Fixed in 5b56c84f9f110340d8e32c0179c0f594c6b84632

DATABASE.md
andreabadesso: We shouldn't. Fixed in aaedf7762e104ed406a6c19f9121ead67d178b79

serverless.yml
andreabadesso: We don't need it. I was testing those parameters and forgot to remove it

Fixed in 7164730cdf738ec4fd951e6db3c2c8abf33e766d

src/api/utxos.ts
andreabadesso: Done in f6dc6531545249cde00b9fb746917a072eb6b7b5

DATABASE.md
andreabadesso: We did, but didn't add to `DATABASE.md`

src/api/utxos.ts
andreabadesso: Fixed in 174b95b34d9ac4141247f57db4188bdb8a18debf

src/api/utxos.ts
andreabadesso: Done in f6dc6531545249cde00b9fb746917a072eb6b7b5

src/api/utxos.ts
andreabadesso: You are right. Fixed in e3d59edd89cda5c3d0baa87f5b8bbfe017ac33ff

src/api/utxos.ts
andreabadesso: Done in cfef8693f1deccc0e88d40d9eb05ec4ce31d935b

src/api/utxos.ts
andreabadesso: Indeed.

Fixed in c67a656f49fda331dc39af7e88c12243586426e7

src/db.ts
andreabadesso: Because `MAX(height)` will always return a row in the query, but with NULL if nothing is found.

`value` is the result of `MAX(height)`:

`SELECT MAX(height) AS` **value**

src/db.ts
andreabadesso: It will still work, it will just ignore the extra parameters in the array

src/db.ts
andreabadesso: I agree!

Done in 8eaca4920335c4670e02c388667711626e1147f6

src/types.ts
andreabadesso: Yes!

Fixed in 7c1d205b458357fcab3d65280840f99a42fc7850

src/types.ts
andreabadesso: It is not anymore, removed it in 7c1d205b458357fcab3d65280840f99a42fc7850

src/types.ts
andreabadesso: It is not.

Fixed in e23e65aa3a26fbfb4f44f4a1cbce4ed252101d8a

src/types.ts
andreabadesso: It is not. 

Fixed in 5a2a383af42aa1bc8c959fcad1988309269be022

tests/db.test.ts
andreabadesso: Wow, you are right. The test was pointless

Fixed in c9d41298ee3c48040b853f3ca81889bca2235de3

src/db.ts
pedroferreira1: As we talked today, this needs a fix and a test that fails in this case.

andreabadesso: Fixed in 142c9d39a6be1ee5c9c0db26665b6b3d9f40cb53

src/db.ts
andreabadesso: We don't, I removed it

db/migrations/20210718180539-add_tx_proposal_version.js
andreabadesso: We don't, I deleted the migration as it was not yet deployed

DATABASE.md
andreabadesso: We don't. 

I removed it in 5a2a383af42aa1bc8c959fcad1988309269be022

src/api/utxos.ts
andreabadesso: As we discussed on slack, I don't think there would be a huge performance gain from having a separate method that will have a different query (returning the address list as strings) as opposed to mapping the query result for the addresses

But I can open an issue to refactor this later, what do you think?

pedroferreira1: I don't think there is need for an issue. I think the code is ok as it is right now.

pedroferreira1 (approved):

DATABASE.md
```diff
@@ -148,6 +138,9 @@ CREATE TABLE `transaction` (
 );
 
 CREATE INDEX transaction_version_idx USING HASH ON `transaction`(`version`);
+CREATE INDEX tx_output_heightlock_idx USING HASH ON `tx_output`(`heightlock`);
+CREATE INDEX tx_output_timelock_idx USING HASH ON `tx_output`(`timelock`);
+CREATE INDEX transaction_height_idx USING HASH ON `transaction`(`height`);
```
luislhl, line 143: Do we have sequelize migrations for those indexes?

db/migrations/20210718180425-remove_tx_proposal_outputs.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.dropTable('tx_proposal_outputs');
```
luislhl, line 5: We will have to be careful with this kind of change when the service is in production, to not break compatibility with the previous version of the application in the meantime between running the migrations and deploying the app.

But for now I think it's ok

src/api/utxos.ts
```diff
@@ -0,0 +1,163 @@
...
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
+
+  const walletAddresses = await getWalletAddresses(mysql, walletId);
+
+  if (value.txId !== undefined) {
+    if (!Number.isInteger(value.index)) {
```
luislhl: I think you could also declare this requirement in the Joi schema: https://github.com/sideway/joi/blob/master/API.md#objectandpeers-options

Unless you think it's better to have it here as an if.

src/api/utxos.ts
```diff
@@ -0,0 +1,163 @@
...
+});
+
+/**
+ * Confirm that the requested addresses belongs to the user's wallet
+ *
+ * @param walletAddresses - The user wallet id
+ * @param addresses - List of addresses to validate
+ * @returns A list with the denied addresses, if any
+ */
+export const validateAddresses = async (walletAddresses: AddressInfo[], addresses: string[]): Promise<string[]> => {
```
luislhl: Why is this function async?

src/api/utxos.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (!utxo) {
+      return closeDbAndGetError(mysql, ApiError.UTXO_NOT_FOUND);
+    }
+
+    // check if the utxo is a member of the user's wallet
+    const denied = await validateAddresses(walletAddresses, [utxo.address]);
+
+    if (denied.length > 0) {
+      // the requested utxo does not belong to the user's wallet.
+      return closeDbAndGetError(mysql, ApiError.FORBIDDEN);
```
luislhl: Shouldn't we have a more descriptive error here like the other ones?

src/api/utxos.ts
```diff
@@ -0,0 +1,163 @@
...
+      message: err.message,
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
+
+  const walletAddresses = await getWalletAddresses(mysql, walletId);
+
+  if (value.txId !== undefined) {
```
luislhl, line 76: Maybe a comment here explaining in which case the `txId` exists would improve readability. I had to go to the design to check the explanation there.

luislhl: I mean, the confusion is not about when it exists, it's obvious that it exists when it's in the query. The comment would be to explain in which case someone sends it in the query.

DATABASE.md
andreabadesso: image

We do, I'll create another PR to remove the `DATABASE.md` file as it is now useless and confusing

src/api/txProposalCreate.ts
```diff
@@ -115,27 +66,39 @@ export const create = walletIdProxyHandler(async (walletId, event) => {
   }
 
   const body = value;
+  const tx = hathorLib.helpersUtils.createTxFromHex(body.txHex, new hathorLib.Network(process.env.NETWORK));
```
luislhl: Should we create a type for the transaction returned here?

For now TypeScript is typing this as `any`, since the wallet-lib doesn't have typing information.

db/migrations/20210718180425-remove_tx_proposal_outputs.js
andreabadesso: You are right, I'll create an issue to discuss the best approach for those kind of migrations

src/api/utxos.ts
andreabadesso: I agree! I didn't find this requirement on the docs

Fixed in 9859d9477dae545229b9dbca7028291eb2e665bf

src/api/utxos.ts
andreabadesso: You are right, fixed in 7ba421a85a3b3a29a9fb900e49bbb868cd5978cf

src/api/utxos.ts
andreabadesso: I agree! Fixed in aca353a8f0d61e39b2d37837404e2aa7b1646275

src/api/txProposalCreate.ts
andreabadesso: I've just tested it and TypeScript was getting the type from the wallet-lib method, but I added the type on the wallet-service as well to make it more explicit

Done in 0246942695655b474ac7a6f4c457bf38c3409de0

src/api/utxos.ts
andreabadesso: I agree

Done in 67ae4d6921d5ff438d25de00f78703e5f6d2ec22

db/migrations/20210718180425-remove_tx_proposal_outputs.js
andreabadesso: Issue created: https://github.com/HathorNetwork/hathor-wallet-service/issues/120

src/api/txProposalCreate.ts
luislhl: Ok, maybe the problem is my VScode, then.

src/api/txProposalCreate.ts
```diff
@@ -145,78 +108,36 @@ export const create = walletIdProxyHandler(async (walletId, event) => {
...
-  const changeOutputs = getChangeOutputs(diff, addresses);
-
-  const finalOutputs = outputs.concat(changeOutputs);
 
-  // we also need to do this check here, as we may have added change outputs
-  if (finalOutputs.length > hathorLib.transaction.getMaxOutputsConstant()) {
-    return closeDbAndGetError(mysql, ApiError.TOO_MANY_OUTPUTS, { outputs: finalOutputs.length });
+  // check if inputs sent by user are not part of another tx proposal
+  if (checkUsedUtxos(inputUtxos)) {
+    return closeDbAndGetError(mysql, ApiError.INPUTS_ALREADY_USED);
```
luislhl, line 129: What happens in case 2 txProposals using the same inputs are created concurrently, and both pass this check? This is a possible scenario, right?

In which future step we detect the duplicity?

I'm asking this more out of curiosity than anything

serverless.yml
```diff
@@ -92,7 +92,6 @@ functions:
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
   loadWalletApi:
-    role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
```
luislhl: This isn't needed anymore?
Can we also remove the role from AWS IAM?

src/db.ts
```diff
@@ -220,11 +222,16 @@ export const updateWalletStatus = async (
   mysql: ServerlessMysql,
   walletId: string,
   status: WalletStatus,
+  retryCount = 0,
 ): Promise<void> => {
   const ts = getUnixTimestamp();
   await mysql.query(
```
luislhl, line 228: Should we use timeout on mysql queries? This could avoid the Lambda timeout being hit in some situations, which is a scenario we know gives us no information about what exactly occurred.

luislhl (approved):

src/api/txProposalCreate.ts
andreabadesso: I think you are right, we won't detect the duplicity, `markUtxosWithProposalId` will overwrite the last `txProposalIndex` on the `tx_output` table and the txProposal that loses the race will be in an invalid state

I will create an issue to properly handle this scenario

andreabadesso: Issue created: https://github.com/HathorNetwork/hathor-wallet-service/issues/122

serverless.yml
andreabadesso: This should not have been removed, it is still needed

Fixed in 6a5cd149a3b70def6f81d3bf6b5bcc20eafa4b94

src/db.ts
andreabadesso: Yes we should, we've talked about this before

I created the issue at https://github.com/HathorNetwork/hathor-wallet-service/issues/123

PR #87, opened 2021-07-02
feat: websocket wallet auth

### LINKED ISSUE #84: [Design] Websocket: Wallet authorization for real-time
# Real-Time Authorization
When subscribing to the events of a wallet, the client should verify that he owns the wallet.
This can be done by passing a token created with the `/auth/token` API when subscribing to the wallet.

## Discussion

Should we use the authorization method when connecting (`$connect` route) instead of when subscribing to the wallet?
- By enforcing that only wallet owners can connect to the API Gateway Websocket, we can mitigate DOS attacks. Too many connections on the API Gateway will consume our resources, also each connection will create at least one key in our Redis instance, creating too many may overload the instance and DOS the actual users.
-  If we expand the use of websockets to use cases where a wallet subscription is not needed, another authentication method will have to be developed (this is not a drawback, only an observation).
    - In this case, if no alternative authentication method is developed, we would be open to a DOS attack described above.
- This way we can enforce a limit of open connections for a given wallet.
- This may impede having one connection for multiple wallets (if such feature is desired).

r4mmer:
The design is available in https://github.com/HathorNetwork/hathor-wallet-service/issues/84


r4mmer:
Serverless offline does not support [yet](https://github.com/dherault/serverless-offline/pull/1222) the websocket authorizer, as described [here](https://github.com/dherault/serverless-offline/issues/951).

So we should test if this works as intended on staging environment.
If this doesn't work, we could move the jwt verification process to `common.js` and import from there.

luislhl:
@r4mmer , could you add a description of what we are trying to achieve in this PR and the main alterations?

package.json
```diff
@@ -50,7 +50,7 @@
     "fork-ts-checker-webpack-plugin": "^3.0.1",
     "jest": "^26.0.1",
     "npm-run-all": "^4.1.5",
-    "serverless": "^1.83.3",
+    "serverless": "^2.51.2",
     "sequelize-cli": "^6.2.0",
     "serverless-dotenv-plugin": "^2.4.2",
```
luislhl: We can uninstall this plugin. It's not needed anymore.

serverless.yml
```diff
@@ -11,15 +18,21 @@ custom:
...
       name: bearerAuthorizer
-      type: TOKEN
-      identitySource: method.request.header.Authorization
-      identityValidationExpression: Bearer (.*)
+      type: token
+      identitySource:
+        - method.request.header.Authorization
+      resultTtlInSeconds: 600
+    wsBearer:
+      name: bearerAuthorizer
```
luislhl: What this name is used for?

I noticed it has the same name as the other authorizer.

src/api/auth.ts
```diff
@@ -177,30 +177,29 @@ const _generatePolicy = (principalId: string, effect: string, resource: string)
...
-    // XXX: find a way to return specific error to frontend or make all errors Unauthorized?
-    //
     // Identify exception from jsonwebtoken by the name property
     // https://github.com/auth0/node-jsonwebtoken/blob/master/lib/TokenExpiredError.js#L5
     if (e.name === 'JsonWebTokenError') {
-      throw new Error('Unauthorized');
+      throw new Error('Invalid Token');
     } else if (e.name === 'TokenExpiredError') {
-      throw new Error('Unauthorized');
+      throw new Error('Expired Token');
```
luislhl: What status code will be returned in the request in those cases, 401?

What defines the status code returned?

src/api/auth.ts
```diff
@@ -7,7 +7,7 @@
 
 import {
   APIGatewayProxyHandler,
-  APIGatewayTokenAuthorizerHandler,
+  CustomAuthorizerHandler,
```
luislhl: Apparently this is deprecated. Shouldn't we use the one recommended by them?

```
/** @deprecated Use APIGatewayAuthorizerHandler or a subtype */
export type CustomAuthorizerHandler = Handler<CustomAuthorizerEvent, APIGatewayAuthorizerResult>;
```

src/api/auth.ts
```diff
@@ -222,5 +221,5 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = async (event)
     return _generatePolicy(walletId, 'Allow', event.methodArn);
   }
 
-  return _generatePolicy(walletId, 'Deny', event.methodArn);
+  throw new Error('Unexpected Error.');
```
luislhl: Is it possible that if the token is expired, we will throw this error?

I'm not sure because we apparently have two places that check the token expiration, the `jwt.verify` in line 192 and the if in line 220.

If the expired token can reach this if in line 220, we will be throwing this `Unexpected Error` for it, when we could be throwing a better error message.

If it can't reach this if in line 220, why do we need to check its expiration in the if?

src/ws/connection.ts
```diff
@@ -8,38 +8,74 @@
...
-    await initWsConnection(redisClient, connInfo);
+  let walletId: string;
+  try {
+    if (!process.env.IS_OFFLINE) {
+      walletId = event.requestContext.authorizer.principalId;
+    } else {
+      // serverless offline does not support websocket authorizers
+      // https://github.com/dherault/serverless-offline/issues/951
+      // For offline testing purposes get walletId from env
+      walletId = process.env.DEV_WALLET_ID;
```
luislhl, line 37: I feel we should have a configuration file (`settings.js`) where all environment variables are read, and any other part of the code imports them from the configuration file, instead of reading them directly.

This will make it easier to know all the env vars we need for the application to work.

src/ws/connection.ts
```diff
@@ -8,38 +8,74 @@
...
-  }
+  // Init connection then join wallet
+  await initWsConnection(
+    redisClient,
+    connInfo,
+  ).then(() => wsJoinWallet(
+    redisClient,
+    connInfo,
+    walletId,
+  ));
```
luislhl: Isn't it possible that the `closeRedisClient` function below will be executed before `wsJoinWallet`? This shouldn't happen, right?

Why don't we also `await wsJoinWallet`?

```
 await initWsConnection(
    redisClient,
    connInfo,
  );
  
  await wsJoinWallet(
    redisClient,
    connInfo,
    walletId,
  );
```

serverless.yml
r4mmer: The name is the function of the authorizer.

src/api/auth.ts
r4mmer: The API Gateway returns 401 for any `Error` thrown in the authorizer.

src/ws/connection.ts
r4mmer: This would be nice, but I think we should implement this on another PR.

src/ws/connection.ts
r4mmer: The `await` also waits for the `then` to resolve.

src/api/auth.ts
r4mmer: Changed to verify expiration only on the `jwt.verify` method and throw more descriptive error if our `verifySignature` fails.

src/api/auth.ts
r4mmer: Changed back to `APIGatewayTokenAuthorizerHandler` since the authorizers will be of type `token` they should have the `authorizationToken` which the `APIGatewayAuthorizerHandler` lacks.

package.json
r4mmer: Should i also remove the `useDotenv: true` from `serverless.yml`?

luislhl: No, this is the new built-in option from Serverless that replaces the plugin. 

There is a warning about this in the beginning of this page: https://www.serverless.com/plugins/serverless-dotenv-plugin

luislhl (dismissed):

src/ws/connection.ts
andreabadesso: I agree with @luislhl that this would be nice. Maybe even throwing an exception if required environment variables are not loaded

@r4mmer can you open an issue describing that?

src/ws/connection.ts
```diff
@@ -8,38 +8,82 @@
...
+  try {
+    // attempt to join wallet
+    await wsJoinWallet(redisClient, connInfo, walletId);
+  } catch (ex) {
+    // wallet limit exceeded, disconnect user
+    if (ex instanceof WalletLimitExceeded) {
+      // disconnect client because he exceeds the limit
+      await disconnectClient(redisClient, connInfo);
+    } else {
+      throw ex;
```
andreabadesso: Shouldn't we also disconnect the client here?

src/ws/connection.ts
```diff
@@ -8,38 +8,82 @@
...
+export const ping = async (
+  event: APIGatewayProxyEvent,
+): Promise<void> => {
+  const redisClient = getRedisClient();
+  const connInfo = connectionInfoFromEvent(event);
+  await sendMessageToClient(redisClient, connInfo, { message: 'PONG' });
 
   await closeRedisClient(redisClient);
-  await closeDbConnection(mysql);
 };
```
andreabadesso, line 81: **Unrelated to line number**

Can you add some basic tests on `ws/redis.ts`? I know that serverless-offline does not allow offline websocket auth but there are methods that we can mock and test here

Also, coverage is quite low on those files and this may break the CI after the latest merge from Luis PR-validation branch

andreabadesso (changes requested):

src/ws/connection.ts
r4mmer: The websocket cannot be tested on the unit tests, if we could disable the CI coverage for the `ws` folder it would be better, but the `redis.ts` should be well tested.

r4mmer: The websocket cannot be tested on the unit tests, if we could disable the CI coverage for the ws folder it would be better, but the redis.ts should be well tested.

src/ws/connection.ts
r4mmer: Done.

andreabadesso (approved):

PR #88, opened 2021-07-05
chore: deployment preparation

luislhl:
This PR contains some changes we are identifying while deploying and testing the service.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -204,6 +204,7 @@ functions:
     timeout: 1
   authTokenApi:
     handler: src/api/auth.tokenHandler
+    timeout: 1
     events:
       - http:
           path: auth/token
--- a/serverless.yml
+++ b/serverless.yml
@@ -48,22 +48,22 @@ provider:
   apiGateway:
     minimumCompressionSize: 1024 # Enable gzip compression for responses > 1 KB
   environment:
+    ADMINTOKEN: ${env:ADMINTOKEN}
     AWS_VPC_DEFAULT_SG_ID: ${env:AWS_VPC_DEFAULT_SG_ID}
     AWS_SUBNET_ID_1: ${env:AWS_SUBNET_ID_1}
     AWS_SUBNET_ID_2: ${env:AWS_SUBNET_ID_2}
     AWS_SUBNET_ID_3: ${env:AWS_SUBNET_ID_3}
     AWS_NODEJS_CONNECTION_REUSE_ENABLED: 1
-    SERVICE_NAME: ${self:service}
-    NETWORK: ${env:NETWORK}
-    MAX_ADDRESS_GAP: ${env:MAX_ADDRESS_GAP}
     BLOCK_REWARD_LOCK: ${env:BLOCK_REWARD_LOCK}
-    ADMINTOKEN: ${env:ADMINTOKEN}
     CONFIRM_FIRST_ADDRESS: ${env:CONFIRM_FIRST_ADDRESS}
     DB_ENDPOINT: ${env:DB_ENDPOINT}
     DB_PORT: ${env:DB_PORT}
     DB_NAME: ${env:DB_NAME}
     DB_USER: ${env:DB_USER}
     DB_PASS: ${env:DB_PASS}
+    SERVICE_NAME: ${self:service}
+    MAX_ADDRESS_GAP: ${env:MAX_ADDRESS_GAP}
+    NETWORK: ${env:NETWORK}
     NEW_TX_SQS: { Ref: WalletServiceNewTxQueue }
     REDIS_HOST: ${env:REDIS_HOST}
     REDIS_PORT: ${env:REDIS_PORT}
@@ -81,6 +81,7 @@ functions:
     timeout: 600
   onNewTxRequest:
     handler: src/txProcessor.onNewTxRequest
+    memorySize: 256
   onHandleReorgRequest:
     handler: src/txProcessor.onHandleReorgRequest
     timeout: 300 # 5 minutes
--- a/serverless.yml
+++ b/serverless.yml
@@ -32,7 +32,11 @@ resources:
 provider:
   name: aws
   runtime: nodejs12.x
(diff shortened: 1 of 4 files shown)


serverless.yml
```diff
@@ -31,24 +31,43 @@ resources:
...
+    DB_PORT: ${env:DB_PORT}
+    DB_NAME: ${env:DB_NAME}
+    DB_USER: ${env:DB_USER}
+    DB_PASS: ${env:DB_PASS}
     SERVICE_NAME: ${self:service}
-    DB_ENDPOINT: ${env:DB_ENDPOINT, 'rds-endpoint-TODO'}
-    DB_NAME: ${self:service}-${env:STAGE}
-    DB_USER: ${env:STAGE}-user
+    MAX_ADDRESS_GAP: ${env:MAX_ADDRESS_GAP}
+    NETWORK: ${env:NETWORK}
```
andreabadesso, line 68: We also need to add the `AUTH_SECRET` environment variable

This can be a random string of any length

I am generating it locally using:
`openssl rand -base64 32`

serverless.yml
```diff
@@ -16,8 +16,8 @@ custom:
       identityValidationExpression: Bearer (.*)
 
 # Create an optimized package for our functions
-package:
-  individually: true
+# package:
```
andreabadesso: Remove instead of commenting it out?

serverless.yml
```diff
@@ -31,24 +31,43 @@ resources:
       Type: "AWS::SQS::Queue"
       Properties:
         QueueName:
-            WalletServiceNewTxQueue_${env:STAGE}
+            WalletServiceNewTxQueue_${opt:stage, 'dev'}
```
andreabadesso, line 30: What is `opt:stage`?

serverless.yml
```diff
@@ -59,8 +78,10 @@ provider:
 functions:
   getLatestBlock:
     handler: src/height.getLatestBlock
+    timeout: 600
   onNewTxRequest:
     handler: src/txProcessor.onNewTxRequest
+    memorySize: 256
```
andreabadesso: I think we can set this for all methods, the 1GB default seems to be too much

andreabadesso (changes requested):

serverless.yml
luislhl: It's the value of the `--stage` option passed to serverless cli.

If we run `serverless deploy --stage prod`, for instance, its value is `prod`.

serverless.yml
```diff
@@ -16,8 +16,8 @@ custom:
       identityValidationExpression: Bearer (.*)
 
 # Create an optimized package for our functions
-package:
-  individually: true
```
pedroferreira1, line 20: What's this configuration for?

serverless.yml
```diff
@@ -59,8 +78,10 @@ provider:
 functions:
   getLatestBlock:
     handler: src/height.getLatestBlock
+    timeout: 600
```
pedroferreira1: Decrease?

pedroferreira1 (approved):

serverless.yml
luislhl: It makes Serverless build one individual package (.zip) for each function.

In conjuction with the webpack plugin, it does more: https://github.com/serverless-heaven/serverless-webpack#optimization--individual-packaging-per-function

However, this was causing our build to be extremely painful. It took a long time and used all memory in the dev's machine.
Apparently it was running Webpack build one time for each function.

After removing this, this build complete in something like 10s.

And the generated package is not so big. So it seems we have no reason to use this.

serverless.yml
luislhl: Done

serverless.yml
luislhl: Done

serverless.yml
luislhl: Done. I can pass you those secrets later.

serverless.yml
```diff
@@ -31,24 +27,45 @@ resources:
...
         QueueName:
-            WalletServiceNewTxQueue_${env:STAGE}
+            WalletServiceNewTxQueue_${opt:stage, 'dev'}
 
 provider:
   name: aws
   runtime: nodejs12.x
-  stage: ${env:STAGE}
-  region: us-east-1
+  memorySize: 256
```
msbrogli, line 37: What's this parameter? RAM? MB? GB? Minimum value?

serverless.yml
```diff
@@ -167,12 +188,16 @@ functions:
...
     handler: src/ws/admin.broadcast
+    timeout: 1
   wsAdminDisconnect:
     handler: src/ws/admin.disconnect
+    timeout: 1
   wsAdminMulticast:
     handler: src/ws/admin.multicast
+    timeout: 1
   authTokenApi:
     handler: src/api/auth.tokenHandler
```
msbrogli, line 206: Shouldn't we add a timeout here as well? Can we set a default value that is used in case one is not provided?

msbrogli (approved):

serverless.yml
luislhl: RAM, MB.

It's the memory allocated for the Lambda. So, Maximum memory it can use.

I will add a comment explaining it.

serverless.yml
luislhl: We can set the default value. The default default value is 6s, so every Lambda that is not defining it should be using 6s as timeout.

I think @andreabadesso and @r4mmer are better positioned to specify which timeout each Lambda should have, or if 6s is enough.

luislhl: I've also set the timeout explicitly for Lambdas that have not set it just for clarity, but I've put 6s, which was already the default.

r4mmer: The `authTokenApi` should be our fastest lambda since it does not access the db or redis. Having a short timeout in this one wouldn't hurt.

andreabadesso (approved):

serverless.yml
luislhl: Done

PR #92, opened 2021-07-06
fix: treat exceptions from jsonwebtoken

r4mmer:


src/api/auth.ts
```diff
@@ -180,10 +180,25 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = async (event)
...
+    data = jwt.verify(
+      sanitizedToken,
+      process.env.AUTH_SECRET,
+    );
+  } catch (e) {
+    // XXX: find a way to return specific error to frontend or make all errors Unauthorized?
+    //
+    // Identify exception from jsonwebtoken by the name property
+    // https://github.com/auth0/node-jsonwebtoken/blob/master/lib/TokenExpiredError.js#L5
+    if (e.name === 'JsonWebTokenError') {
```
andreabadesso, line 194: What is this error? An invalid JWT was sent from the user?

src/api/auth.ts
```diff
@@ -180,10 +180,25 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = async (event)
...
-    sanitizedToken,
-    process.env.AUTH_SECRET,
-  );
+  let data;
+  try {
+    data = jwt.verify(
+      sanitizedToken,
+      process.env.AUTH_SECRET,
+    );
+  } catch (e) {
```
andreabadesso, line 189: I think we should log every unhandled error here

src/api/auth.ts
```diff
@@ -180,10 +180,25 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = async (event)
...
+    // XXX: find a way to return specific error to frontend or make all errors Unauthorized?
```
andreabadesso, line 190: I think it would be nice to inform the user the type of error, if possible

andreabadesso (changes requested):

src/api/auth.ts
r4mmer: This is the most generic error, invalid jwt (like `"Bearer 123"`) or secret mismatch, any error that prevents the lib from verifying that this token is valid except for errors related to the headers `exp` and `nbf`.

src/api/auth.ts
r4mmer: Done.

src/api/auth.ts
r4mmer: I need to test this, but if the error thrown is an instance of `Error`, the API Gateway will return the message inside to the user along with a 401 (i think). I don't like exposing internal errors to the user but we could do it.

pedroferreira1 (approved):

src/api/auth.ts
andreabadesso: I agree with not exposing internal errors, but we are in control of the error that will get thrown here, right?

We could just throw an error with the nessage we want inside the catch block

andreabadesso (approved):

PR #93, opened 2021-07-07
chore: added an arn to allow the loadWallet lambda to invoke loadWall…

andreabadesso:
This is necessary so the `loadWallet` can invoke `loadWalletAsync`.

There is no support for creating the role on `serverless.yml` yet, so @luislhl created it manually through AWS and we are configuring the ARN on serverless.yml

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -71,6 +71,7 @@ functions:
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
   loadWalletApi:
+    role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
     handler: src/api/wallet.load
     events:
       - http:


luislhl (approved):
> There is no support for creating the role on serverless.yml yet, so @luislhl created it manually through AWS and we are configuring the ARN on serverless.yml

Actually there is support. In Serverless you can create anything using CloudFormation syntax.

But it seemed too complicated at first, so I created manually just to get faster feedback if that would work.

If you prefer, we can try on serverless too.

andreabadesso:
> > There is no support for creating the role on serverless.yml yet, so @luislhl created it manually through AWS and we are configuring the ARN on serverless.yml
> 
> Actually there is support. In Serverless you can create anything using CloudFormation syntax.
> 
> But it seemed too complicated at first, so I created manually just to get faster feedback if that would work.
> 
> If you prefer, we can try on serverless too.

I think we can open an issue to configure it on `serverless.yml` in the future

serverless.yml
```diff
@@ -71,6 +71,7 @@ functions:
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
   loadWalletApi:
+    role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
```
pedroferreira1, line 74: Is it gonna work for mainnet, testnet and dev deploys?

pedroferreira1 (approved):

serverless.yml
andreabadesso: The role is currently configured to work specifically for the `loadWalletAsync` lambda on every stage:

`arn:aws:lambda:*:769498303037:function:hathor-wallet-service-*-loadWalletAsync`

PR #94, opened 2021-07-07
feat(db): migrations

r4mmer:


db/models/address.js
```diff
@@ -0,0 +1,43 @@
...
+      type: DataTypes.STRING(34),
+      allowNull: false,
+      primaryKey: true,
+    },
+    index: {
+      type: DataTypes.INTEGER.UNSIGNED,
+      allowNull: true,
+      defaultValue: null,
+    },
+    wallet_id: {
```
luislhl, line 27: Why aren't we using Foreign Keys for references like this one?

db/models/addressbalance.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+const {
+  Model
+} = require('sequelize');
```
luislhl, line 4: Did you have a specific reason for using Javascript instead of Typescript?

db/config.js
```diff
@@ -0,0 +1,54 @@
...
+    development = {
+      storage: process.env.DB_FILE || './db.sqlite3',
+      dialect: 'sqlite',
+    };
+    break;
+}
+
+module.exports = {
+  development,
+  test: {
```
luislhl, line 29: What exactly selects between those options? An EnvVar?

db/models/address.js
r4mmer: I translated the same database schema we are currently using, it does not have foreign keys. We can add them, but for this first PR I think we should aim to have the same db described in `DATABASE.md`

db/models/addressbalance.js
r4mmer: Sequelize does not support typescript for migration scripts, and since we are not using the models (we won't change the queries or use the ORM part of sequelize in this PR) I made them for description and reference only (later we could easily translate to typescript using these as base).

db/config.js
r4mmer: There are 2 ways to select the environment, using the EnvVar `NODE_ENV` or by the `--env` parameter (default is `development`). Sequelize docs on the matter [here](https://sequelize.org/master/manual/migrations.html#configuration)

db/models/address.js
luislhl: Makes sense. But do we already have another issue to discuss the use of Foreign Keys?

luislhl (approved):

db/config.js
```diff
@@ -0,0 +1,54 @@
...
+  production: {
+    username: process.env.PROD_DB_USERNAME,
+    password: process.env.PROD_DB_PASSWORD,
+    database: process.env.PROD_DB_NAME,
+    host: process.env.PROD_DB_HOSTNAME,
+    port: process.env.PROD_DB_PORT,
+    dialect: 'mysql',
+    dialectOptions: {
+      bigNumberStrings: true,
+      // ssl: {
```
andreabadesso: Remove this?

db/migrations/20210707174416-create-tx-proposal.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable('tx_proposal', {
+      id: {
+        type: Sequelize.STRING(36),
+        allowNull: false,
+        primaryKey: true,
+      },
+      wallet_id: {
```
andreabadesso, line 10: You could use camelCase here and define the field name as `wallet_id`

What do you think?

package.json
```diff
@@ -48,11 +50,13 @@
     "fork-ts-checker-webpack-plugin": "^3.0.1",
     "jest": "^26.0.1",
     "npm-run-all": "^4.1.5",
+    "sequelize-cli": "^6.2.0",
     "serverless": "^1.73.1",
     "serverless-dotenv-plugin": "^2.4.2",
     "serverless-offline": "^6.4.0",
     "serverless-prune-plugin": "^1.5.0",
     "serverless-webpack": "^5.5.0",
+    "sqlite3": "^5.0.2",
```
andreabadesso: What are we using it for?

I think we should remove it until we refactor the queries to be able to run on sqlite3

db/seeders/mainnet.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.bulkInsert('metadata', [{
```
andreabadesso: We are not using `metadata` anymore, we are currently fetching the latest block from the `transactions` table

We should also remove it from `DATABASE.md`

package.json
```diff
@@ -48,11 +50,13 @@
...
     "ts-jest": "^26.1.1",
     "ts-loader": "^5.3.3",
     "typescript": "^3.9.5",
```
andreabadesso: *Unrelated to line number:*

I think we should add the comments from `DATABASE.md` on the models, example:

on `tx_output` table, we should add a comment explaining that `tx_id` might be a block

andreabadesso (changes requested):

db/models/address.js
r4mmer: No we don't

db/migrations/20210707174416-create-tx-proposal.js
r4mmer: For migrations I don't think it would make sense, and we could add this to the models, but since we will be using the models just for reference (for now) I think we should keep them as closely related to the migration schema as we can.

db/config.js
r4mmer: Done.

package.json
r4mmer: Done.

package.json
r4mmer: Done.

db/seeders/mainnet.js
r4mmer: Done.

db/migrations/20210706163010-create-address.js
```diff
@@ -0,0 +1,29 @@
+'use strict'; // eslint-disable-line
```
pedroferreira1, line 1: No need to do it in this PR but in this table definition is possible to add db indexes already?

db/config.js
```diff
@@ -0,0 +1,51 @@
...
+      storage: process.env.DB_FILE || './db.sqlite3',
+      dialect: 'sqlite',
+    };
+    break;
+}
+
+module.exports = {
+  development,
+  test: {
+    username: process.env.CI_DB_USERNAME,
```
pedroferreira1, line 30: This is for testing or for CI?

db/migrations/20210706164553-create-address-balance.js
```diff
@@ -0,0 +1,46 @@
+'use strict'; // eslint-disable-line
+module.exports = {
+  up: async (queryInterface, Sequelize) => { // eslint-disable-line
+    await queryInterface.createTable('address_balance', {
+      address: {
+        type: Sequelize.STRING(34),
+        allowNull: false,
+        primaryKey: true,
```
pedroferreira1, line 8: When you set 2 `primaryKey: true` it means they are unique together? Because `address` can be repeated, just like `token_uid`

db/migrations/20210707190612-create-transaction.js
```diff
@@ -0,0 +1,41 @@
...
+        type: Sequelize.BOOLEAN,
+        allowNull: false,
+        defaultValue: false,
+      },
+      height: {
+        type: Sequelize.INTEGER.UNSIGNED,
+        allowNull: true,
+        defaultValue: null,
+      },
+    }).then(() => queryInterface.addIndex(
```
pedroferreira1, line 28: Nice, is it our only index for now?

db/models/address.js
pedroferreira1: I think it's important to create. And I agree to keep this out of this PR context

db/migrations/20210706163010-create-address.js
r4mmer: Yes, we are already adding an index for the `transaction` table.

db/config.js
r4mmer: CI, but we could add another for testing only if we need it.

db/migrations/20210706164553-create-address-balance.js
r4mmer: Yes, it translates to SQL as `PRIMARY KEY (address, token_id)`, both columns together are the primary key.

db/migrations/20210707190612-create-transaction.js
r4mmer: Yes.

andreabadesso (approved):

PR #95, opened 2021-07-07
fix(auth): policy generated

r4mmer:
ARN of API Gateway resource: `arn:aws:execute-api:<region>:<account>:<project>/<stage>/<http-method>/<path>`
(Any path parameter is a `*`)

When a user authenticates with his wallet he should receive permission to access the `/wallet/*` and `/tx/*` instead of the resource he currently is accessing.

### DIFF
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -150,6 +150,7 @@ export const tokenHandler: APIGatewayProxyHandler = async (event) => {
  * Generates a aws policy document to allow/deny access to the resource
  */
 const _generatePolicy = (principalId: string, effect: string, resource: string) => {
+  const resourcePrefix = `${resource.split('/').slice(0, 2).join('/')}/*`;
   const policyDocument: PolicyDocument = {
     Version: '2012-10-17',
     Statement: [],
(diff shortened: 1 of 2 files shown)


r4mmer:
Paths before and after this PR.

- `GET /wallet` => `GET /wallet/status`
- `POST /wallet` => `POST /wallet/init`
- `/addresses` => `/wallet/addresses`
- `/balances` => `/wallet/balances`
- `/txhistory` => `/wallet/history`
- `/txproposals` => `/tx/proposal`
- `/txproposals/<id>` => `/tx/proposal/<id>`

src/api/auth.ts
```diff
@@ -158,7 +159,10 @@ const _generatePolicy = (principalId: string, effect: string, resource: string)
   const statementOne: Statement = {
     Action: 'execute-api:Invoke',
     Effect: effect,
-    Resource: resource,
+    Resource: [
+      `${resourcePrefix}/wallet/*`,
+      `${resourcePrefix}/tx/*`,
```
andreabadesso, line 164: From what I understood here, the generated resources will be something like:

`arn:aws:execute-api:us-east-1:769498303037:x7896x5fv6/*/wallet/*`

What is the first `*` for?

andreabadesso (approved):

src/api/auth.ts
r4mmer: There's also the stage, so it would be:

`arn:aws:execute-api:us-east-1:769498303037:x7896x5fv6/staging/*/wallet/*` (The stage is also part of the path).

The first `*` is the method (`GET`, `POST`, etc.) and the second is the rest of the path (or path variable, if exists)

pedroferreira1 (approved):

PR #97, opened 2021-07-09
ci: add pr-validation pipeline

luislhl:


docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: "3.8"
+services:
```
pedroferreira1, line 2: Should we add the redis here as well?

pedroferreira1 (approved):

docker-compose.yml
luislhl: Done

andreabadesso (approved):

PR #104, opened 2021-07-14
chore: added indexes for tx_output and transaction tables

andreabadesso:


luislhl (approved):

pedroferreira1 (changes requested):
I like the indexes, I just don't like that the models are not reflecting the same as in the database.

Shouldn't we add in the models as well? Just like here: https://github.com/HathorNetwork/hathor-wallet-service/blob/dev/db/models/transaction.js#L46

andreabadesso:
@pedroferreira1 You are absolutely correct. Fixed in f79b38a5123393a20876f0bc392e4fe0530d1982

pedroferreira1 (approved):

PR #107, opened 2021-07-19
feat: wallet tokens api

andreabadesso:
This was a request from @pedroferreira1 so he can be able to list all tokens a wallet has using the wallet-lib without needing to fetch the full wallet tx history.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -1188,6 +1188,30 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
 };
 
 /**
+ * Gets a list of tokens that a given walllet has ever interacted with
+ *
+ * @returns A list of tokens.
+ */
+export const getWalletTokens = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+): Promise<string[]> => {
+  const tokenList: string[] = [];
+  const results: DbSelectResult = await mysql.query(
+    `SELECT DISTINCT(token_id)
+       FROM \`wallet_tx_history\`
+      WHERE \`wallet_id\` = ?`,
+    [walletId],
+  );
+
+  for (const result of results) {
+    tokenList.push(<string> result.token_id);
+  }
+
+  return tokenList;
+};
+
+/**
  * Get a wallet's transaction history for a token.
  *
  * @remarks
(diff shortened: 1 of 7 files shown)


src/db.ts
```diff
@@ -1187,6 +1187,30 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
   return balances;
 };
 
+/**
+ * Gets a list of tokens that a given walllet has ever interacted with
```
pedroferreira1: Triple 'l' in walllet

src/types.ts
```diff
@@ -348,6 +348,7 @@ export class WalletTokenBalance {
 export interface TxTokenBalance {
   txId: string;
   timestamp: number;
+  voided?: boolean | null;
```
pedroferreira1: Why this can be null? I always need this info

src/db.ts
```diff
@@ -1187,6 +1187,30 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
...
+ *
+ * @returns A list of tokens.
+ */
+export const getWalletTokens = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+): Promise<string[]> => {
+  const tokenList: string[] = [];
+  const results: DbSelectResult = await mysql.query(
+    `SELECT DISTINCT(token_id)
```
msbrogli, line 1201: Is `wallet_id` and `token_id` indexed? Maybe there should be an index `(token_id, wallet_id)` but we need to check whether this query would be able to use it. Otherwise, it might run `O(n)` and get slower and slower over time.

src/types.ts
msbrogli: From the API, it seems we always load this piece of information. Am I wrong about it?

msbrogli (changes requested):

msbrogli (approved):

src/db.ts
andreabadesso: They are not but I don't think `DISTINCT(token_id)` would use the index

I think I can investigate it further on https://github.com/HathorNetwork/hathor-wallet-service/issues/101

What do you think?

src/types.ts
andreabadesso: You are both right, fixed in 43e50c044820496089eb809000a4db95ae5f0f97

src/db.ts
andreabadesso: Fixed in fae4f886e6ada5b59907957a8ee4f04a9735e46f

pedroferreira1 (approved):

src/db.ts
msbrogli: It seems good to me. Thanks!

PR #112, opened 2021-07-30
feat: handle failure in wallet load

### LINKED ISSUE #110: refactor: handle error properly on loadWalletApi and loadWalletAsync
## Motivation

The `loadWallet` API works by initializing a wallet and invoking an async lambda function (`loadWalletAsync`)

When a failure happened in the `loadWalletAsync`, the user's wallet would be in the state "creating" forever.

This PR will transition the user wallet to an `error` state and allow him to try again (with a max retry count)

## Acceptance criteria

The user should get feedback about wallet errors and be able to retry his request on failures

andreabadesso:


tests/api.test.ts
```diff
@@ -444,6 +453,16 @@ test('POST /wallet', async () => {
...
+  // Clean database so our pubkey is free to be used again:
+
+  await cleanDatabase(mysql);
+
+  const spy = jest.spyOn(Wallet, 'lambdaInvoke');
+
+  const mockImplementation = jest.fn(() => Promise.resolve());
+
+  spy.mockImplementation(mockImplementation);
+
```
pedroferreira1, line 468: I think we should create a test for the retry feature

pedroferreira1 (changes requested):

jest.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/tests/utils.ts'],
   coverageThreshold: {
     global: {
-      branches: 88,
+      branches: 85,
```
r4mmer: Why reduce the coverage threshold?

src/api/wallet.ts
```diff
@@ -111,38 +156,24 @@ export const load: APIGatewayProxyHandler = async (event) => {
...
+     * we only care if the invokation failed or not
+     */
+    await invokeLoadWalletAsync(xpubkey, maxGap);
   } catch (e) {
-    // TODO handle
+    console.error('Error on lambda wallet invoke', e);
+
+    const newRetryCount = wallet.retryCount ? wallet.retryCount + 1 : 1;
+    // update wallet status to 'error'
+    await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, newRetryCount);
```
r4mmer, line 170: When there is an error loading the wallet we update status to error and return `success=true`?

src/api/wallet.ts
```diff
@@ -167,27 +199,43 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
-    walletId,
-    xpubkey,
-  };
+    return {
+      success: true,
+      walletId,
+      xpubkey,
+    };
+  } catch (e) {
+    console.error('Erroed on loadWalletAsync: ', e);
```
r4mmer, line 232: Is this log message ok? also, should we create an issue to have better loggin, instead of console statements?

r4mmer (changes requested):
Why not use an SQS for real async? Testing would be harder but the actual implementation would make more sense than having 2 lambdas running waiting for the loadAsync to finish

src/db.ts
```diff
@@ -2100,6 +2107,19 @@ export const filterUtxos = async (
     throw new Error('Addresses can\'t be empty.');
   }
 
+  const queryParams = [
```
pedroferreira1: What's this? Is it missing a rebase?

tests/api.test.ts
```diff
@@ -456,4 +476,35 @@ test('POST /wallet', async () => {
...
+  // wallet should be in error state:
+  event = makeGatewayEventWithAuthorizer(returnBody.status.walletId, null);
+  result = await walletGet(event, null, null) as APIGatewayProxyResult;
+  returnBody = JSON.parse(result.body as string);
+  expect(result.statusCode).toBe(200);
+  expect(returnBody.success).toBe(true);
+  expect(returnBody.status.status).toStrictEqual('error');
+
+  // retrying should succeed
```
pedroferreira1, line 501: Maybe add a test that fails the maximum of retries and after that we can't retry anymore?

tests/db.test.ts
```diff
@@ -1617,6 +1619,11 @@ test('filterUtxos', async () => {
     txProposalId: null,
     txProposalIndex: null,
   });
+
+  // authorities != 0 and maxUtxos == 1 should return only one authority utxo
+  utxos = await filterUtxos(mysql, { addresses: [addr1, addr2], biggerThan: 0, smallerThan: 3, authority: 1, tokenId, maxUtxos: 1 });
```
pedroferreira1: Another PR code?

src/api/wallet.ts
andreabadesso: I agree this is weird. I refactored it to still return success but with the updated status on the response body on this commit: 841423027fb9e932a5601d31b7d47963cf92f908

The idea here is that the wallet client expects this to be async, the first request is to begin the load wallet mechanism, he will then poll the server until the state changes from `CREATING` to `READY` or `ERROR` 

So the first request worked, but the state changed to `ERROR` instantly

Does it make sense?

src/api/wallet.ts
andreabadesso: It is logging correctly

I agree to have better logging though, here is the newly created issue: https://github.com/HathorNetwork/hathor-wallet-service/issues/118

tests/db.test.ts
andreabadesso: You are right, it was missing a rebase.

I fixed it, can you check again?

tests/api.test.ts
andreabadesso: I agree, done in 034f34524ddf607e150a18280aeb3c717a423ac6

src/db.ts
andreabadesso: Yes! It was missing a rebase, can you check again?

jest.config.js
andreabadesso: We shouldn't, I just increased the coverage and set it back to 88

andreabadesso:
> Why not use an SQS for real async? Testing would be harder but the actual implementation would make more sense than having 2 lambdas running waiting for the loadAsync to finish

The lambda won't wait for the `loadWalletAsync` function to finish, it invokes it using `InvocationType: 'Event'` which invokes it asynchronously (https://docs.aws.amazon.com/pt_br/lambda/latest/dg/API_Invoke.html)

We could refactor it to signal the client that his wallet is ready using WebSockets, but it might finish the load process when the user has his app on background (with the WebSocket conn closed)

src/api/wallet.ts
r4mmer: Ok, but the status returned is the whole wallet info or just the `wallet.status`?

andreabadesso: The whole wallet information

pedroferreira1 (approved):

r4mmer (approved):

PR #116, opened 2021-08-06
feat: void old unconfirmed transactions

andreabadesso:
# Motivation

If for some reason the daemon misses a `UPDATE_TX` WebSocket message that is voiding a mempool transaction, we need to have a way to void those transactions, otherwise, they would be "stuck" unconfirmed on our database forever, calculating the wrong balance for the involved addresses.

# Implementation

There is a lambda verifying the database for transactions that have no height yet are old enough so we can be certain that it should have been confirmed by a block -- it is currently set to run every 20 minutes

It also validates each old transaction on the full node so we are 100% sure that they are voided.

If the transaction is not voided on the full node and we missed its block confirmation, we should alert and throw an error.

# Acceptance criteria

The lambda should void all transactions whilst handling balance recalculations for all transactions that are "stuck" on the database.


src/mempool.ts
```diff
@@ -0,0 +1,58 @@
...
+ * confirmed by a block and are not voided.
+ */
+export const onHandleOldVoidedTxs = async () => {
+  const VOIDED_TX_OFFSET: number = parseInt(process.env.VOIDED_TX_OFFSET || `${20 * 60}`, 10); // 20 minutes default
+  const now: number = getUnixTimestamp();
+  const date: number = now - VOIDED_TX_OFFSET;
+
+  // fetch voided transactions that are older than 20m
+  const voidedTransactions: Tx[] = await getMempoolTransactionsBeforeDate(mysql, date);
+  console.log(`Found ${voidedTransactions.length} voided transactions older than 20m`);
```
luislhl: Maybe we should use the variable `VOIDED_TX_OFFSET` in this log, instead of always logging `20m`.

msbrogli:
> If for some reason the daemon misses a `UPDATE_TX` WebSocket message that is voiding a mempool transaction, we need to have a way to void those transactions, otherwise, they would be "stuck" unconfirmed on our database forever, calculating the wrong balance for the involved addresses.

How can it happen? I mean, as long as the messages are arriving from the websocket, I feel the protocol guarantees the proper delivery.

> If the transaction is not voided on the full node and we missed its block confirmation, we should alert and throw an error.

It might happen if the number of transactions is high enough. For example, let's say there are many transactions arriving and we have more than two tips. In this case, the block must choose only two and the remaining will wait for the next block.

src/mempool.ts
msbrogli: I feel it is a good message log for now (but it might happen in the future if the number of transactions grow enough). See my comment in the PR for an example.

src/mempool.ts
```diff
@@ -0,0 +1,58 @@
...
+    /* We could fetch the height from the first_block and confirm it on our database, but this probably is telling us that
+     * there is a problem with the sync mechanism, so I think we should throw here and let it crash
+     *
+     * TODO: [ALERT] should be replaced with the correct string to get alerted on CloudWatch/Slack
+     */
+    if (!isVoided) {
+      console.error(`[ALERT] Transaction ${tx.txId} is not yet confirmed on our database but it is not voided on the fullnode.`);
+      await closeDbConnection(mysql);
+
+      throw new Error(`Transaction ${tx.txId} not voided`);
```
msbrogli: If we can recover from it, why would we throw an exception? I mean, it should definitely open an issue to be analyzed later, but the service should keep working.

msbrogli (changes requested):

src/mempool.ts
```diff
@@ -0,0 +1,58 @@
...
+  console.log(`Found ${voidedTransactions.length} voided transactions older than 20m`);
+
+  // confirm that all of them are actually voided on the fullnode
+  for (const tx of voidedTransactions) {
+    const isVoided = await isTxVoided(tx.txId);
+
+    /* We could fetch the height from the first_block and confirm it on our database, but this probably is telling us that
+     * there is a problem with the sync mechanism, so I think we should throw here and let it crash
+     *
+     * TODO: [ALERT] should be replaced with the correct string to get alerted on CloudWatch/Slack
```
luislhl: I think this can be removed, we already agreed on the `[ALERT]` string as the correct one.

serverless.yml
```diff
@@ -78,6 +78,10 @@ provider:
         Fn::GetAtt: [ WalletServiceNewTxQueue, Arn ]
 
 functions:
+  onHandleOldVoidedTxs:
+    handler: src/mempool.onHandleOldVoidedTxs
+    events:
+      - schedule: rate(20 minutes)
```
luislhl: Maybe we can even use the `VOIDED_TX_OFFSET` env var here in this configuration.

src/db.ts
```diff
@@ -2185,3 +2185,40 @@ export const getTxProposalInputs = async (
...
+  );
+  const transactions = [];
+
+  for (const result of results) {
+    const tx: Tx = {
+      txId: result.tx_id as string,
+      timestamp: result.timestamp as number,
+      version: result.version as number,
+      voided: result.voided as boolean,
+      height: result.height as number,
```
pedroferreira1: From the query, height is null, right?

tests/db.test.ts
```diff
@@ -1631,3 +1634,17 @@ test('filterUtxos should throw if addresses are empty', async () => {
...
+
+  const transactions = [
+    [TX_IDS[0], 1, 2, false, null],
+    [TX_IDS[1], 5, 2, false, null],
+    [TX_IDS[2], 10, 2, false, null],
+  ];
+
+  await addToTransactionTable(mysql, transactions);
+
+  expect(await getMempoolTransactionsBeforeDate(mysql, 11)).toHaveLength(3);
```
pedroferreira1: Should we test timestamp as 9? So we will get only the first two and also check that it got the correct txs?

src/mempool.ts
```diff
@@ -0,0 +1,58 @@
...
+  const now: number = getUnixTimestamp();
+  const date: number = now - VOIDED_TX_OFFSET;
+
+  // fetch voided transactions that are older than 20m
+  const voidedTransactions: Tx[] = await getMempoolTransactionsBeforeDate(mysql, date);
+  console.log(`Found ${voidedTransactions.length} voided transactions older than 20m`);
+
+  // confirm that all of them are actually voided on the fullnode
+  for (const tx of voidedTransactions) {
+    const isVoided = await isTxVoided(tx.txId);
```
pedroferreira1: Depending on the number of transaction we need to check on the full node this method might take more than 6 seconds, right? I am thinking about the default timeout of the lambda but not sure if it's relevant.

src/mempool.ts
andreabadesso: I agree, done in 959e44e03f5fbc453bb771f1dd80abf5c40c3bb7

src/mempool.ts
andreabadesso: Done in 5214328ab16d76347b3727c8237873ea7a67319a

serverless.yml
andreabadesso: Done in 959e44e

src/db.ts
andreabadesso: Yes it is, do you think we should return a hard-coded `null` there? The type cast will also return null (as it is coming from the database)

tests/db.test.ts
andreabadesso: I agree, done in 605dc4c144aefc805d38501c1bf84038cfa9d0a2

src/mempool.ts
andreabadesso: I've tested with the testnet synced database and it averaged below 1s, I think we can leave it on the default timeout and observe it before increasing it. What do you think?

andreabadesso:
@msbrogli 

> How can it happen? I mean, as long as the messages are arriving from the websocket, I feel the protocol guarantees the proper delivery.

It can (and probably will) happen when the daemon is restarted because we don't have a voided transaction API on the fullnode, so we can't fetch the list of transactions that were voided recently

I agree that as long as the messages are arriving from the WebSocket the protocol guarantees the proper delivery though

> It might happen if the number of transactions is high enough. For example, let's say there are many transactions arriving and we have more than two tips. In this case, the block must choose only two and the remaining will wait for the next block.

I feel that in this case, it would be better to just increase the env variable (from 20 minutes to something higher that guarantees that the tx should have been confirmed. Maybe in the future this could change from a environmental variable to some value that is generated observing the average transaction confirmation time. What do you think?

src/mempool.ts
andreabadesso: I agree, refactored it in d65122eb9a7d6ff184f28e35ae8ed65164949bac to download the `first_block` transaction from the fullnode, get its height and confirm the transaction using it

src/db.ts
pedroferreira1: If the type cast returns null I think it's ok. I was just thinking if it was really needed in the return because it's always null anyway.

src/mempool.ts
pedroferreira1: Yeah, I agree. I think we just need to make sure things will keep consistency if a timeout happens in the middle of the processing. From what I saw, I don't see any problems. What do you think?

tests/db.test.ts
pedroferreira1: Should we check if the txs are the correct one? you are just checking the length, right?

src/mempool.ts
pedroferreira1: You changed only the serverless yml in this commit

pedroferreira1 (approved):

src/mempool.ts
andreabadesso: @pedroferreira1 you are absolutely correct

Fixed in ce4e42fbbd8627a0464543917d4df3e979411d1c

Also fixed an error on the logic that was not multiplying the env variable setting by 60, calculating it as seconds instead of minutes

tests/db.test.ts
andreabadesso: I agree, done in 9f71e85bc7ee5cd58dda88f541001c3536453719

src/mempool.ts
andreabadesso: I agree, I don't see any problems since we will try again in the next scheduled run

r4mmer (approved):

luislhl (approved):

PR #117, opened 2021-08-09
feat: rollback balance calculation on txProcessor errors

### LINKED ISSUE #114: feat: use transactions and rollback on error on txProcessor
Today we are not handling errors on txProcessor, so the txProcessor `onNewTx` function can fail and leave the database on an invalid state (for that specific transaction and the balances it mutates)

We should handle errors on the request and rollback all operations that are related to that transaction

andreabadesso:


src/txProcessor.ts
```diff
@@ -217,72 +220,82 @@ export const addNewTx = async (tx: Transaction, now: number, blockRewardLock: nu
...
-    || tx.version === hathorLib.constants.MERGED_MINED_BLOCK_VERSION) {
-    // unlock older blocks
-    const utxos = await getUtxosLockedAtHeight(mysql, now, tx.height);
-    await unlockUtxos(mysql, utxos, false);
+  await beginTransaction(mysql);
 
-    // set heightlock
-    heightlock = tx.height + blockRewardLock;
-  }
+  try {
```
msbrogli: I don't feel we should use the transactions with this huge try/catch block. Why don't we create a decorator that does:

```js
function dbTransactionDecorator(wrapped) {
  return async function() {
    try {
      await beginTransaction(mysql);
      const result = wrapped.apply(this, arguments);
      await commitTransaction(mysql);
      return result;
    } catch {
      await rollbackTransaction(mysql);
    }
  }
}
```

msbrogli (changes requested):

src/db.ts
```diff
@@ -2185,3 +2185,36 @@ export const getTxProposalInputs = async (
   }
   return inputs;
 };
+
+/**
```
pedroferreira1: Should we move this code to another file? This one already has > 2k lines and has the methods to connect lambdas <> db.

I feel those new methods belong to a db utils, or any other name, what do you think?

andreabadesso: I agree, done in d3421a66f29f21ca6a0ce695912cc606ede41009

src/txProcessor.ts
andreabadesso: I agree that a decorator would be better, I've just committed something like your suggestion in d3421a66f29f21ca6a0ce695912cc606ede41009

The only problem with this approach is that typescript does not support decorators yet, they do have an experimental version of decorators but only for class methods

So to be able to use the decorator pattern, I had to ignore some type errors as the decorator will expect any function as a parameter

What do you think?

msbrogli: I got it. I saw your solution and I think we shouldn't require the caller to know the decorator is necessary. But I don't think we should move forward with it because of the type errors.

In this case, maybe we should move the code to `_unsafeAddNewTx()` and create a `addNewTx()` that works as the wrapper. For example:

```js
async function addNewTx(...) {
  try {
    await beginTransaction(mysql);
    const result = await _unsafeAddNewTx(...);
    await commitTransaction(mysql);
    return result;
  } catch {
    await rollbackTransaction(mysql);
  }
}
```

What do you think?

andreabadesso: I think it is better than the huge try/catch block

I agree that it is the best solution until TypeScript supports function decorators

Done in 65b85c1215c96ca94ef5479a4e31c64da6331540

What do you think?

pedroferreira1 (approved):

src/txProcessor.ts
msbrogli: I like it. But it seems we are globally disabling a linter check. Should we do it locally?

msbrogli (approved):

PR #133, opened 2021-08-25
chore: add prometheus metrics endpoint

luislhl:
### Aceptance Criteria
- There should be a new `/metrics` endpoint in API Gateway
- The new endpoint should return the metric `wallet_service:best_block_height` by making a query in the database

### Questions
- Should we have authentication in this endpoint?

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -63,7 +63,6 @@ provider:
     DB_NAME: ${env:DB_NAME}
     DB_USER: ${env:DB_USER}
     DB_PASS: ${env:DB_PASS}
-    ENVIRONMENT: ${opt:stage, 'dev'}
     SERVICE_NAME: ${self:service}
     MAX_ADDRESS_GAP: ${env:MAX_ADDRESS_GAP}
     NETWORK: ${env:NETWORK}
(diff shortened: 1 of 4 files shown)


serverless.yml
```diff
@@ -63,6 +63,7 @@ provider:
     DB_NAME: ${env:DB_NAME}
     DB_USER: ${env:DB_USER}
     DB_PASS: ${env:DB_PASS}
+    ENVIRONMENT: ${opt:stage, 'dev'}
```
andreabadesso: Use `STAGE`? We are already using `STAGE` on `loadWallet`

andreabadesso (approved):

serverless.yml
luislhl: Will change that

andreabadesso (approved):

pedroferreira1:
> Should we have authentication in this endpoint?

Maybe not an auth but at least a rate limit or cache to prevent someone from abusing it and calling this lambda a billion times

pedroferreira1 (approved):

luislhl:
> Maybe not an auth but at least a rate limit or cache to prevent someone from abusing it and calling this lambda a billion times

Rate limit makes sense. We have an issue to make sure all endpoints have one, so I'll let this improvement to this issue.

PR #140, opened 2021-08-31
fix: force index update tx output spent by

andreabadesso:
# Acceptance criteria

The `updateTxOutputSpentBy` query should use the composite `PRIMARY` key index on (tx_id, index) when querying more than 185 elements on the IN query

## Motivation

Mainnet sync can't go past transaction `000000000f249a44a9a1759df096b18b2b25da44e5f5d2e6acec456ac79a0ab9` as that tx has 255 outputs.

Upon investigation, I found out that the `updateTxOutputSpentBy` is not using the (tx_id, index) on that transaction

For some reason, MySQL won't use the index if there are more than 185 elements, so I forced it to use it on this specific query.

I need to investigate further why this is happening, but I think we should accept this PR until then as we are only forcing the index on this specific query

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -623,8 +623,25 @@ export const updateTxOutputSpentBy = async (mysql: ServerlessMysql, inputs: TxIn
   // entries might be empty if there are no inputs
   if (entries.length) {
     // get the rows before deleting
+
+    /* We are forcing this query to use the PRIMARY index because MySQL is not using the index when there is
+     * more than 185 elements in the IN query. I couldn't find a reason for that. Here is the EXPLAIN with exactly 185
+     * elements:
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):
I was looking at this file and there are at least 5 more queries where this could be a problem. I think we should create an issue and investigate the reason for this and/or fix the other queries as well

luislhl (approved):
Odd behavior.

Should we drop a question in Stack Overflow about this? Maybe we get a free answer.

andreabadesso:
> I was looking at this file and there are at least 5 more queries where this could be a problem. I think we should create an issue and investigate the reason for this and/or fix the other queries as well

You are absolutely right. Just had this same problem on https://github.com/HathorNetwork/hathor-wallet-service/pull/170

PR #141, opened 2021-08-31
refactor: remove unecessary outputs mapping on tx proposal create

### LINKED ISSUE #121: feat: add support to create tx proposal for NFT tx

pedroferreira1:
The NFT integration in the wallet lib create another output type and the tx proposal create code was mapping the tx outputs to a p2pkh object. 

I could refactor this method to create a specific method for each output type (we also have now in the output object a `parsedScript` attribute) but this mapping is not needed. It was basically parsing and returning an array of outputs that was just sent to the wallet service, so the sender already has all the information in the response.

That's why I decided to just remove this response key instead of refactor it to make it work with all output types.

### Acceptance criteria

- Tx proposal create and send must work with an NFT transaction

### DIFF
--- a/src/api/txProposalCreate.ts
+++ b/src/api/txProposalCreate.ts
@@ -72,19 +72,6 @@ export const create = walletIdProxyHandler(async (walletId, event) => {
     return closeDbAndGetError(mysql, ApiError.TOO_MANY_OUTPUTS, { outputs: tx.outputs.length });
   }
 
-  const getToken = (tokenData, tokens) => {
-    if (tokenData === 0) {
-      return '00';
-    }
-
-    if (!tokenData) {
-      return null;
-    }
-
-    return tokens[hathorLib.wallet.getTokenIndex(tokenData) - 1];
-  };
-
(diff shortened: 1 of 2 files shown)


src/api/txProposalCreate.ts
```diff
@@ -85,13 +85,6 @@ export const create = walletIdProxyHandler(async (walletId, event) => {
   };
 
   const tokens = tx.tokens;
-  const outputs: IWalletOutput[] = tx.outputs.map((output) => ({
-    address: output.address.base58,
-    value: output.value,
-    token: getToken(output.tokenData, tokens),
```
luislhl, line 91: Should we delete the `getToken` function and the `tokens` variable above? IThey are not being used anywhere else apparently.

luislhl (approved):

src/api/txProposalCreate.ts
pedroferreira1: Deleted, thanks

andreabadesso (approved):

PR #143, opened 2021-09-01
fix: handling nft outputs on the wallet-service instead of the daemon

andreabadesso:
# Acceptance criteria

The wallet-service should ignore outputs with invalid data, like the ones from NFT transactions and store the rest of the outputs with the correct index.

## Motivation

Before this PR, the daemon was sending the list of outputs without the custom data output and the wallet-service was using the index from the iterator (`map(outputs, (output, index) => ...`) to store the output index.

We could set the output index on the daemon, but I think a better solution is to leave this to the wallet service, so we can be independent of the service that is sending transactions to it.

### DIFF
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -509,9 +509,8 @@ export const walletIdProxyHandler = (handler: WalletProxyHandler): APIGatewayPro
 );
 
 export const prepareOutputs = (outputs: TxOutput[], txId): TxOutputWithIndex[] => {
-  /* tslint:disable:no-unused-variable */
-  const [_, preparedOutputs] = outputs.reduce(
-    ([currIndex, newOutputs], output) => {
+  const preparedOutputs: [number, TxOutputWithIndex[]] = outputs.reduce(
+    ([currIndex, newOutputs]: [number, TxOutputWithIndex[]], output: TxOutput): [number, TxOutputWithIndex[]] => {
       if (!output.decoded
           || output.decoded.type === null
           || output.decoded.type === undefined) {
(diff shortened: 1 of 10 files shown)


src/commons.ts
```diff
@@ -506,3 +507,31 @@ export const walletIdProxyHandler = (handler: WalletProxyHandler): APIGatewayPro
     return handler(walletId, event, context);
   }
 );
+
+export const prepareOutputs = (outputs: TxOutput[], txId): TxOutputWithIndex[] => {
+  /* tslint:disable:no-unused-variable */
```
pedroferreira1: Why did you disable this check?

pedroferreira1 (approved):

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "lib": ["es2017"],
     "removeComments": true,
     "moduleResolution": "node",
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
r4mmer, line 6: Why make this change?

r4mmer (approved):

src/commons.ts
andreabadesso: Removed it on da41cf744db5224147a7957b3a4855e54c4c7495

tsconfig.json
andreabadesso: Removed it and typed the reduce that requried it on da41cf744db5224147a7957b3a4855e54c4c7495

r4mmer (approved):

PR #146, opened 2021-09-15
fix: websocket routes need api gateway responses

andreabadesso:
# Acceptance criteria

The user must be able to connect to websocket without receiving a 502

## Motivation

We are not answering requests to websocket routes with proper API Gateway responses. It works on serverless-offline as it is not using API Gateway to answer requests, but fails after it is deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ This action is idempotent, the lambda just responds with a `PONG` message.
 
 #### WebSocket Action: Join Wallet
 - Trigger: Client initiated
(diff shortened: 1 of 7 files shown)


luislhl (approved):

r4mmer (changes requested):
We should create a test for fail connection on redis, to make sure it aborts and does not keep the lambda hanging until it times out.

r4mmer (approved):

PR #150, opened 2021-09-26
chore: tuned sqs event sourcing batch size to have faster ws realtime…

andreabadesso:
# Acceptance criteria

Transactions should be processed by `wsTxNotifyNew` as they are received on the SQS queue in real time

## Motivation

Currently, with `batchSize: 10`, the `newTx` event is apparently only processed when there are at least 10 new transactions (to fill the batch size), so they are taking around 2 minutes to appear on the websocket feed

To solve this, I've set the [MaximumBatchingWindowInSeconds](https://docs.aws.amazon.com/pt_br/lambda/latest/dg/API_EventSourceMappingConfiguration.html) to 3s to the most it will take (before filling 10 transactions in a batch) is 3s

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -224,8 +224,8 @@ functions:
             Fn::GetAtt:
               - WalletServiceNewTxQueue
               - Arn
-          batchSize: 1
-          maximumBatchingWindow: 0
+          batchSize: 10
+          maximumBatchingWindow: 3
   wsTxNotifyUpdate:
     handler: src/ws/txNotify.onUpdateTx
     timeout: 1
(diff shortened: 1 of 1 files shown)


luislhl (approved):
This makes sense in our current situation where we have few transactions in the network.

But I think we should keep an eye on this, because in the future the network may have a lot more transactions, and having each one processed individually will increase costs. In a high volume scenario, it makes sense to have a batchSize higher than 1.

luislhl (approved):
I think the last change solves my previous concerns

r4mmer (approved):

PR #164, opened 2021-10-27
feat: added miners list endpoint

andreabadesso:
# Acceptance criteria

- We should return a list of miners
- We consider a miner an address that received at least one block or merged mining block transaction
- The API should not be exposed on the API Gateway
- We should show for each miner, the first and the last block he mined

### DIFF
--- a/db/models/miner.js
+++ b/db/models/miner.js
@@ -29,6 +29,11 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: false,
       primaryKey: false,
     },
+    count: {
+      type: DataTypes.BIGINT.UNSIGNED,
+      allowNull: false,
+      primaryKey: false,
+    },
   }, {
     sequelize,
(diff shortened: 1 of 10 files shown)


r4mmer (approved):
I think we may need pagination when listing miners if the list grows too much.

db/models/miner.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+const {
+  Model
+} = require('sequelize');
+module.exports = (sequelize, DataTypes) => {
+  class Miner extends Model {
```
pedroferreira1, line 6: Should we add in this model a count of blocks mined?

pedroferreira1 (approved):

db/models/miner.js
andreabadesso: Done in 2d0ab9feb37324a06d97e16a9f07af721f1b0c22

PR #165, opened 2021-11-08
feat: added total supply api

andreabadesso:
# Acceptance criteria

- The total supply API must return the sum of all HTR utxos, excluding the burned ones.
- The API must only be called using aws-sdk and should not have a API Gateway API


tests/db.test.ts
```diff
@@ -1722,3 +1723,32 @@ test('getMinersList', async () => {
     { address: 'address3', firstBlock: 'txId3', lastBlock: 'txId5', count: 3 },
   ]));
 });
+
+test('getTotalSupply', async () => {
```
pedroferreira1, line 1727: Should we add a utxo with burn address in the test?

pedroferreira1 (approved):

tests/db.test.ts
andreabadesso: Done in 9cef83a175806bee4220610bd34f1afcc5ad2388

src/db/index.ts
```diff
@@ -2303,3 +2303,29 @@ export const getMinersList = async (
...
+export const getTotalSupply = async (
+  mysql: ServerlessMysql,
+): Promise<number> => {
+  const results: DbSelectResult = await mysql.query(`
+    SELECT SUM(value) as value
+      FROM tx_output
+     WHERE spent_by IS NULL
+       AND token_id = '00'
+       AND voided = FALSE
+       AND address != 'HDeadDeadDeadDeadDeadDeadDeagTPgmn'`);
```
r4mmer: Should we add this address as a constant somewhere?

src/db/index.ts
```diff
@@ -2303,3 +2303,29 @@ export const getMinersList = async (
...
+ * @returns The calculated sum
+ */
+export const getTotalSupply = async (
+  mysql: ServerlessMysql,
+): Promise<number> => {
+  const results: DbSelectResult = await mysql.query(`
+    SELECT SUM(value) as value
+      FROM tx_output
+     WHERE spent_by IS NULL
+       AND token_id = '00'
```
r4mmer: Should we use `token_id` as a parameter? We could get the total supply of any token.

r4mmer (changes requested):

src/db/index.ts
andreabadesso: Added it as a constant on top of the file on commit de275e0beb1077b66e85dc50e42ea42c6e8a7158

The testnet has no burn address and tests run on mainnet so I guess this is fine for now as we don't have a constants file on this project

src/db/index.ts
andreabadesso: That's a great idea. Done in fbf73d358495d0c6d109d002ccb7d1599eb6e32f

I'm receiving the token_id on body because this is not a HTTP API exposed by the Api Gateway, it is called using the aws-sdk

r4mmer (approved):

PR #174, opened 2021-12-13
fix: added methods to unlock old timelocked utxos

andreabadesso:
### Acceptance Criteria
- Timelocked utxos should be unlocked on the first block after its timelock expiration time
- Balances should be re-calculated considering the unlocked balance from those utxos

### Motivation

The wallet service will currently only evaluate whether an timelocked utxo should be unlocked when a tx that spends this utxo is synced. This worked because transactions were not made using the wallet service, but now that transactions will be made using the `txProposalCreate` mechanism, those utxos will never be returned from the `findUtxos` lambda and so will never be unlocked.

**NOTICE**: I am running the `unlockTimelockedUtxos` method on every block transaction so the maximum delay between the actual timelock timestamp and the unlock will be the current interval between blocks.

### Security Checklist

### DIFF
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -15,8 +15,8 @@ import {
   getLatestHeight,
   getWalletBalances as dbGetWalletBalances,
   getWalletUnlockedUtxos,
+  getTimelockedUtxos,
   unlockUtxos as dbUnlockUtxos,
-  unlockTimelockedUtxos as dbUnlockTimelockedUtxos,
   updateAddressLockedBalance,
   updateWalletLockedBalance,
   getVersionData,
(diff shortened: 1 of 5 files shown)


r4mmer (approved):

r4mmer (commented):
Maybe we can have a method to update balance from `addressBalanceMap`, the process to update address balance then update all wallets affected is present in more than one place

src/commons.ts
```diff
@@ -109,6 +110,50 @@ export const unlockUtxos = async (mysql: ServerlessMysql, utxos: DbTxOutput[], u
...
+ * because of their timelocks expiring
+ *
+ * @param mysql - Database connection
+ * @param now - Current timestamp
+ */
+export const unlockTimelockedUtxos = async (mysql: ServerlessMysql, now: number): Promise<void> => {
+  const utxos: DbTxOutput[] = await dbUnlockTimelockedUtxos(mysql, now);
+  if (utxos.length === 0) return;
+
+  const outputs: TxOutput[] = utxos.map((utxo) => {
```
pedroferreira1: This map is exactly the same as in the `unlockUtxos` method, as far as I understood, and if there are differences we can parametrize them. I think would be better to have a method to map utxo -> TxOutput

src/commons.ts
```diff
@@ -109,6 +110,50 @@ export const unlockUtxos = async (mysql: ServerlessMysql, utxos: DbTxOutput[], u
...
+/**
+ * Update the unlocked/locked balances for addresses and wallets connected to the UTXOs that were unlocked
+ * because of their timelocks expiring
+ *
+ * @param mysql - Database connection
+ * @param now - Current timestamp
+ */
+export const unlockTimelockedUtxos = async (mysql: ServerlessMysql, now: number): Promise<void> => {
+  const utxos: DbTxOutput[] = await dbUnlockTimelockedUtxos(mysql, now);
+  if (utxos.length === 0) return;
```
pedroferreira1: Actually this whole method seems to be exactly the same as `unlockUtxos`. What's the difference between them?

And looks like you missed one method call `await dbUnlockUtxos(mysql, utxos);`, or you don't need this method here?

Why don't you use the same method?

pedroferreira1: I think I understand now, the method `dbUnlockTimelockedUtxos` already calls `dbUnlockUtxos`.

Why don't you change a bit and create a `getTimelockedUtxosToUnlock` instead of `dbUnlockTimelockedUtxos`? So you just get the utxos and call `unlockUtxos`, which is pretty much the same as this method here?

src/txProcessor.ts
```diff
@@ -241,6 +242,14 @@ const _unsafeAddNewTx = async (tx: Transaction, now: number, blockRewardLock: nu
 
     // add miner to the miners table
     await addMiner(mysql, blockRewardOutput.decoded.address, tx.tx_id);
+
+    // here we check if we have any utxos on our database that is locked but
+    // has its timelock < now
+    //
+    // we've decided to do this here considering that it is acceptable to have
+    // a delay between the actual timelock expiration time and the next block
+    // (that will unlock it).
```
pedroferreira1: Maybe make it clear here that we are adding this delay only in the wallet, the protocol will accept any transaction spending a timelocked output as soon as it's expired.

pedroferreira1 (changes requested):

src/commons.ts
andreabadesso: You are right.

Refactored in 6784c9aec96d79f9941822faf29c8ad2f6a19640

src/db/index.ts
```diff
@@ -2332,3 +2332,28 @@ export const getTotalSupply = async (
 
   return results[0].value as number;
 };
+
+/**
+ * Get timelocked utxos from the database
```
pedroferreira1: I don't like the docstring and the method name. You are not getting timelocked utxos, you are actually getting utxos that must be unlocked from timelock.

pedroferreira1 (changes requested):

src/txProcessor.ts
andreabadesso: Done in 530468475a3f40c1d3f00c464f62c31cbe8ec0b0

src/db/index.ts
andreabadesso: Renamed the method (and docstring) on 959e7577e4b95f74582d5e9a904a148a32af8ee6

pedroferreira1 (approved):

PR #178, opened 2022-02-03
refactor: using auth xpubkey for authorization

andreabadesso:
## Summary

Implements https://github.com/HathorNetwork/internal-issues/issues/76

### Acceptance Criteria
- The start wallet API should now require an `auth_xpubkey` to auth the user
- The start wallet API should verify if the user owns both the wallet's `xpub` and the `auth_xpubkey` he is setting for his wallet
- There should be an API to change the `auth_xpubkey`
- The authorization should now be done using the `auth_xpubkey` set when starting the wallet

### Security Checklist


src/api/auth.ts
```diff
@@ -21,6 +21,9 @@ import jwt from 'jsonwebtoken';
 import bitcore from 'bitcore-lib';
 import { ApiError } from '@src/api/errors';
 import hathorLib from '@hathor/wallet-lib';
+import { Wallet } from '@src/types';
+import { getWalletFromAuthXpub } from '@src/db';
+import { verifySignature, closeDbConnection, getDbConnection } from '@src/utils';
```
r4mmer: Maybe remove the `getWalletId` import above since it's not being used

src/api/wallet.ts
```diff
@@ -92,6 +101,125 @@ export const invokeLoadWalletAsync = async (xpubkey: string, maxGap: number): Pr
...
+      message: err.message,
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
+
+  const xpubkeyStr = value.xpubkey;
+  const authXpubkeyStr = value.authXpubkey;
+  const maxGap = parseInt(process.env.MAX_ADDRESS_GAP, 10);
```
r4mmer:
```suggestion
```

db/models/wallet.js
```diff
@@ -23,6 +23,10 @@ module.exports = (sequelize, DataTypes) => {
       type: DataTypes.STRING(120),
       allowNull: false,
     },
+    auth_xpubkey: {
```
r4mmer, line 26: Should `auth_xpubkey` be unique? We don't need this for `xpubkey` because the id is generated from the `xpubkey`.

src/api/wallet.ts
```diff
@@ -92,6 +101,124 @@ export const invokeLoadWalletAsync = async (xpubkey: string, maxGap: number): Pr
...
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
+
+  const xpubkeyStr = value.xpubkey;
+  const authXpubkeyStr = value.authXpubkey;
+
+  const timestamp = value.timestamp;
```
r4mmer, line 177: I think we should validate that timestamp is not too old.

src/api/wallet.ts
```diff
@@ -121,11 +248,42 @@ export const load: APIGatewayProxyHandler = async (event) => {
     return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
   }
 
-  const xpubkey = value.xpubkey;
+  const xpubkeyStr = value.xpubkey;
+  const authXpubkeyStr = value.authXpubkey;
   const maxGap = parseInt(process.env.MAX_ADDRESS_GAP, 10);
 
+  const timestamp = value.timestamp;
```
r4mmer, line 280: Same here

src/api/wallet.ts
```diff
@@ -121,11 +248,42 @@ export const load: APIGatewayProxyHandler = async (event) => {
...
+  const xpubkeySignature = value.xpubkeySignature;
+  const authXpubkeySignature = value.authXpubkeySignature;
+
   // is wallet already loaded/loading?
-  const walletId = getWalletId(xpubkey);
+  const walletId = getWalletId(xpubkeyStr);
+
+  if (process.env.CONFIRM_FIRST_ADDRESS === 'true') {
```
r4mmer: Maybe have logic to test the first address inside a helper since it's used on other handlers.

src/api/wallet.ts
```diff
@@ -121,11 +248,42 @@ export const load: APIGatewayProxyHandler = async (event) => {
...
+    if (firstAddress !== expectedFirstAddress) {
+      return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, {
+        message: `Expected first address to be ${expectedFirstAddress} but it is ${firstAddress}`,
+      });
+    }
+  }
+
+  const signaturesValid = validateSignatures(walletId, timestamp, xpubkeyStr, xpubkeySignature, authXpubkeyStr, authXpubkeySignature);
+
+  if (!signaturesValid) {
```
r4mmer:
```suggestion
  if (!validateSignatures(walletId, timestamp, xpubkeyStr, xpubkeySignature, authXpubkeyStr, authXpubkeySignature)) {
```

src/api/wallet.ts
r4mmer: I saw that before we didn't check the first address if the wallet already existed, do we need to check it for every request?

tests/integration.test.ts
```diff
@@ -44,7 +45,7 @@ const txId1 = 'txId1';
 block.tx_id = txId1;
 block.timestamp = now;
 block.height = 1;
-block.outputs = [createOutput(0, blockReward, 'HNwiHGHKBNbeJPo9ToWvFWeNQkJrpicYci')];
+block.outputs = [createOutput(0, blockReward, 'HBCQgVR8Xsyv1BLDjf9NJPK1Hwg4rKUh62')];
```
r4mmer: Why not use `ADDRESSES[0]`? same for other addresses on this file

tests/utils.ts
```diff
@@ -12,8 +12,10 @@ import { WalletBalanceEntry, AddressTableEntry, TokenTableEntry } from '@tests/t
 
 import { RedisClient } from 'redis';
 
+export const TEST_SEED = 'neither image nasty party brass oyster treat twelve olive menu invest title fan only rack draw call impact use curtain winner horn juice unlock';
```
r4mmer, line 15: Should we have a test to check that the xpub, auth_xpub and addresses are generated from this seed correctly?

r4mmer (changes requested):

src/api/auth.ts
andreabadesso: Fixed in 7f01b12080229859bc6796c39cc26b6a853864b9, thanks

src/api/wallet.ts
andreabadesso: Why?

db/models/wallet.js
andreabadesso: Wow, you are right, I hadn't thought about it on the design

I don't like the idea of it being unique, I think that maybe we should require the user to send the `walletId` of his wallet and store it (encrypted with the PIN) on the wallet-lib

What do you think, @r4mmer and @pedroferreira1 ?

pedroferreira1: I don't see any reasons why someone would like to use the same auth xpub for more than one wallet. Why you don't like it being unique? I feel it's natural to be unique.

I was thinking about multisig or any other reasons to reuse the auth xpub but nothing came to my mind. But also if you decide to go with `walletId` stored and sending it in the request it seems ok as well (but I think I prefer setting it as unique).

pedroferreira1: What's the reason to have an API to change the auth xpub? Maybe this is a good reason to allow repeated xpubs here

src/db/index.ts
```diff
@@ -170,6 +175,32 @@ export const getWallet = async (mysql: ServerlessMysql, walletId: string): Promi
...
+};
+
+/**
+ * Get wallet data from authXpub
+ *
+ * @param mysql - Database connection
+ * @param walletId - The wallet id
+ * @returns The wallet information or null if it was not found
+ */
+export const getWalletFromAuthXpub = async (mysql: ServerlessMysql, authXpub: string): Promise<Wallet> => {
```
pedroferreira1: If we have this method here, then I think we must require the `authXpub` to be unique. If we don't do that, we should change this method to get from `walletId` only

src/db/index.ts
```diff
@@ -170,6 +175,32 @@ export const getWallet = async (mysql: ServerlessMysql, walletId: string): Promi
...
+  const results: DbSelectResult = await mysql.query('SELECT * FROM `wallet` WHERE `auth_xpubkey` = ?', authXpub);
+  if (results.length) {
+    const result = results[0];
+    return {
```
pedroferreira1: Should we create a method to get pretty return of wallet data? If we add a new field in the future we must remember to add it in all returns.

tests/api.test.ts
```diff
@@ -551,72 +631,209 @@ test('POST /wallet', async () => {
...
+  const derivedPrivKey = walletUtils.deriveXpriv(xpriv, accountDerivationIndex);
+  const address = derivedPrivKey.publicKey.toAddress(network.getNetwork()).toString();
+  const message = new bitcore.Message(String(now).concat(walletId).concat(address));
+  const xpubkeySignature = message.sign(derivedPrivKey.privateKey);
+
+  // auth purpose path (m/280'/280')
+  const authDerivedPrivKey = HathorWalletServiceWallet.deriveAuthXpriv(xpriv);
+  const authAddress = authDerivedPrivKey.publicKey.toAddress(network.getNetwork());
+  const authMessage = new bitcore.Message(String(now).concat(walletId).concat(authAddress));
+  const authXpubkeySignature = authMessage.sign(authDerivedPrivKey.privateKey);
```
pedroferreira1, line 1029: Should we have a test with a wrong signature here and the in the other xpubkey? Just so we can confirm that we are only creating the wallet with both signatures valid

src/api/wallet.ts
r4mmer: The `maxGap` was not being used

andreabadesso: Oh, this was on the `changeAuthXpub` method, right?

I must have removed it before reading your comment, thanks

src/api/wallet.ts
andreabadesso: Good idea, done in f053b93eaf64af3f61919142148b637efda6cc76

src/api/wallet.ts
andreabadesso: Done in 3255889978e41c2b5df0daee084a753e61e90f48

src/api/wallet.ts
andreabadesso: You are right, we don't need to, and those calls are expensive

Fixed in 64f6fe4594d72462ef4885a283a0afec677635e0

tests/integration.test.ts
andreabadesso: Way better, thanks

Done in 7015d74648ab042a6ab1ead0533a2e66316956c1

db/models/wallet.js
andreabadesso: I don't quite remember why I thought it would be a bad idea to set the `auth_xpubkey` to be unique, I guess I was thinking that since we don't enforce the user to use the different purpose path, he could use anything he wants to and that could lead to two users using the same `auth_xpubkey`, but both would have to prove they own it, so I dont think this is a problem

I have set the column as unique on b0ff81bac9db71f955b9cef41832ceb988d52258

src/api/wallet.ts
andreabadesso: I agree, done in e16ea1c38e44364cdb87a6fb8d703998f4925634

src/api/wallet.ts
andreabadesso: I agree, done in e16ea1c38e44364cdb87a6fb8d703998f4925634

src/db/index.ts
andreabadesso: We should, done in 35069d8d69b4eb1b04fe4da9d8b1e14848a41cbe

tests/api.test.ts
andreabadesso: We should, done in 73488047aac3920280f71eca8bfee84c669919eb

Also fixed a potential problem with invalid user input

tests/utils.ts
andreabadesso: That's a great idea, thanks!

Done in 83d8bb74a92033529a6d1b6e94ed55753cc8cb57

r4mmer (approved):

db/models/wallet.js
andreabadesso: @r4mmer 

I've refactored the `tokenHandler` to accept the `walletId` as a parameter

This way, we don't have to set the `auth_xpubkey` to unique

The problem with setting `auth_xpubkey` as unique is that someone might want to use multiple wallets from the same seed

This is the refactor commit: 06f4126b4d167231c83fb71369f12bdd7e2b1d1b

And this is the commit on the wallet-lib that sends the wallet-id:
https://github.com/HathorNetwork/hathor-wallet-lib/pull/316/commits/cbea7e52e3c702879eae4f17d5b0d1bfb8bbabc5

src/db/index.ts
andreabadesso: This method has been removed as we are not using it anymore since the refactor on 06f4126b4d167231c83fb71369f12bdd7e2b1d1b

pedroferreira1 (approved):

Issue #185, opened 2022-03-11
feat: create migration file on Sequelize to include two new columns


lucas3003:
### Description

As pointed on the [Token API design document](https://github.com/HathorNetwork/hathor-explorer-service/issues/123), we will need to make a migration to include modification_time and insertion_time as new columns on Token database table.

### Acceptance Criteria

Token table must have two new columns: modification_time and insertion_time. These will be used by Logstash to push data to ElasticSearch.

PR #186, opened 2022-03-16
feat: add created_at and updated_at as token columns

lucas3003:
### Acceptance Criteria

This PR will include ```modification_time``` and ```insertion_time``` as part of Token table columns. This will be necessary to Logstash when ingesting data and sending it to Elastic Search.

This is how the records will look like:

Screen Shot 2022-03-16 at 16 43 43

Also, a test was performed locally to check how long this migration should take on 300,000 records, and it took 0.145s.

Closes issue #185

### Security Checklist


tests/db.test.ts
```diff
@@ -1112,6 +1112,23 @@ test('storeTokenInformation and getTokenInformation', async () => {
...
+test('validateTokenTimestamps', async () => {
+  expect.hasAssertions();
+
+  const info = new TokenInfo('tokenId', 'tokenName', 'TKNS');
+  storeTokenInformation(mysql, info.id, info.name, info.symbol);
+  let result = await mysql.query('SELECT * FROM `token` WHERE `id` = ?', [info.id]);
+
+  expect(result[0].insertion_time).toStrictEqual(result[0].modification_time);
+
+  await new Promise((r) => setTimeout(r, 3000));
```
andreabadesso: I don't like this delay here, I think tests should run as fast as possible

Can you use fake timers somehow? Or at least decrease the time

db/migrations/20220315192400-add-timestamp-fields-token.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+
```
andreabadesso: Extra line

db/models/token.js
```diff
@@ -27,6 +27,16 @@ module.exports = (sequelize, DataTypes) => {
       type: DataTypes.STRING(5),
       allowNull: false,
     },
+    insertion_time: {
```
andreabadesso: We are not going to backfill for now, right?

andreabadesso (changes requested):

db/models/token.js
lucas3003: The idea is that all existing records must be populated with CURRENT_TIMESTAMP on both insertion_time and modification_time

tests/db.test.ts
lucas3003: I couldn't use fake timers because they mock timers used on the tested code, which is not this case. I agree this timeout is too long, and I will reduce it to the lowest possible value (1100 ms).

andreabadesso (approved):

pedroferreira1 (approved):

PR #188, opened 2022-03-16
refactor: using bitcoinjs instead of bitcore-lib

andreabadesso:
## Motivation

I profiled the app using both `bitcore-lib` and `bitcoinjs` and the latter showed a big performance boost on the lambdas using the refactored in this PR. Here are some results:

### authTokenApi

`bitcore-lib`:

```
Duration: 2294.66 ms	Billed Duration: 1591 ms	 Memory Size: 256 MB	Max Memory Used: 98 MB	Init Duration: 703.73 ms
```

`bitcoinjs-lib`:
```
Duration: 596.45 ms	Billed Duration: 597 ms	Memory Size: 256 MB	Max Memory Used: 113 MB
```

### bearerAuthorizer

`bitcore-lib`:

```
Duration: 403.07 ms	Billed Duration: 404 ms	Memory Size: 256 MB	Max Memory Used: 108 MB	
```

`bitcoinjs-lib`:

```
Duration: 23.91 ms	Billed Duration: 24 ms	Memory Size: 256 MB	Max Memory Used: 112 MB
```

### Acceptance Criteria
- We should use bitcoinjs for the following methods, instead of using the `wallet-lib`'s correspondent methods:
  - `getAddressFromXpub`
  - `verifySignature`
  - `xpubDeriveChild`
  - `getAddresses`
  - `getAddressAtIndex`

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "bip32": "^3.0.1",
     "bitcoinjs-lib": "^6.0.1",
     "bitcoinjs-message": "^2.2.0",
-    "bitcore-lib": "^8.25.25",
     "joi": "^17.4.0",
     "jsonwebtoken": "^8.5.1",
     "mysql": "^2.18.1",
@@ -67,6 +66,7 @@
     "typescript": "^3.9.5",
     "typescript-eslint": "0.0.1-alpha.0",
     "webpack": "^4.29.0",
-    "webpack-node-externals": "^1.7.2"
+    "webpack-node-externals": "^1.7.2",
+    "bitcore-lib": "^8.25.25"
   }
 }
--- a/package.json
+++ b/package.json
@@ -23,6 +23,9 @@
     "aws-lambda": "^1.0.6",
     "aws-sdk": "^2.916.0",
     "axios": "^0.21.1",
+    "bip32": "^3.0.1",
+    "bitcoinjs-lib": "^6.0.1",
+    "bitcoinjs-message": "^2.2.0",
     "bitcore-lib": "^8.25.25",
     "joi": "^17.4.0",
     "jsonwebtoken": "^8.5.1",
@@ -33,6 +36,7 @@
     "sequelize": "^6.6.4",
     "serverless-mysql": "^1.5.4",
     "source-map-support": "^0.5.19",
+    "tiny-secp256k1": "^2.2.1",
     "uuid": "^8.3.0"
   },
   "devDependencies": {
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -180,6 +180,7 @@ const _generatePolicy = (principalId: string, effect: string, resource: string) 
   authResponse.context = context;
 
   // XXX: to get the resulting policy on the logs, since we can't check the cached policy
+  // eslint-disable-next-line
   console.info('Generated policy:', authResponse);
   return authResponse;
 };
@@ -207,6 +208,7 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = async (event) 
     } else if (e.name === 'TokenExpiredError') {
       throw new Error('Unauthorized');
     } else {
+      // eslint-disable-next-line
       console.log('Error on bearerAuthorizer: ', e);
       throw e;
     }
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -22,6 +22,7 @@ import { Wallet } from '@src/types';
 import { getWallet } from '@src/db';
 import {
   verifySignature,
+  getAddressFromXpub,
   closeDbConnection,
   getDbConnection,
   validateAuthTimestamp,
@@ -112,8 +113,7 @@ export const tokenHandler: APIGatewayProxyHandler = async (event) => {
     };
   }
 
-  const xpubkey = bitcore.HDPublicKey(authXpubStr);
-  const address = xpubkey.publicKey.toAddress(hathorLib.network.getNetwork());
+  const address = getAddressFromXpub(authXpubStr);
   const walletId = wallet.walletId;
 
   if (!verifySignature(signature, timestamp, address, walletId)) {
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -15,9 +15,7 @@ import {
 import { v4 as uuid4 } from 'uuid';
 import Joi from 'joi';
 import jwt from 'jsonwebtoken';
-import bitcore from 'bitcore-lib';
 import { ApiError } from '@src/api/errors';
-import hathorLib from '@hathor/wallet-lib';
 import { Wallet } from '@src/types';
 import { getWallet } from '@src/db';
 import {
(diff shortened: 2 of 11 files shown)


src/utils.ts
```diff
@@ -9,8 +9,31 @@ import { createHash, HexBase64Latin1Encoding } from 'crypto';
...
+  },
+  testnet: {
+    messagePrefix: '\x18Hathor Signed Message:\n',
+    bech32: 'tn',
+    bip32: { public: 76067358, private: 70568132 },
+    pubKeyHash: 40,
+    scriptHash: 100,
+    wif: 128,
+  },
+};
```
r4mmer: Should we get this from the wallet-lib?
And is the `bech32` necessary for bitcoinjs to work? Because we don't have this defined on Hathor.

package.json
```diff
@@ -23,6 +23,9 @@
     "aws-lambda": "^1.0.6",
     "aws-sdk": "^2.916.0",
     "axios": "^0.21.1",
+    "bip32": "^3.0.1",
```
pedroferreira1, line 26: Why do you need this library as a direct dependency? I thought we would add only `bitcoinjs-lib`

package.json
```diff
@@ -33,6 +36,7 @@
     "sequelize": "^6.6.4",
     "serverless-mysql": "^1.5.4",
     "source-map-support": "^0.5.19",
+    "tiny-secp256k1": "^2.2.1",
```
pedroferreira1, line 38: Why do you need this library as a direct dependency? I thought we would add only `bitcoinjs-lib`

package.json
```diff
@@ -23,6 +23,9 @@
     "aws-lambda": "^1.0.6",
     "aws-sdk": "^2.916.0",
     "axios": "^0.21.1",
+    "bip32": "^3.0.1",
+    "bitcoinjs-lib": "^6.0.1",
+    "bitcoinjs-message": "^2.2.0",
     "bitcore-lib": "^8.25.25",
```
pedroferreira1: Can we remove it?

src/api/auth.ts
```diff
@@ -30,8 +29,6 @@ import {
 
 const EXPIRATION_TIME_IN_SECONDS = 1800;
 
-hathorLib.network.setNetwork(process.env.NETWORK);
```
pedroferreira1, line 33: Are we still using `hathorLib`? If yes, I think we should keep this. If not, we should remove it from the package.json

src/api/wallet.ts
```diff
@@ -129,14 +128,11 @@ export const validateSignatures = (
   authXpubkeySignature: string,
 ): boolean => {
   // verify that the user owns the xpubkey
-  const xpubkey = bitcore.HDPublicKey(xpubkeyStr);
-  const xpubAddress = xpubkey.publicKey.toAddress(network.getNetwork());
-  const xpubValid = verifySignature(xpubkeySignature, timestamp, xpubAddress, walletId.toString());
+  const xpubAddress = getAddressFromXpub(xpubkeyStr);
+  const xpubValid = verifySignature(xpubkeySignature, timestamp, xpubAddress.toString(), walletId.toString());
```
pedroferreira1: Why do you need `xpubAddress.toString()` here and in the `src/api/auth.ts` file you don't need to use `.toString` when verifying signature?

Below you don't use `toString` as well.

src/utils.ts
```diff
@@ -260,3 +258,106 @@ export const validateAuthTimestamp = (timestamp: number, now: number): [boolean,
...
+    return false;
+  }
+};
+
+/**
+ * Returns an address (as a string) from a string xpubkey
+ *
+ * @param xpubkey - The xpubkey
+ *
+ * @returns the address derived from the xpubkey
```
pedroferreira1, line 349: It returns a string, right? Just confirming because one method is still doing a `toString` after getting the value from this method here

src/utils.ts
```diff
@@ -260,3 +258,106 @@ export const validateAuthTimestamp = (timestamp: number, now: number): [boolean,
...
+
+/**
+ * Returns an address from a xpubkey on a specific index
+ *
+ * @param xpubkey - The xpubkey
+ * @param index - The address index to derive
+ *
+ * @returns The derived address
+ */
+export const getAddressAtIndex = (pubkey: string, addressIndex: number): string => {
```
r4mmer: Just to be sure, is it the `pubkey` or `xpubkey`?

src/utils.ts
```diff
@@ -260,3 +258,106 @@ export const validateAuthTimestamp = (timestamp: number, now: number): [boolean,
...
+
+/**
+ * Derives a xpubkey at a specific index
+ *
+ * @param xpubkey - The xpubkey
+ * @param index - The index to derive
+ *
+ * @returns The derived xpubkey
+ */
+export const xpubDeriveChild = (pubkey: string, index: number): string => (
```
r4mmer: Same here

package.json
andreabadesso: `bip32` [is a library from `bitcoinjs`](https://github.com/bitcoinjs/bip32), they separated the bip32 implementation to a different lib

package.json
andreabadesso: Because `bitcoinjs-lib` does not have an implementation for secp256k1, it leaves the choice of the library to use to the user

I chose `tiny-secp256k1` because it uses web assembly

package.json
andreabadesso: We are using it on the tests

We could replace it with `bitcoinjs-lib` on the tests as well, but since the mobile wallets will still use bitcore-lib, I think it is best if we generate signatures using it

What do you think?

src/api/auth.ts
andreabadesso: We are still using `hathorLib` on the project, but not on this lambda. Why would we keep the `setNetwork` here?

src/api/wallet.ts
andreabadesso: We don't need it, it is already a string. Thanks

Fixed in 5a1d014889a4e6492926a18b041f5c59a421ac07

src/utils.ts
andreabadesso: Right, the .toString was pointless, fixed in 5a1d014889a4e6492926a18b041f5c59a421ac07

src/utils.ts
andreabadesso: It should be `xpubkey`, thanks

Fixed in 1ff7534d77e46167589680ad6e8e7e72a9e7511d

src/utils.ts
andreabadesso: Same here, fixed in 1ff7534d77e46167589680ad6e8e7e72a9e7511d

src/utils.ts
andreabadesso: Refactored in ecea6a0f3d4d607de0b6b9d0ee6f5b20d07bfd5d

I don't think `bech32` is necessary for the operations we are doing, but it is a required param for the `Network` object

r4mmer (approved):

src/api/auth.ts
pedroferreira1: Yeah, no need to keep it in this lambda.

package.json
andreabadesso: Moved it to a dev dependency on f2bba4621b4df9488a2c15f2d3f2333974348374

pedroferreira1 (approved):

PR #199, opened 2022-04-08
feat: added version to tx_history api

andreabadesso:
## Motivation

The `wallet-desktop` expects a `version` attribute on the transactions to display the type of transaction properly on the transaction history screen

### Acceptance Criteria
- Added `version` to the transaction object on `tx_history` API response

### Security Checklist

### DIFF
--- a/tests/api.test.ts
+++ b/tests/api.test.ts
@@ -29,6 +29,7 @@ import {
   addToWalletBalanceTable,
   addToWalletTable,
   addToWalletTxHistoryTable,
+  addToTransactionTable,
   cleanDatabase,
   makeGatewayEvent,
   makeGatewayEventWithAuthorizer,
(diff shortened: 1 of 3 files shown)


tests/api.test.ts
```diff
@@ -375,6 +376,10 @@ test('GET /txhistory', async () => {
     ['my-wallet', 'tx2', '00', 7, 1001, false],
     ['my-wallet', 'tx2', 'token3', 7, 1001, true],
   ]);
+  await addToTransactionTable(mysql, [
+    ['tx1', 100, 2, false, null],
+    ['tx2', 100, 2, false, null],
```
pedroferreira1: Maybe add test with another version as well

pedroferreira1 (approved):

r4mmer (approved):
I know that we already have this information on another table and searching for it makes sense, but why not add the version on the `wallet_tx_history` table?

This would involve more work to update, but the update of old data can be done in the migration and it would be faster to get the tx history (a query that will be made very frequently)

tests/api.test.ts
andreabadesso: Done in 52fa4cf4d8c8eb2180e339d829b0549bc81cc6d7

andreabadesso:
> I know that we already have this information on another table and searching for it makes sense, but why not add the version on the `wallet_tx_history` table?
> 
> This would involve more work to update, but the update of old data can be done in the migration and it would be faster to get the tx history (a query that will be made very frequently)

That's a good point, I would just suggest to do it in another PR, I will create an issue for that

pedroferreira1 (approved):

PR #200, opened 2022-04-08
refactor: utxo api should now be a tx_output api

andreabadesso:
## Motivation

The wallet-desktop needs to decide if a wallet ever had an authority to decide whether to display the administrative tab or not, so we should be able to query utxos that have already been spent (even if that means that it is no longer an utxo)

### Acceptance Criteria
- We should accept a `skipSpent` parameter to return `tx_outputs` that were already spent
- The `skipSpent` parameter should default to `true` (since this is an `utxo` API)

### Security Checklist


src/types.ts
```diff
@@ -630,6 +630,7 @@ export interface IFilterUtxo {
   biggerThan?: number;
   smallerThan?: number;
   maxUtxos?: number;
+  skipSpent?: boolean;
```
pedroferreira1, line 633: Unrelated to this line.

We should add `spentBy` also in the `DbTxOutput` interface

pedroferreira1 (approved):

r4mmer (approved):

src/types.ts
andreabadesso: We already have it

andreabadesso:
@pedroferreira1 @r4mmer 
I did a big refactor on this PR, basically, there are now two APIs with the same implementation: The `getFilteredTxOutputs` API that will accept the `skipSpent` parameter and a `getFilteredUtxos` API that will always have `true` for the `skipSpent` parameter.

They will both call the same internal method `_getFilteredTxOutputs`, so the implementation should be the same

The idea is to maintain backwards compatibility with the already deployed wallet-service enabled mobile app and deprecate the `getFilteredUtxos` lambda after we migrate the lib to use the `getFilteredTxOutputs` method and make sure that all our users upgraded their mobile apps

Can you please review again?

src/api/txOutputs.ts
```diff
@@ -31,15 +31,16 @@ const bodySchema = Joi.object({
...
+  skipSpent: Joi.boolean().optional().default(true),
   txId: Joi.string().optional(),
   index: Joi.number().optional().min(0),
 }).and('txId', 'index');
 
 /*
  * Filter utxos
  *
- * This lambda is called by API Gateway on GET /wallet/utxos
+ * This lambda is called by API Gateway on GET /wallet/tx_outputs
```
pedroferreira1: This one is still `/wallet/utxos`, right?

src/api/txOutputs.ts
```diff
@@ -71,73 +73,131 @@ export const getFilteredUtxos = walletIdProxyHandler(async (walletId, event) =>
...
+/*
+ * Filter tx_outputs
+ *
+ * This lambda is called by API Gateway on GET /wallet/tx_outputs
+ */
+export const getFilteredTxOutputs = walletIdProxyHandler(async (walletId, event) => {
+  const multiQueryString = event.multiValueQueryStringParameters || {};
+  const queryString = event.queryStringParameters || {};
+
+  const eventBody = {
```
pedroferreira1, line 104: Can you use here something like this?

```
const eventBody = { ...queryString, addresses: multiQueryString.addresses }
```

If this is better we could also change above

pedroferreira1 (approved):

src/api/txOutputs.ts
andreabadesso: I prefer to be strict on the attributes we get from the `queryString` as they come from user input to prevent Joi from having to run checks on multiple parameters

The user can send whatever he wants on the query string, we will just use what we need

What do you think?

Your suggestion would work though

andreabadesso:
@pedroferreira1 and @r4mmer  Did a small refactor on a837207521d927ddbf1494b14dffe19c6ed92eb2


The `tx_outputs` API accepts a `(tx_id, index)` query to return a single `tx_output`, I refactored the db method that was called on that case (from `getUtxo` to `getTxOutput`) and added a `skipSpent` parameter to it, so when we are querying for utxos, we return `null` even if we have the `tx_id, index` on our database

src/api/txOutputs.ts
andreabadesso: You are absolutely correct, fixed in d9bb0d3594c5ac15e314cb5eedc427809474460f

Also added a deprecation notice

andreabadesso:
@pedroferreira1 and @r4mmer Did another small refactor on this API to return an empty array when the `tx_output` is not found: 90cea8598aa99d1a184730f2aac478c98f1e310a

src/api/txOutputs.ts
pedroferreira1: Makes sense, we can keep it like this.

r4mmer (approved):

PR #202, opened 2022-04-11
feat: token info api

andreabadesso:
## Motivation

Currently, our wallet desktop is querying the fullnode for token details, we need an API to return the information necessary to show token details on the "About" section of our wallet.

### Important

We are currently limiting the max number of authorities returned by this API to `100`. This was done to prevent users that created a large number of authorities (since this is a free operation) to spam this API, causing a lot of bandwidth use

### Acceptance Criteria

- We should have an API to return token details
- The API should require the user to be authenticated
- This API should have the following information:
  - Total supply of the requested token
  - Total number of transactions on the requested token
  - List of unlocked utxos with authorities for the requested token
  - Details (uid, name, symbol) for the requested token

### Security Checklist


tests/utils.ts
```diff
@@ -767,3 +767,40 @@ export const redisCleanup = (
 ): void => {
   client.flushdb();
 };
+
+export const getAuthData = (now: number): any => {
```
andreabadesso, line 771: This is just a refactor I did but ended up not using anywhere else than the `loadWallet should update wallet status to ERROR if an error occurs` test

src/db/index.ts
```diff
@@ -2389,3 +2389,65 @@ export const getExpiredTimelocksUtxos = async (
...
+   */
+  const results: DbSelectResult = await mysql.query(`
+  SELECT *
+    FROM tx_output
+   WHERE authorities > 0
+     AND token_id = ?
+     AND voided = FALSE
+     AND locked = FALSE
+     AND spent_by IS NULL
+   LIMIT 100
```
pedroferreira1: I think we shouldn't have this limit here. Why don't you create another method `hasMintAuthority` and `hasMeltAuthority`?

With this method you might be returning the wrong information to the wallet desktop. If you have 100 mint authorities and 1 melt authority, then you might be returning only the mint and will show in the desktop wallet that there are no melt authorities.

I think would be better to create a method that we can trust and fulfills the needs that we have.

src/db/index.ts
```diff
@@ -2389,3 +2389,65 @@ export const getExpiredTimelocksUtxos = async (
...
+
+/**
+ * Get the total sum of transactions for a given tokenId
+ *
+ * @param mysql - Database connection
+ * @param tokenId - The token id to fetch transactions
+
+ * @returns The calculated total sum of transactions
+ */
+export const getTotalTransactions = async (
```
pedroferreira1, line 2452: When you count the `tx_output` table you might be ignoring some transactions, if I'm not wrong.

One example is a destroy authority transaction, that you have inputs but doesn't have outputs.

Another one is a possible situation where you have a transaction with a single output that is a `DataScript`. Given that you don't store data script outputs in this table, you would also be ignoring this transaction.

Don't you have another table to gather this information? Maybe the tx_history table?

src/api/tokens.ts
```diff
@@ -22,3 +35,52 @@ export const get = walletIdProxyHandler(async (walletId) => {
...
+      message: err.message,
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
+
+  const tokenId = value.token_id;
+
+  const tokenInfo: TokenInfo = await getTokenInformation(mysql, tokenId);
```
pedroferreira1, line 68: If the tokens does not exist in the database what happens? Should we get this info here and return 400? Better than doing all the other queries.

pedroferreira1 (changes requested):

src/api/tokens.ts
```diff
@@ -22,3 +35,52 @@ export const get = walletIdProxyHandler(async (walletId) => {
...
+  const totalSupply: number = await getTotalSupply(mysql, tokenId);
+  const totalTransactions: number = await getTotalTransactions(mysql, tokenId);
+  const availableAuthorities: DbTxOutput[] = await getAvailableAuthorities(mysql, tokenId);
```
r4mmer: Should we make these queries as a `Promise.all`?

src/db/index.ts
andreabadesso: You are absolutely right, we can use the `address_tx_history` table.

Done in 926a4b8a7c60faec094ecd6bf3ef67666dd9e4e5

I also had to add an index on the `token_id` column to the `address_tx_history` table, this was done in c36a00b541830c54d52b0da82624e83a3d7c8de4

src/api/tokens.ts
andreabadesso: You are right, done in 9ef91e72f5a51ca3fbe5777f14247e18cbe2b1c9

src/api/tokens.ts
andreabadesso: We absolutely should, thanks!

Done in 2e48bad426314e9f756a48bf210180cf1e8d1943

src/db/index.ts
andreabadesso: I updated the API to return `true` or `false` on the authorities object, can you please check again? I would link the commit here but it was done on multiple commits as it is a bigger refactor 

I removed the LIMIT from the `getAvailableAuthorities` but didn't remove the method, even though it isn't being used anymore I feel that it might in the future

src/db/index.ts
```diff
@@ -716,6 +716,41 @@ export const getUtxo = async (
...
+ * @param tokenId - The token id to search authorities for
+ * @param authority - The authority to search for, can be one of (TOKEN_MINT_MASK, TOKEN_MELT_MASK)
+ *
+ * @returns The requested UTXO
+ */
+export const getAuthorityUtxo = async (
+  mysql: ServerlessMysql,
+  tokenId: string,
+  authority: number,
+): Promise<DbTxOutput> => {
```
pedroferreira1: Should this be `Promise<DbTxOutput | null>`? If we have no authorities, we return `null`, right?

tests/api.test.ts
```diff
@@ -1076,3 +1063,89 @@ test('loadWallet should update wallet status to ERROR if an error occurs', async
...
+
+  await addToTokenTable(mysql, [
+    { id: token1.id, name: token1.name, symbol: token1.symbol },
+    { id: token2.id, name: token2.name, symbol: token2.symbol },
+  ]);
+
+  await addToUtxoTable(mysql, [
+    ['txId', 0, token1.id, ADDRESSES[0], 100, 0, null, null, false, null], // total tokens created
+    ['txId', 1, token1.id, ADDRESSES[0], 0, constants.TOKEN_MINT_MASK, null, null, false, null], // mint
+    ['txId', 2, token1.id, ADDRESSES[0], 0, constants.TOKEN_MINT_MASK, null, null, false, null], // melt
```
pedroferreira1: The mask is mint but the comment says melt

pedroferreira1 (approved):

src/db/index.ts
andreabadesso: Fixed in 917c25e0e495b49576e8f7844a7d05970545ff42

tests/api.test.ts
andreabadesso: Fixed in 917c25e0e495b49576e8f7844a7d05970545ff42

r4mmer (approved):

PR #209, opened 2022-04-28
refactor: adding CORs headers to all responses using middy

andreabadesso:
# Motivation

The `wallet-desktop` requests to the wallet-service gets a [cross-origin request blocked error](https://developer.mozilla.org/pt-BR/docs/Web/HTTP/CORS/Errors) if the server does not respond with a `Access-Control-Allow-Origin` header on the response

I'm also adding a very nice [middleware library](https://github.com/middyjs/middy) that we can use for other things in the wallet service, like parameter validation, MySQL connection sharing, authorization, and others later on

### Acceptance Criteria
- Our responses should include the `Access-Control-Allow-Origin` header and it should be set to `*`

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,8 @@
   },
   "dependencies": {
     "@hathor/wallet-lib": "^0.31.0",
+    "@middy/core": "^2.5.7",
+    "@middy/http-cors": "^2.5.7",
     "@types/redis": "^2.8.28",
     "aws-lambda": "^1.0.6",
     "aws-sdk": "^2.916.0",
@@ -39,11 +41,12 @@
     "uuid": "^8.3.0"
   },
   "devDependencies": {
-    "@types/aws-lambda": "^8.10.76",
+    "@types/aws-lambda": "^8.10.95",
     "@types/jest": "^26.0.24",
     "@types/node": "^10.12.18",
     "@typescript-eslint/eslint-plugin": "^3.3.0",
     "@typescript-eslint/parser": "^3.3.0",
+    "bitcore-lib": "^8.25.25",
     "dotenv": "^10.0.0",
     "eslint": "^7.27.0",
     "eslint-config-airbnb-base": "^14.2.1",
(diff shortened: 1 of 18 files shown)


serverless.yml
```diff
@@ -30,6 +30,15 @@ plugins:
 
 resources:
   Resources:
+    GatewayResponseDefault4XX:
```
andreabadesso, line 36: This is needed to be able to include the CORs headers on our custom authorizer

Taken from:
https://serverless.com/blog/cors-api-gateway-survival-guide/

package.json
```diff
@@ -39,11 +41,12 @@
     "uuid": "^8.3.0"
   },
   "devDependencies": {
-    "@types/aws-lambda": "^8.10.76",
+    "@types/aws-lambda": "^8.10.95",
     "@types/jest": "^26.0.24",
     "@types/node": "^10.12.18",
     "@typescript-eslint/eslint-plugin": "^3.3.0",
     "@typescript-eslint/parser": "^3.3.0",
+    "bitcore-lib": "^8.25.25",
```
r4mmer, line 49: Why are we adding bitcore-lib as a dependency?

andreabadesso: We are not adding it in this PR, the diff is just a change of position in the object keys order

We have bitcore lib as a dev dependency to be able to sign messages in the same way the lib does, but we don't use it in production

tests/api.test.ts
```diff
@@ -1161,3 +1210,38 @@ test('GET /wallet/tokens/token_id/details', async () => {
...
+  // test CORS headers
+  await _testCORSHeaders(txProposalSend, null, null);
+});
+
+test('DELETE /tx/proposal/{txProposalId}', async () => {
+  expect.hasAssertions();
+
+  // test CORS headers
+  await _testCORSHeaders(txProposalDestroy, null, null);
+});
```
andreabadesso: Those methods are being tested on different test files, so I 'm just adding tests for the CORs headers here to make sure they are all responding with the proper headers

serverless.yml
luislhl: Maybe we should add this as a comment in the code

r4mmer (approved):

luislhl (approved):

PR #220, opened 2022-05-10
feat: better monitoring and integrity checks

andreabadesso:
### Acceptance Criteria
- We should improve logs for the `onNewTxRequest`, `onHandleReorg`, `onHandleOldVoidedTxs`
- We should wrap the `handleReorg` method with a mysql transaction

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -7,7 +7,7 @@ custom:
   logLevelMap:  
     mainnet: info
     testnet: info
-    dev: info
+    dev-testnet: info
   logLevel: ${self:custom.logLevelMap.${opt:stage}, 'debug'}
   webpack:
     webpackConfig: ./webpack.config.js
--- a/serverless.yml
+++ b/serverless.yml
@@ -3,7 +3,12 @@ frameworkVersion: '2'
 
 useDotenv: true
 
-custom:
+custom: 
+  logLevelMap:  
+    mainnet: info
+    testnet: info
+    dev: info
+  logLevel: ${self:custom.logLevelMap.${opt:stage}, 'debug'}
   webpack:
     webpackConfig: ./webpack.config.js
     includeModules: true
@@ -94,6 +99,7 @@ provider:
     DEFAULT_SERVER: ${env:DEFAULT_SERVER}
     WS_DOMAIN: ${env:WS_DOMAIN}
     TX_HISTORY_MAX_COUNT: ${env:TX_HISTORY_MAX_COUNT}
+    LOG_LEVEL: ${self:custom.logLevel}
   iamRoleStatements:
     - Effect: Allow
       Action:
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -54,6 +54,7 @@ import {
   AddressTotalBalance,
   WalletProxyHandler,
 } from '@src/types';
+import { Logger } from 'winston';
 
 import {
   getUnixTimestamp,
@@ -350,19 +351,22 @@ export const searchForLatestValidBlock = async (mysql: ServerlessMysql): Promise
  * @returns A new list of voided transactions that are linked to the received list and a list of tx_outputs affected
  * by this iteration.
  */
-export const handleVoidedTxList = async (mysql: ServerlessMysql, txs: Tx[]): Promise<[Tx[], DbTxOutput[]]> => {
-  // eslint-disable-next-line
-  console.log(`Setting ${txs.length} transactions as voided.`);
+export const handleVoidedTxList = async (mysql: ServerlessMysql, logger: Logger, txs: Tx[]): Promise<[Tx[], DbTxOutput[]]> => {
+  logger.debug(`Setting ${txs.length} transactions as voided.`, {
+    transactions: txs,
+  });
   await markTxsAsVoided(mysql, txs);
-  // eslint-disable-next-line
-  console.log(`Setting WalletTxHistory as voided from ${txs.length} transactions.`);
+  logger.debug(`Setting WalletTxHistory as voided from ${txs.length} transactions.`);
   await markWalletTxHistoryAsVoided(mysql, txs);
-  // eslint-disable-next-line
-  console.log(`Setting AddressTxHistory as voided from ${txs.length} transactions.`);
+  logger.debug(`Setting AddressTxHistory as voided from ${txs.length} transactions.`);
   await markAddressTxHistoryAsVoided(mysql, txs);
 
   // tx outputs are the list of all outputs in the transaction list
   const txOutputs: DbTxOutput[] = await getTxOutputs(mysql, txs);
+
+  logger.debug(`Fetched ${txOutputs.length} utxos from the voided transaction list`, {
+    txOutputs,
+  });
 
   // get outputs that were spent in txOutputs
   const txOutputsTxIds: Set<string> = txOutputs.reduce(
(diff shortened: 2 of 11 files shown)


serverless.yml
```diff
@@ -3,7 +3,12 @@ frameworkVersion: '2'
 
 useDotenv: true
 
-custom:
+custom: 
+  logLevelMap:  
+    mainnet: info
+    testnet: info
+    dev: info
+  logLevel: ${self:custom.logLevelMap.${opt:stage}, 'debug'}
```
andreabadesso, line 11: Any stage other than `mainnet`, `testnet` or `dev` will default to `debug`

serverless.yml
```diff
@@ -94,6 +99,7 @@ provider:
     DEFAULT_SERVER: ${env:DEFAULT_SERVER}
     WS_DOMAIN: ${env:WS_DOMAIN}
     TX_HISTORY_MAX_COUNT: ${env:TX_HISTORY_MAX_COUNT}
+    LOG_LEVEL: ${self:custom.logLevel}
```
andreabadesso, line 102: We can also change it manually on the serverless AWS page if we want to change it during execution

src/commons.ts
```diff
@@ -350,20 +351,23 @@ export const searchForLatestValidBlock = async (mysql: ServerlessMysql): Promise
...
-  // eslint-disable-next-line
-  console.log(`Setting ${txs.length} transactions as voided.`);
+export const handleVoidedTxList = async (mysql: ServerlessMysql, logger: Logger, txs: Tx[]): Promise<[Tx[], DbTxOutput[]]> => {
+  logger.debug(`Setting ${txs.length} transactions as voided.`, {
+    transactions: txs,
+  });
   await markTxsAsVoided(mysql, txs);
-  // eslint-disable-next-line
-  console.log(`Setting WalletTxHistory as voided from ${txs.length} transactions.`);
+  logger.debug(`Setting WalletTxHistory as voided from ${txs.length} transactions.`);
```
andreabadesso, line 359: We are already logging the full transactions array on the log above this one

src/txProcessor.ts
```diff
@@ -198,12 +222,17 @@ export const handleVoidedTx = async (tx: Transaction): Promise<void> => {
...
   const txId = tx.tx_id;
   const network = process.env.NETWORK;
 
+  // add the tx id to all logs from this method, so we can search by txId on CloudWatch
+  const logger = _logger;
+  logger.defaultMeta = {
+    ...logger.defaultMeta,
+    txId,
+  };
+
```
pedroferreira1, line 235: Should we log the full tx data here?

src/txProcessor.ts
```diff
@@ -234,6 +266,7 @@ const _unsafeAddNewTx = async (tx: Transaction, now: number, blockRewardLock: nu
     || tx.version === hathorLib.constants.MERGED_MINED_BLOCK_VERSION) {
     // unlock older blocks
     const utxos = await getUtxosLockedAtHeight(mysql, now, tx.height);
+    logger.debug(`Block transaction, unlocking ${utxos.length} locked utxos at height ${tx.height}`);
```
pedroferreira1: I'm think about logging all utxos being unlocked here. Not sure if it's worth it

src/txProcessor.ts
```diff
@@ -270,14 +303,22 @@ const _unsafeAddNewTx = async (tx: Transaction, now: number, blockRewardLock: nu
 
   // add transaction outputs to the tx_outputs table
   markLockedOutputs(outputs, now, heightlock !== null);
+  logger.debug(`Adding ${txId} to database`, tx);
```
pedroferreira1: This is the log I requested in the beginning of the method. I know here is just when it's really added but the tx data could be logged in the beginning and here we just log "Adding txID to database" What do you think?

pedroferreira1 (commented):
I just didn't understand why the log parameters were added as a first (or second) parameter in the methods. Why don't we just add as the last parameter in all of them?

src/txProcessor.ts
andreabadesso: On this line we are adding the `txId` to every log message that will be logged next, so we can filter by it

Added a new log to log the full tx on 9e75e9d8951937e8c5344bd7dd40a5d988885035

src/txProcessor.ts
andreabadesso: I agree, done in b57b87301e570e3597ddd9ace24b36a97818a26f

src/txProcessor.ts
andreabadesso: I agree, removed the fulltx log here and added it on top of the txProcessor method in b57b87301e570e3597ddd9ace24b36a97818a26f

src/txProcessor.ts
```diff
@@ -183,12 +203,16 @@ export const onSearchForLatestValidBlockRequest: APIGatewayProxyHandler = async
 export const handleVoidedTx = async (tx: Transaction): Promise<void> => {
   const txId = tx.tx_id;
   const transaction: Tx = await fetchTx(mysql, txId);
+  const logger = createDefaultLogger();
```
msbrogli, line 206: Can't we instantiate `logger` in a separate file and import as a global variable in each file? Here is an example: https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/src/logger.js

I feel it is better than passing it all along method calls.

serverless.yml
msbrogli: Do we have it explained in the SOP? Do we have to restart the already started lambdas execution environments?

tests/integration.test.ts
```diff
@@ -141,6 +158,9 @@ test('receive blocks and txs and then start wallet', async () => {
   await txProcessor.onNewTxEvent(txEvent);
   await checkAfterReceivingTx1(false);
 
+  // txEvent2 uses utxos that are not from the received blocks, so we must add them to the database
+  await addToUtxoTable(mysql, tx2Inputs);
```
msbrogli, line 162: What change affected these tests? I feel you basically improved logging and added some integrity checks. I couldn't find any fix or new feature.

msbrogli (changes requested):

pedroferreira1 (approved):

src/txProcessor.ts
andreabadesso: We could, but if we do that, we would either lose the `requestId` (that is added on the handlers) or risk having wrong `requestIds` on the requests (as multiple parallel calls would mutate the global logger object)

That's what I'm doing with this line:

```javascript
  logger.defaultMeta = {
    requestId: context.awsRequestId,
  };
```

tests/integration.test.ts
andreabadesso: The integrity checks broke those tests because `txEvent2` spends `utxos` that are not on our database, I just added them to the database

andreabadesso:
> Do we have it explained in the SOP? Do we have to restart the already started lambdas execution environments?

@msbrogli 

Added a SOP line explaining how to change the log level 262ad8f7111b1614476ff63ecee398e679d685ac

tests/integration.test.ts
msbrogli: Got it. Thanks!

serverless.yml
msbrogli: I just saw your message saying that you added it to the SOP. Thanks!

msbrogli (approved):

PR #226, opened 2022-05-18
Add weight to transaction table and update sync algorithm accordingly

lucas3003:
### Acceptance Criteria
- Add weight to `transactions` table
- Update sync algorithm to insert weight data on the new column

### Security Checklist

### DIFF
--- a/db/models/transaction.js
+++ b/db/models/transaction.js
@@ -41,6 +41,11 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    weight: {
+      type: DataTypes.INTEGER.UNSIGNED,
+      allowNull: true,
+      defaultValue: null,
+    },
   }, {
     sequelize,
     modelName: 'Transaction',
--- a/db/models/transaction.js
+++ b/db/models/transaction.js
@@ -41,8 +41,10 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    // We will support null only for the first data migration. The current entries do not have this value and manual insertion will be needed.
+    // Once we backfill the old data, we must stop accepting null for weight
     weight: {
-      type: DataTypes.INTEGER.UNSIGNED,
+      type: DataTypes.FLOAT.UNSIGNED,
       allowNull: true,
       defaultValue: null,
     },
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -48,7 +48,7 @@ import {
 } from '@src/utils';
 import {
   getWalletFromDbEntry,
-  getTxsFromDBResult
+  getTxsFromDBResult,
 } from '@src/db/utils';
 
 const BLOCK_VERSION = [
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -614,7 +614,7 @@ export const updateTx = async (
   height: number,
   timestamp: number,
   version: number,
-  weight: number
+  weight: number,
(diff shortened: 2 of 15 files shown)


db/models/transaction.js
```diff
@@ -41,6 +41,11 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    weight: {
+      type: DataTypes.INTEGER.UNSIGNED,
+      allowNull: true,
```
pedroferreira1, line 48: I think we can comment here that we support `null` only for data migration, because the data already there won't have the `weight` and we will need to manually input it there.

It's important to describe this here in case someone wants to do anything with `weight`, it's safe to assume that will be there the value.

src/db/index.ts
```diff
@@ -614,7 +614,8 @@ export const updateTx = async (
   height: number,
   timestamp: number,
   version: number,
-): Promise<void> => addOrUpdateTx(mysql, txId, height, timestamp, version);
+  weight: number,
```
pedroferreira1, line 620: Missing new parameter in the docstring

src/db/index.ts
```diff
@@ -633,14 +634,15 @@ export const addOrUpdateTx = async (
   height: number,
   timestamp: number,
   version: number,
+  weight: number,
```
pedroferreira1, line 641: Missing new parameter in the docstring

db/models/transaction.js
```diff
@@ -41,6 +41,11 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    weight: {
+      type: DataTypes.INTEGER.UNSIGNED,
```
pedroferreira1: `weight` is not an integer, it's a float in our full node.

src/db/index.ts
```diff
@@ -633,14 +634,15 @@ export const addOrUpdateTx = async (
...
-  const entries = [[txId, height, timestamp, version]];
+  const entries = [[txId, height, timestamp, version, weight]];
 
   await mysql.query(
-    `INSERT INTO \`transaction\` (tx_id, height, timestamp, version)
+    `INSERT INTO \`transaction\` (tx_id, height, timestamp, version, weight)
      VALUES ?
-         ON DUPLICATE KEY UPDATE height = ?`,
-    [entries, height],
+         ON DUPLICATE KEY UPDATE height = ?, weight = ?`,
```
pedroferreira1: Why do we need the `weight` in the `ON DUPLICATE KEY UPDATE`? I thought this was useful only for transactions that could get their height after a block confirmation.

src/db/index.ts
```diff
@@ -1748,6 +1750,7 @@ export const getTxsAfterHeight = async (
       version: result.version as number,
       voided: result.voided as boolean,
       height: result.height as number,
+      weight: result.weight as number,
```
pedroferreira1: I feel we could create a method to get `Tx` object from a db result. With that, if we decide to add a new column in this table in the future we won't need to add this new column in all places like you did with `weight`. What do you think?

pedroferreira1 (changes requested):

db/models/transaction.js
lucas3003: Done.

src/db/index.ts
lucas3003: Added.

src/db/index.ts
lucas3003: Added

andreabadesso (approved):

pedroferreira1 (approved):

PR #231, opened 2022-06-01
feat: add timestamp to transaction table

lucas3003:
### Acceptance Criteria
- Include timestamp fields on `transaction` table to be used by Logstash.

This is an example of the final table structure (Weight data is mocked):

image

### Security Checklist

### DIFF
--- a/db/migrations/20220531191235-add-timestamp-fields-tx.js
+++ b/db/migrations/20220531191235-add-timestamp-fields-tx.js
@@ -2,21 +2,37 @@
 
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    await queryInterface.addColumn('transaction', 'created_at', {
-      type: 'TIMESTAMP',
-      allowNull: false,
-      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
-    });
+    const transaction = await queryInterface.sequelize.transaction();
+    try {
+      await queryInterface.addColumn('transaction', 'created_at', {
+        type: 'TIMESTAMP',
+        allowNull: false,
+        defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
+      });
 
-    await queryInterface.addColumn('transaction', 'updated_at', {
-      type: 'TIMESTAMP',
-      allowNull: false,
-      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
-    });
+      await queryInterface.addColumn('transaction', 'updated_at', {
+        type: 'TIMESTAMP',
+        allowNull: false,
+        defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
+      });
+
+      await transaction.commit();
+    } catch (err) {
+      await transaction.rollback();
+      throw err;
+    }
   },
 
   down: async (queryInterface) => {
-    await queryInterface.removeColumn('transaction', 'created_at');
-    await queryInterface.removeColumn('transaction', 'updated_at');
+    const transaction = await queryInterface.sequelize.transaction();
+    try {
+      await queryInterface.removeColumn('transaction', 'created_at');
+      await queryInterface.removeColumn('transaction', 'updated_at');
+
+      await transaction.commit();
+    } catch (err) {
+      await transaction.rollback();
+      throw err;
+    }
   }
 };
--- a/db/migrations/20220531191235-add-timestamp-fields-tx.js
+++ b/db/migrations/20220531191235-add-timestamp-fields-tx.js
@@ -8,13 +8,23 @@ module.exports = {
         type: 'TIMESTAMP',
         allowNull: false,
         defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
-      });
+      }, { transaction });
 
       await queryInterface.addColumn('transaction', 'updated_at', {
         type: 'TIMESTAMP',
         allowNull: false,
         defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
-      });
+      }, { transaction });
+
+      await queryInterface.addIndex(
+        'transaction',
+        ['updated_at'],
+        {
+          name: 'transaction_updated_at_idx',
+          fields: 'updated_at',
+          transaction,
+        }
+      );
 
       await transaction.commit();
     } catch (err) {
@@ -26,8 +36,9 @@ module.exports = {
   down: async (queryInterface) => {
     const transaction = await queryInterface.sequelize.transaction();
     try {
-      await queryInterface.removeColumn('transaction', 'created_at');
-      await queryInterface.removeColumn('transaction', 'updated_at');
+      await queryInterface.removeColumn('transaction', 'created_at', { transaction });
+      await queryInterface.removeColumn('transaction', 'updated_at', { transaction });
+      await queryInterface.removeIndex('transaction', 'transaction_updated_at_idx', { transaction });
 
       await transaction.commit();
     } catch (err) {
--- a/db/migrations/20220531191235-add-timestamp-fields-tx.js
+++ b/db/migrations/20220531191235-add-timestamp-fields-tx.js
@@ -1,0 +1,22 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('transaction', 'created_at', {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
+    });
+
+    await queryInterface.addColumn('transaction', 'updated_at', {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
+    });
+  },
+
+  down: async (queryInterface) => {
+    await queryInterface.removeColumn('transaction', 'created_at');
+    await queryInterface.removeColumn('transaction', 'updated_at');
+  }
+};
--- a/db/models/transaction.js
+++ b/db/models/transaction.js
@@ -71,6 +71,9 @@ module.exports = (sequelize, DataTypes) => {
     }, {
       name: 'transaction_height_idx',
       fields: ['height'],
+    }, {
+      name: 'transaction_updated_at_idx',
+      fields: ['updated_at'],
     }],
   });
   return Transaction;
--- a/db/models/transaction.js
+++ b/db/models/transaction.js
@@ -48,11 +48,22 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    createdAt: {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
+    },
+    updatedAt: {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
+    },
   }, {
     sequelize,
     modelName: 'Transaction',
     tableName: 'transaction',
     timestamps: false,
+    underscored: true,
     indexes: [{
       name: 'transaction_version_idx',
       fields: ['version'],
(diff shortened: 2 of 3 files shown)


db/models/transaction.js
```diff
@@ -48,11 +48,22 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    createdAt: {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
```
andreabadesso: I think this might be broken on other model files as well, but we don't have `Sequelize` imported here

Not sure if we can use sequelize or `DataTypes`

andreabadesso (changes requested):

db/models/transaction.js
lucas3003: Changed to `sequelize`as suggested [here](https://stackoverflow.com/a/56302187)

db/migrations/20220531191235-add-timestamp-fields-tx.js
```diff
@@ -0,0 +1,22 @@
...
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
+    });
+
+    await queryInterface.addColumn('transaction', 'updated_at', {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
+    });
```
luislhl: Should we use a transaction that contains the 2 operations, to make sure they both either run totally or not at all?

Like in https://sequelize.org/docs/v6/other-topics/migrations/#migration-skeleton

db/migrations/20220531191235-add-timestamp-fields-tx.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+
+  down: async (queryInterface) => {
+    await queryInterface.removeColumn('transaction', 'created_at');
+    await queryInterface.removeColumn('transaction', 'updated_at');
```
luislhl: Same question about using a transaction

db/models/transaction.js
```diff
@@ -48,11 +48,22 @@ module.exports = (sequelize, DataTypes) => {
...
+    updatedAt: {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
+    },
   }, {
     sequelize,
     modelName: 'Transaction',
     tableName: 'transaction',
     timestamps: false,
```
luislhl, line 65: Shouldn't we just change this to `true` instead of creating the columns manually?

https://sequelize.org/docs/v6/core-concepts/model-basics/#timestamps

db/models/transaction.js
```diff
@@ -48,11 +48,22 @@ module.exports = (sequelize, DataTypes) => {
...
+    underscored: true,
```
luislhl, line 66: What's this for?

db/models/transaction.js
```diff
@@ -48,11 +48,22 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    createdAt: {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: sequelize.literal('CURRENT_TIMESTAMP'),
+    },
+    updatedAt: {
+      type: 'TIMESTAMP',
```
luislhl, line 57: What's the difference of using a string directly for the type, and using `DataTypes`?

db/models/transaction.js
```diff
@@ -48,11 +48,22 @@ module.exports = (sequelize, DataTypes) => {
       allowNull: true,
       defaultValue: null,
     },
+    createdAt: {
+      type: 'TIMESTAMP',
+      allowNull: false,
+      defaultValue: sequelize.literal('CURRENT_TIMESTAMP'),
+    },
+    updatedAt: {
```
luislhl, line 56: Should we have an index in this field, since the query that will be performed by logstash filters based on it?

Or is the query performance acceptable even without the index?

I think this table is big enough to possibly take a longer time for the query to complete, right?

db/models/transaction.js
lucas3003: Sequelize provides DateTypes.DATE (reference [here](https://sequelize.org/docs/v6/core-concepts/model-basics/#data-types)), but JDBC Logstash input plugin requires `TIMESTAMP` (reference [here](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html#plugins-inputs-jdbc-tracking_column_type)). We can only achieve that by passing the string `TIMESTAMP`

db/models/transaction.js
lucas3003: I think we can measure the response and check if we will need the index. There is no big impact for us if the query takes, for example, 10 seconds longer (Only if this exhausts the RDS). I think we can create a separate KTLO for that. What do you think?

db/models/transaction.js
lucas3003: Transfrom `createdAt` and `updatedAt`on JSON to `created_at` and `updated_at` on database columns

db/models/transaction.js
lucas3003: This create `DATETIME` fields, while Logstash needs `TIMESTAMP`(reference [here](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html#plugins-inputs-jdbc-tracking_column_type))

db/migrations/20220531191235-add-timestamp-fields-tx.js
lucas3003: Fixed

db/migrations/20220531191235-add-timestamp-fields-tx.js
lucas3003: Fixed

db/models/transaction.js
pedroferreira1: How is that query? You store in logastash the last timestamp and then you query for the updateAt > last_timestamp? What's the interval between calls?

I'm thinking about the downside of having the index, and I see none, except for the space usage, which is pretty okay thinking about a better performance. Is there any problem if we add this index?

pedroferreira1 (approved):

db/migrations/20220531191235-add-timestamp-fields-tx.js
```diff
@@ -0,0 +1,38 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const transaction = await queryInterface.sequelize.transaction();
+    try {
+      await queryInterface.addColumn('transaction', 'created_at', {
+        type: 'TIMESTAMP',
+        allowNull: false,
+        defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
+      });
```
luislhl: I think you need to pass a fourth parameter to `addColumn` with the transaction.

```
      await queryInterface.addColumn('transaction', 'created_at', {
        type: 'TIMESTAMP',
        allowNull: false,
        defaultValue: Sequelize.literal('CURRENT_TIMESTAMP'),
      }, { transaction });
```

db/models/transaction.js
luislhl: I guess I agree with Pedro that the index wouldn't have any major downside.

Besides space usage, probably a slight increase in the time needed for ADD and UPDATE operations, but that shouldn't do any harm.

I think we should at least test the performance with and without the index to have a better picture. Testing this shouldn't be hard. I can give you permission on the dev table maybe, and you can use it to test.

What do you think?

lucas3003: > How is that query? You store in logastash the last timestamp and then you query for the updateAt > last_timestamp? What's the interval between calls?

You are right. We store the last evaluated timestamp and then query for updateAt > last_timestamp. The first query will be the bigger. We will run this query every minute, but this is totally configurable.

lucas3003: I will make the test to compare the results with and without the index.

lucas3003: I made some tests and I did not detect a performance improvement with index. I got 64s as best result without index and 65s with index. The queries are ranging from 60s-120s of elapsed time. With this scenario, I think it is not worth adding the index. What do you think?

This was the query I used:

```
SELECT t.*,
         UNIX_TIMESTAMP(t.updated_at) AS unix_ts_in_secs
          FROM `transaction` t
         WHERE (UNIX_TIMESTAMP(t.updated_at) > 0
           AND t.updated_at < NOW())
```

Also, I will update the Logstash pipeline. I am making this query every minute. I will update it to wait at least 3 minutes for the next query. What do you think?

pedroferreira1: How was the index added? Depending on the index type, won't have an improvement in a query that searches in a range.

lucas3003: I created using ```CREATE INDEX updated_at_idx ON `transaction` (updated_at)```

luislhl: I think the query is not using the index for some reason.

If you run an `explain` you can see the query plan:

```
explain SELECT t.*, UNIX_TIMESTAMP(t.updated_at) AS unix_ts_in_secs FROM `transaction` t WHERE (UNIX_TIMESTAMP(t.updated_at) > 0 AND t.updated_at < NOW());

+----+-------------+-------+------------+------+----------------+------+---------+------+---------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys  | key  | key_len | ref  | rows    | filtered | Extra       |
+----+-------------+-------+------------+------+----------------+------+---------+------+---------+----------+-------------+
|  1 | SIMPLE      | t     | NULL       | ALL  | updated_at_idx | NULL | NULL    | NULL | 2005574 |    50.00 | Using where |
+----+-------------+-------+------------+------+----------------+------+---------+------+---------+----------+-------------+
```

It knows there is a possible key to use, but it chooses not to use it. We can see that because of the `NULL` in the key column.

Maybe we are hitting a case similar to https://stackoverflow.com/questions/18297509/mysql-index-on-timestamp-column-not-used-for-large-date-ranges

Because you are using a large date range in your test (between 0 and NOW), the engine chooses not to use the index becuse it thinks it would be slower to do so.

However, I tried a simpler query with a smaller range, but it still doesn't use the index:
```
MySQL [hathor_wallet_service_dev]> explain SELECT t.* FROM `transaction` t WHERE t.updated_at > NOW() - INTERVAL 1 DAY;
+----+-------------+-------+------------+------+----------------+------+---------+------+---------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys  | key  | key_len | ref  | rows    | filtered | Extra       |
+----+-------------+-------+------------+------+----------------+------+---------+------+---------+----------+-------------+
|  1 | SIMPLE      | t     | NULL       | ALL  | updated_at_idx | NULL | NULL    | NULL | 2005634 |    50.00 | Using where |
+----+-------------+-------+------------+------+----------------+------+---------+------+---------+----------+-------------+
```

Then I went to look how many records are in this date range, and it seems all of them are:
```
MySQL [hathor_wallet_service_dev]> SELECT count(*) FROM `transaction` t WHERE t.updated_at > NOW() - INTERVAL 1 DAY;
+----------+
| count(*) |
+----------+
|  1840880 |
```

Does it make sense that all records have the `updated_at` field greater than yesterday? I didn't follow what happened in this database to make all of them be updated.

But I guess in real world situations, where only a small set of the records would be get by our query, the engine would choose to use the index. We would need to maybe manipulate the data so we can test this behavior.

Does it make sense?

I also recommend keeping the query simple to test the index, like the one I did above. Things like transformations in the left side of the comparison can sometimes interfere with the index as well (like `UNIX_TIMESTAMP(t.updated_at)`)

After we confirm the index works for simpler queries, we could test the more complex one as well.

luislhl: But anyway, if with smaller ranges and less records in the range we still can't make the index work, or if this is not the expected scenario in the real world, I agree on increasing the job periodicity.

lucas3003: I updated the timestamp for 200 entries, and ran the EXPLAIN query. This was the result:

|id|select_type|table|partitions|type|possible_keys|key|key_len|ref|rows|filtered|Extra|
|--|-----------|-----|----------|----|-------------|---|-------|---|----|--------|-----|
|1|SIMPLE|t||range|updated_at_idx|updated_at_idx|4||204|100.0|Using index condition|

From this, can we confirm it is using the index for smaller queries? If so, I think it is worth using the index. The SELECT query is taking 478ms on smaller queries (200 entries).

lucas3003: Added the index in the new commit.

db/migrations/20220531191235-add-timestamp-fields-tx.js
lucas3003: Fixed in the new commit

luislhl (approved):

PR #235, opened 2022-06-18
feat: auto review nfts

tuliomir:
> :construction: This issue is still a work in progress

## Summary
This issue aims to design a new backend that will manage push notifications for all our applications, but mainly for the user wallets. This should be done in the simplest and modular way to allow for easy maintenance and expansion.

In its first iteration it will serve only the Wallet Mobile.

Original [requirements document](https://docs.google.com/document/d/1i3V1pnhQgL5g1yrDg8C4mFUPMh9Cg9hEipI_AjQmXL0/edit#heading=h.bylg7xwny4xf).

## Motivation
The Mobile Wallet is now being released with real-time updates powered by the wallet service. Push notifications are a great complement to this experience, allowing the user to also be informed about changes on his wallet while away from the device.

This feature demands a backend service to manage the notifications, receiving requests from Hathor's internal services and passing them to the user device's infrastructre - in our case, [Google Firebase Cloud Messaging](https://firebase.google.com/docs/cloud-messaging) (FCM) and [Apple Push Notification Service](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns/) (APNS).

More on motivation for the front-end is available on the document.

## Guide-level description
The push notification system must know two key data points about a user to properly work:
- What are the addresses of the user's wallet ( or its `xpub` to derive them locally )
- What is the device of the user

Once this information is stored, one of our services should be able to "listen" to the network for interactions with these addresses to trigger relevant push notifications.

To achieve this use case, we can use the [**wallet service**](https://github.com/HathorNetwork/hathor-wallet-service) for:
- Allowing a `wallet-mobile` to register a wallet using the service's currently available endpoints
- Link a device to a previously registered wallet using a new endpoint
- Reuse the existing `onNewTxRequest` to monitor transactions for this wallet

> :construction: Currently being investigated: what is the best path to build this service?

Once a wallet-related transaction is identified, it is necessary to call the device's push provider. This could be done on the `wallet service` itself, or a new service could be built for this purpose:
|  | Reuse the wallet service | Creating a new service |
|--|--|--|
| Pros | Easy to implement | Modularization, Follows the single-responsibility principle |
| Cons | Feature bloat | Big infrastructure overhead ( deploy, monitoring, ... ) |

## Reference-level description

- Typescript/Express server
  - Most tutorials present this solution
  - May have scalability or availability issues
  - Inputs made via queue ( Amazon SQS? )
- Serverless lambda endpoint
  - Easily scalable and high availability
  - APNS supports a non-optimized [token-authenticated HTTP interface](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server) that allows this.
- Amazon SNS
  - Easy-to-implement solution
  - Third party dependency and vendor lock-in
  - Potential privacy issue by sharing sensitive information via plain text

### Concerns to solve
**Privacy**
- What will be the content of the notification?
  - If we add the transaction id, one could infer the user's addresses from the tx outputs

**Resource Management**
- Keeping track of user wallets 
  - How will our services know which device to send a push notification to when a transaction arrives?
  - Will the device subscribe to a list of addresses within our backend?
  - How many addresses will we allow each user/wallet to subscribe to?
- Load handling
  - A large amount of transaction outputs may increase the load on our services
  - Important to use queues on handling these loads, to avoid loss of event notifications on high loads
  - Delays are more acceptable


src/txProcessor.ts
```diff
@@ -79,6 +84,7 @@ export const IGNORE_TXS = {
  * calls the appropriate function to handle the transaction.
  *
  * @param event - The SQS event
+ * @deprecated
```
tuliomir, line 83: This is not related to the scope of this PR, but was relevant to the design process.

events/nftCreationTx.ts
```diff
@@ -0,0 +1,102 @@
...
+  token_name: 'New NFT',
+  token_symbol: 'NNFT',
+  tokens: [],
+  // Properties exclusive to the wallet-service, in comparison with the lib's sample tx
+  is_block: false,
+  type: 'network:new_tx_accepted',
+  throttled: false,
+};
+
+const cloneObj = (obj) => JSON.parse(JSON.stringify(obj));
```
andreabadesso: We have lodash on the project, you can use https://lodash.com/docs/#clone

tuliomir: Done on [a47b736](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/a47b736ba32a1183d05ac9ca9c6a3408b69504f3).

src/utils.ts
```diff
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { createHash, HexBase64Latin1Encoding } from 'crypto';
+import { BinaryToTextEncoding, createHash } from 'crypto';
```
tuliomir, line 8: Related to dependencies update:
`HexBase64Latin1Encoding` was removed on the latest version of `@types/node`.
Changed it to match the current `hash.digest()` contract.

src/txProcessor.ts
r4mmer: But is it still deployed? Or is this lambda removed?

src/txProcessor.ts
```diff
@@ -144,6 +141,23 @@ export const onNewTxRequest: APIGatewayProxyHandler = async (event, context) =>
...
     };
   }
+
+  // Validating for NFTs only after the tx is successfully added
+  try {
+    if (NftUtils.isTransactionNFTCreation(tx)) {
+      // This process is not critical, so we run it in a fire-and-forget manner.
+      // In case of errors, just log the asynchronous exception and take no action on it.
+      NftUtils.invokeNftHandlerLambda(tx.tx_id)
+        .catch((err) => logger.error('Errored on nftHandlerLambda invocation', err));
```
r4mmer: Why not let the outer catch handle this error?

src/txProcessor.ts
```diff
@@ -215,6 +229,53 @@ export const handleVoidedTx = async (tx: Transaction): Promise<void> => {
...
 };
 
+interface NewNftEvent {
+  nftUid: string
+}
+
+interface NftValidationResults {
+  success: boolean;
+  message?: string;
+}
```
r4mmer: These should be on `src/types.ts`.
Also, since this is a very specific handler I think we can make a handler type like in this [example](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/handler.d.ts#L12)

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+   * @param {string} nftUid
+   * @param {Record<string, unknown>} metadata
+   */
+  static async _updateMetadata(nftUid: string, metadata: Record<string, unknown>): Promise<unknown> {
+  // invoke lambda asynchronously to metadata update
+    const lambda = new Lambda({
+      apiVersion: '2015-03-31',
+      endpoint: process.env.STAGE === 'dev'
+        ? 'http://localhost:3001'
+        : 'https://lambda.eu-central-1.amazonaws.com',
```
r4mmer: Isn't there a dev stage on aws? Should we use `process.env.ENVIRONMENT` here?

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+    });
+
+    const params = {
+    // FunctionName is composed of: service name - stage - function name
+      FunctionName: `hathor-explorer-service-${NftUtils.getExplorerServiceStage(process.env.STAGE)}-create_or_update_dag_metadata`,
```
r4mmer: Instead of this I think we should have the whole function name as a configuration. This would also eliminate the need for a `getExplorerServiceStage` method.

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+      // End the loop successfully
+        return response;
+      }
+
+      logger.warn('Failed metadata update', {
+        nftUid,
+        statusCode: response.StatusCode,
+        message: response.Payload.toString(),
+      });
+      ++retryCount;
```
r4mmer, line 105: We should log each failed attempt, this would make debugging easier if any errors are happening on this invocation.

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    // Exceeded retry limit
+    throw new Error('Metadata update failed.');
```
r4mmer: Should we log something here to raise an alert?
If we intend to raise an alert from this, we should add more metadata to make the investigation simpler.
I think just the `tx_id` is enough actually.

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+    const lambda = new Lambda({
+      apiVersion: '2015-03-31',
+      endpoint: process.env.STAGE === 'dev'
+        ? 'http://localhost:3002'
+        : `https://lambda.${process.env.AWS_REGION}.amazonaws.com`,
+    });
+
+    const params = {
+    // FunctionName is composed of: service name - stage - function name
+      FunctionName: `${process.env.SERVICE_NAME}-${process.env.STAGE}-onNewNftEvent`,
```
r4mmer: Same here.

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+   * Invokes this application's own intermediary lambda `onNewNftEvent`.
+   * This is to improve the failure tolerance on this non-critical step of the sync loop.
+   */
+  static async invokeNftHandlerLambda(txId: string): Promise<void> {
+  // invoke lambda asynchronously to handle NFT metadata addition
+    const lambda = new Lambda({
+      apiVersion: '2015-03-31',
+      endpoint: process.env.STAGE === 'dev'
+        ? 'http://localhost:3002'
+        : `https://lambda.${process.env.AWS_REGION}.amazonaws.com`,
```
r4mmer: Same here

r4mmer (changes requested):

src/txProcessor.ts
tuliomir: This is still deployed, I only added this line to help developers and static code analyzers identify that this should not be used anymore.
The lambda itself will be removed in another PR.

src/txProcessor.ts
tuliomir: Because we want the `onNewTxRequest` lambda to return with success as soon as possible. Note that `invokeNftHandlerLambda()` returns a function and we are not using `await` to wait for its response here.

I refactored this code on [f88dc663](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/f88dc6630a292a953aaffb5a065c8b65c3becf10) to make it simpler on `txProcessor` and clearer on `NftUtils`.
Do you have suggestions on it?

src/txProcessor.ts
tuliomir: Moved these types to inline on [a35b1310](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/a35b13103bfea43d7c41196b358d21e4519f5c8b).

src/utils/nft.utils.ts
tuliomir: Added the `retryCount` to the logger warning on [6380e15f](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/6380e15fda86d877153bd92744e97e2872875fff). Do you think another information would be useful to have here?

src/utils/nft.utils.ts
tuliomir: Good point. I assumed the logs on _retry_ instances would be enough, but it's always good to have this information at hand on every log.
Added the `tx_id` to the error message on [6380e15f](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/6380e15fda86d877153bd92744e97e2872875fff).

serverless.yml
```diff
@@ -30,13 +30,23 @@ custom:
   apiGatewayThrottling:
     maxRequestsPerSecond: 500
     maxConcurrentRequests: 250
+  stage: ${opt:stage, 'dev'}
+  explorerServiceStage: dev # The wallet-service `dev-testnet` environments maps to the explorer-service's `dev`
+  serverlessIfElse:
```
andreabadesso, line 35: I think you could use [`Fn::Split`](https://docs.aws.amazon.com/pt_br/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-split.html) to avoid adding a new library, I'm not sure about the correct syntax, but my guess is that it would look something like:

```
Fn::Select:
  - 0
  - Fn::Split
    - '-'
    - ${opt:stage, 'dev'}
```

Wouldn't waste too much time on this though, as the library is just a dev dependency

serverless.yml
```diff
@@ -131,6 +141,14 @@ functions:
...
       walletWarmer:
         enabled: false
+    iamRoleStatementsInherit: true
+    iamRoleStatements:
+      - Effect: Allow
+        Action:
+          - lambda:InvokeFunction
+          - lambda:InvokeAsync
+        Resource:
+          arn:aws:lambda:eu-central-1:769498303037:function:hathor-wallet-service-${self:custom.stage}-onNewNftEvent
```
andreabadesso: Can't we use something like what was described [here](https://stackoverflow.com/a/61992891)?

serverless.yml
```diff
@@ -153,13 +171,26 @@ functions:
...
+        Resource:
+          arn:aws:lambda:eu-central-1:769498303037:function:hathor-explorer-service-${self:custom.explorerServiceStage}-create_or_update_dag_metadata
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
     warmup:
       walletWarmer:
         enabled: false
   loadWalletApi:
-    role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
+    role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda # TODO: Use local iamRoleStatements instead of this rule
```
andreabadesso: TODO

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * A helper for generating and updating a NFT Token's metadata.
+ */
+
+export class NftUtils {
+  /**
+   * The wallet-service `dev-testnet` environments maps to the explorer-service's `dev`
+   */
+  static getExplorerServiceStage(walletStage: string): string {
+    if (walletStage === 'dev-testnet') {
```
andreabadesso: Can't we add an environment variable in `serverless.yml`, using `custom.explorerServiceStage` and use it here?

src/utils/nft.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+      tx.version !== hathorLib.constants.CREATE_TOKEN_TX_VERSION // Must be a token creation tx
+    || !tx.token_name // Must have a token name
+    || !tx.token_symbol // Must have a token symbol
+    ) {
+      return false;
+    }
+
+    // Continue with a deeper validation
+    let isNftCreationTx;
+    const libTx: hathorLib.CreateTokenTransaction = hathorLib.helpersUtils.createTxFromHistoryObject(tx);
```
andreabadesso: Shouldn't we move this inside the try/catch?

src/utils/nft.utils.ts
andreabadesso: Another option would be to set the `explorerServiceStage` as an env variable, using `custom.explorerServiceStage` on `serverless.yml`

src/utils/nft.utils.ts
andreabadesso: If we want to be notified about this failure, add `[ALERT]` to the log string so alert-manager gets it

src/utils/nft.utils.ts
andreabadesso: @tuliomir Also add `[ALERT]` to the error string, so alert-manager will get it

andreabadesso (changes requested):

src/utils/nft.utils.ts
tuliomir: On [56f859db](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/56f859db652843dc2974e61e5f9bfa54b7873a24) I added two new environment variables, one for the `wallet-service` endpoint itself and another for the `explorer-service` endpoint.

The function name itself I kept as a code string, because I think it shouldn't be so easy to change it.
Can you both check if this is a valid suggestion?

src/utils/nft.utils.ts
tuliomir: I refactored this method on [f88dc663](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/f88dc6630a292a953aaffb5a065c8b65c3becf10) to catch errors on this function, and also to make the `txProcessor` code simpler.

serverless.yml
tuliomir: Removed the comment and created issue #257 to track this `todo`

src/utils/nft.utils.ts
tuliomir: Good idea! Done on [56f859db](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/56f859db652843dc2974e61e5f9bfa54b7873a24).

serverless.yml
tuliomir: Improved on [aefd31f0](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/aefd31f0d1755845de720487b71d59263b298593).

src/utils/nft.utils.ts
tuliomir: I made some changes on [56f859db](https://github.com/HathorNetwork/hathor-wallet-service/pull/235/commits/56f859db652843dc2974e61e5f9bfa54b7873a24) that involve this code being moved to environment variables.

r4mmer (approved):

andreabadesso (approved):

serverless.yml
tuliomir: It is a good idea, but it's not compatible with the structure we've made at the moment: the `explorerServiceStage` that we need to calculate is used on the environment variables and, more importantly, on a text interpolation on `functions.onNewNftEvent.iamRoleStatements.Resource`.

Since the evaluation of the `Fn::Select` and `Fn::Split` only occur at the CloudFormation stage, it seems to not work very well with the `serverless-offline`, nor it passes the validations on `npx serverless print`.

So we may do this on another PR/issue later, but will keep with what's been already tested and worked well so far for this PR.

PR #236, opened 2022-07-07
fix: rebuild address balances

andreabadesso:
## Motivation

Currently, when handling balance recalculations on a re-org, the query is summing values from not only locked utxos, but also from unlocked utxos on the `locked_balance` column.

This PR solves https://github.com/HathorNetwork/hathor-wallet-service/pull/236
(Which also includes a more in-depth explanation of the problem)

### Acceptance Criteria
- We should properly ignore unlocked utxos when recalculating `unlocked_balance` for an address token_id
- We should improve the test case to consider a more real scenario

### Security Checklist

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -2041,6 +2041,7 @@ export const rebuildAddressBalancesFromUtxos = async (
         AND timelock IS NULL
         AND spent_by IS NULL
         AND voided = FALSE
+        AND locked = FALSE
         AND address IN (?)
    GROUP BY address, token_id
   `, [addresses]);
@@ -2068,6 +2069,7 @@ export const rebuildAddressBalancesFromUtxos = async (
            OR \`timelock\` IS NOT NULL)
           AND spent_by IS NULL
           AND voided = FALSE
+          AND locked = TRUE
           AND address IN (?)
      GROUP BY \`address\`, \`token_id\`
    ON DUPLICATE KEY UPDATE
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -2037,9 +2037,7 @@ export const rebuildAddressBalancesFromUtxos = async (
             0, -- timelock_expires
             COUNT(DISTINCT \`tx_id\`) -- transactions
        FROM \`tx_output\`
-      WHERE heightlock IS NULL
-        AND timelock IS NULL
-        AND spent_by IS NULL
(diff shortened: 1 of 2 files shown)


src/db/index.ts
```diff
@@ -2037,10 +2037,9 @@ export const rebuildAddressBalancesFromUtxos = async (
             0, -- timelock_expires
             COUNT(DISTINCT \`tx_id\`) -- transactions
        FROM \`tx_output\`
-      WHERE heightlock IS NULL
-        AND timelock IS NULL
-        AND spent_by IS NULL
+      WHERE spent_by IS NULL
         AND voided = FALSE
+        AND locked = FALSE
```
andreabadesso, line 2042: `locked = FALSE` is enough to capture only unlocked utxos, no need to verify `heightlock` and `timelock`

src/db/index.ts
```diff
@@ -2064,10 +2063,9 @@ export const rebuildAddressBalancesFromUtxos = async (
               MIN(\`timelock\`) AS timelock_expires,
               COUNT(DISTINCT \`tx_id\`) -- transactions
          FROM \`tx_output\`
-        WHERE (\`heightlock\` IS NOT NULL
-           OR \`timelock\` IS NOT NULL)
-          AND spent_by IS NULL
+        WHERE spent_by IS NULL
           AND voided = FALSE
+          AND locked = TRUE
```
andreabadesso, line 2068: Same here

tests/db.test.ts
```diff
@@ -1542,6 +1542,8 @@ test('rebuildAddressBalancesFromUtxos', async () => {
     { value: 5, address: addr1, token: 'token1', locked: false, spentBy: null },
     { value: 15, address: addr1, token: 'token1', locked: false, spentBy: null },
     { value: 25, address: addr2, token: 'token2', timelock: 500, locked: true, spentBy: null },
+    { value: 75, address: addr2, token: 'token1', heightlock: 70, locked: true, spentBy: null },
+    { value: 150, address: addr2, token: 'token1', heightlock: 70, locked: true, spentBy: null },
```
andreabadesso, line 1546: These two would fail without the fix, the `lockedBalance` would be `260` on the check below

lucas3003 (approved):

luislhl (approved):

PR #243, opened 2022-07-13
feat: using serverless-plugin-warmup to keep some lambdas warm

andreabadesso:
## Motivation
The idea is to keep lambdas that are called from the wallets on load warm to prevent cold starts that take up to 10 seconds per request, causing the wallet load to take a long time.

### Acceptance Criteria
- We should warm only the lambdas that are necessary for the wallet load mechanism:
  - `loadWalletAsync`
  - `loadWalletApi`
  - `getWalletStatusApi`
  - `getAddressesApi`
  - `getNewAddresses`
  - `getBalanceApi`
  - `getTokensApi`
  - `getVersionData`
  - `getTxHistoryApi`

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -8,7 +8,6 @@ custom:
     mainnet: info
     testnet: info
     dev-testnet: info
-  logLevel: ${self:custom.logLevelMap.${opt:stage}, 'debug'}
   webpack:
     webpackConfig: ./webpack.config.js
     includeModules: true
@@ -129,8 +128,8 @@ functions:
   onNewTxEvent:
     handler: src/txProcessor.onNewTxEvent
   loadWalletAsync:
-    warmup: true
     handler: src/api/wallet.loadWallet
+    warmup: true
   loadWalletApi:
     role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
     handler: src/api/wallet.load
--- a/serverless.yml
+++ b/serverless.yml
@@ -32,6 +32,8 @@ plugins:
   - serverless-webpack
   - serverless-prune-plugin
   - serverless-api-gateway-throttling
+  - serverless-plugin-warmup
+
 
 resources:
   Resources:
@@ -127,10 +129,12 @@ functions:
   onNewTxEvent:
     handler: src/txProcessor.onNewTxEvent
   loadWalletAsync:
+    warmup: true
     handler: src/api/wallet.loadWallet
   loadWalletApi:
     role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
     handler: src/api/wallet.load
+    warmup: true
     events:
       - http:
           path: wallet/init
@@ -145,6 +149,7 @@ functions:
           cors: true
   getWalletStatusApi:
     handler: src/api/wallet.get
+    warmup: true
     events:
       - http:
           path: wallet/status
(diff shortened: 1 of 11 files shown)


serverless.yml
```diff
@@ -127,10 +129,12 @@ functions:
   onNewTxEvent:
     handler: src/txProcessor.onNewTxEvent
   loadWalletAsync:
+    warmup: true
     handler: src/api/wallet.loadWallet
   loadWalletApi:
     role: arn:aws:iam::769498303037:role/WalletServiceLoadWalletLambda
     handler: src/api/wallet.load
+    warmup: true
```
luislhl, line 141: Is it possible to detect from inside the function that it's just being warmed? This way we could avoiding performing any real work in the code.

serverless.yml
```diff
@@ -213,13 +222,15 @@ functions:
...
     handler: src/api/version.get
+    warmup: true
     events:
       - http:
           path: version
           method: get
           cors: true
   getTxHistoryApi:
     handler: src/api/txhistory.get
+    warmup: true
```
luislhl, line 237: On which basis did you select what functions need warmup?

serverless.yml
andreabadesso: It is, that is what I'm working right now

I want to use `middy` to detect when it's a warmup call (`event.source === 'serverless-plugin-warmup'`) and return immediately, saving some processing time

serverless.yml
andreabadesso: This is still a WIP, but I want to warmup all functions related to the app startup, I will document this on the PR before sending it to review

luislhl:
> The idea is to keep lambdas that are called from the wallets on load warm to prevent cold starts that take up to 10 seconds per request, causing the wallet load to take a long time.

This would probably solve for almost all cases right now, because we don't have so many users yet.

But at some point in the future when we have more users, if we have more than 1 person at the same time making requests to load the wallet, the lambda warm-up would only solve the cold-starts for 1 of them, right?

>  take up to 10 seconds per request

Did you measure this or is this based on some documentation?

andreabadesso:
@luislhl 

> But at some point in the future when we have more users, if we have more than 1 person at the same time making requests to load the wallet, the lambda warm-up would only solve the cold-starts for 1 of them, right?

The serverless-plugin-warmup has a `concurrency` option that calls the lambdas in parallel, as per documentation:

```
concurrency: The number of times that each of your lambda functions will be called in parallel. This can be used in a best-effort attempt to force AWS to spin up more parallel containers for your lambda. (defaults to 1)
```

When we start the rollout we can monitor the service to check if the lambdas are being cold started and increase this value. What do you think?

> Did you measure this or is this based on some documentation?

Just checked again the cold start times and it is averaging at 1-2s. But it is based on my observation only, not on any documentation from aws

luislhl:
> When we start the rollout we can monitor the service to check if the lambdas are being cold started and increase this value. What do you think?

How could we have an alert that warn us if they are being cold started?

One possible thing to monitor could be the start time of the wallets themselves. Monitoring this on the clients could generate privacy concerns, but maybe we could do this in the backend somehow. This way we could have alerts for when too many wallets are taking too long.

Another thing that is even easier is to monitor the `Concurrent executions` metric of the Lambdas, so we detect when we start having too many concurrent executions, which means we would benefit from increasing the `concurrency` option in the plugin.

What do you think?

Maybe these items could be added to the issue we have about monitoring the wallet-service

luislhl (approved):

r4mmer (approved):

PR #244, opened 2022-07-21
refactor: storing and using highest address index on the wallet table

andreabadesso:
### Acceptance Criteria
- We should store the highest used address index on the wallet table every time we generate new addresses
- We should use the highest used address on the `getNewAddresses` query to prevent a query on the big address table

### Security Checklist

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -80,7 +80,7 @@ export const generateAddresses = async (mysql: ServerlessMysql, xpubkey: string,
   const derivedXpub = xpubDeriveChild(xpubkey, 0);
 
   let highestCheckedIndex = -1;
-  let highestUsedIndex = -1;
+  let lastUsedAddressIndex = -1;
   do {
     const addrMap = getAddresses(derivedXpub, highestCheckedIndex + 1, maxGap);
     allAddresses.push(...Object.keys(addrMap));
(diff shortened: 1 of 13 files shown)


src/db/index.ts
```diff
@@ -127,6 +127,7 @@ export const generateAddresses = async (mysql: ServerlessMysql, xpubkey: string,
     addresses: allAddresses.slice(0, totalAddresses),
     newAddresses,
     existingAddresses,
+    highestUsedIndex,
```
tuliomir: I think it would be good to add an assertion that this new property is being returned with the correct value on each of the calls on the `tests/db.test.ts:130` suite.

src/db/index.ts
```diff
@@ -292,6 +298,14 @@ export const addNewAddresses = async (mysql: ServerlessMysql, walletId: string,
      VALUES ?`,
     [entries],
   );
+
+  // Store on the wallet table the highest used index
+  await mysql.query(
+    `UPDATE \`wallet\`
+        SET \`last_used_address_index\` = ?
```
luislhl, line 305: The terms `last used` and `highest used` don't necessarily mean the same thing, although in this context they do.

But wouldn't it be better to stick to one of them both in the DB field and in the JS variable?

luislhl (approved):

tests/utils.ts
```diff
@@ -498,15 +498,26 @@ export const addToUtxoTable = async (
 
 export const addToWalletTable = async (
   mysql: ServerlessMysql,
-  entries: unknown[][],
+  entries: WalletTableEntry[],
```
tuliomir, line 501: Nice improvement! :+1:

tuliomir (changes requested):
I think just the tests on the new field would be important.

src/db/index.ts
andreabadesso: That's a great idea, done in c981d1fed7515e8eafb3af6333008d8eee1f1085


Thanks

src/db/index.ts
andreabadesso: I agree, done in 775a69f0c0c858ac4321c79a069e09074ee10f0c

luislhl (approved):

tuliomir (approved):

PR #258, opened 2022-07-28
fix: re-calculating the transactions count correctly on a reorg or vo…

andreabadesso:
This PR fixes https://github.com/HathorNetwork/hathor-explorer-service/issues/179

### Acceptance Criteria
- We should re-calculate the number of transactions properly on re-org by using the number of transactions that were affected by the re-org

### Security Checklist

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -2040,7 +2040,7 @@ export const rebuildAddressBalancesFromUtxos = async (
             BIT_OR(\`authorities\`), -- unlocked_authorities
             0, -- locked_authorities
             0, -- timelock_expires
-            COUNT(DISTINCT \`tx_id\`) -- transactions
+            0 -- transactions
        FROM \`tx_output\`
       WHERE spent_by IS NULL
         AND voided = FALSE
@@ -2066,7 +2066,7 @@ export const rebuildAddressBalancesFromUtxos = async (
               SUM(\`value\`) AS locked_balance,
               BIT_OR(\`authorities\`) AS locked_authorities,
               MIN(\`timelock\`) AS timelock_expires,
-              COUNT(DISTINCT \`tx_id\`) -- transactions
+              0 -- transactions
          FROM \`tx_output\`
         WHERE spent_by IS NULL
           AND voided = FALSE
@@ -2076,8 +2076,25 @@ export const rebuildAddressBalancesFromUtxos = async (
    ON DUPLICATE KEY UPDATE
     locked_balance = VALUES(locked_balance),
     locked_authorities = VALUES(locked_authorities),
-    timelock_expires = VALUES(timelock_expires),
-    transactions = transactions + VALUES(\`transactions\`)
+    timelock_expires = VALUES(timelock_expires)
+   `, [addresses]);
+
+  // update address balances with the correct amount of transactions
+  await mysql.query(`
+    INSERT INTO \`address_balance\` (
+      \`address\`,
+      \`token_id\`,
+      \`transactions\`
+    )
+       SELECT address,
+              token_id,
+              COUNT(DISTINCT(\`tx_id\`)) AS transactions -- transactions
+         FROM \`tx_output\`
+        WHERE voided = FALSE
+          AND address IN (?)
+     GROUP BY \`address\`, \`token_id\`
+   ON DUPLICATE KEY UPDATE
+    transactions = VALUES(transactions)
    `, [addresses]);
 };
 
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -2015,6 +2015,11 @@ export const rebuildAddressBalancesFromUtxos = async (
   addresses: string[],
   txList: string[],
 ): Promise<void> => {
+  if (txList.length === 0) {
+    // This should never happen, we should throw so the re-org is rolled back
+    // and an error is triggered for manual inspection
+    throw new Error('Attempted to rebuild address balances but no transactions were affected');
+  }
   // first we need to store the transactions count before deleting
   const oldAddressTokenTransactions: DbSelectResult = await mysql.query(
     `SELECT \`address\`, \`token_id\` AS tokenId, \`transactions\`
@@ -2088,10 +2093,6 @@ export const rebuildAddressBalancesFromUtxos = async (
     locked_authorities = VALUES(locked_authorities),
     timelock_expires = VALUES(timelock_expires)
    `, [addresses]);
-
-  if (txList.length === 0) {
-    return;
-  }
 
   const addressTransactionCount: StringMap<number> = await getAffectedAddressTxCountFromTxList(mysql, txList);
   const finalTxCount = oldAddressTokenTransactions.map(({ address, tokenId, transactions }) => {
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -2498,6 +2498,15 @@ export const getAvailableAuthorities = async (
   return utxos;
 };
 
+/**
+ * Get the number of transactions for each token from the address_tx_history table
+ * given a list of transactions
+ *
+ * @param mysql - Database connection
+ * @param txList - A list of affected transactions to get the addresses token transaction count
+
+ * @returns A Map with address_tokenId as key and the transaction count as values
+ */
 export const getAffectedAddressTxCountFromTxList = async (
   mysql: ServerlessMysql,
   txList: string[],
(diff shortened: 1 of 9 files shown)


tests/api.test.ts
```diff
@@ -1265,9 +1265,9 @@ test('GET /wallet/tokens/token_id/details', async () => {
   ]);
 
   await addToAddressTxHistoryTable(mysql, [
-    [ADDRESSES[0], 'txId', token1.id, 100, 0],
-    [ADDRESSES[0], 'txId2', token2.id, 250, 0],
-    [ADDRESSES[0], 'txId3', token2.id, 0, 0],
+    { address: ADDRESSES[0], txId: 'txId', tokenId: token1.id, balance: 100, timestamp: 0 },
+    { address: ADDRESSES[0], txId: 'txId2', tokenId: token2.id, balance: 250, timestamp: 0 },
+    { address: ADDRESSES[0], txId: 'txId3', tokenId: token2.id, balance: 0, timestamp: 0 },
```
andreabadesso, line 1270: Just changed the signature of the `addToAddressTxHistoryTable` to receive an object (`AddressTxHistoryTableEntry`) instead of an array

tests/db.test.ts
```diff
@@ -390,14 +392,14 @@ test('initWalletTxHistory', async () => {
...
-    [addr2, txId2, token2, -5, timestamp2],
-    [addr3, txId2, token1, 3, timestamp2],
+    { address: addr1, txId: txId1, tokenId: token1, balance: 10, timestamp: timestamp1 },
+    { address: addr1, txId: txId1, tokenId: token2, balance: 7, timestamp: timestamp1 },
+    { address: addr2, txId: txId1, tokenId: token2, balance: 5, timestamp: timestamp1 },
+    { address: addr3, txId: txId1, tokenId: token1, balance: 3, timestamp: timestamp1 },
+    { address: addr1, txId: txId2, tokenId: token1, balance: -1, timestamp: timestamp2 },
+    { address: addr1, txId: txId2, tokenId: token3, balance: 3, timestamp: timestamp2 },
+    { address: addr2, txId: txId2, tokenId: token2, balance: -5, timestamp: timestamp2 },
+    { address: addr3, txId: txId2, tokenId: token1, balance: 3, timestamp: timestamp2 },
```
andreabadesso, line 402: Same here

tests/db.test.ts
```diff
@@ -1427,13 +1429,13 @@ test('fetchAddressTxHistorySum', async () => {
...
+    { address: addr1, txId: txId1, tokenId: token1, balance: 10, timestamp: timestamp1 },
+    { address: addr1, txId: txId2, tokenId: token1, balance: 20, timestamp: timestamp2 },
+    { address: addr1, txId: txId3, tokenId: token1, balance: 30, timestamp: timestamp2 },
     // total: 60
-    [addr2, txId1, token2, 20, timestamp1],
-    [addr2, txId2, token2, 20, timestamp2],
-    [addr2, txId3, token2, 10, timestamp2],
+    { address: addr2, txId: txId1, tokenId: token2, balance: 20, timestamp: timestamp1 },
+    { address: addr2, txId: txId2, tokenId: token2, balance: 20, timestamp: timestamp2 },
+    { address: addr2, txId: txId3, tokenId: token2, balance: 10, timestamp: timestamp2 },
```
andreabadesso, line 1438: Same here

tests/db.test.ts
```diff
@@ -1610,13 +1671,13 @@ test('markAddressTxHistoryAsVoided', async () => {
...
```
andreabadesso, line 1680: Same here

tests/db.test.ts
```diff
@@ -1926,11 +1987,11 @@ test('getTotalTransactions', async () => {
...
-    ['address1', 'txId1', 'token1', -5, 1000],
-    ['address1', 'txId2', 'token1', 5, 1000],
-    ['address1', 'txId3', 'token1', 10, 1000],
-    ['address2', 'txId4', 'token2', -5, 1000],
-    ['address2', 'txId5', 'token2', 50, 1000],
+    { address: 'address1', txId: 'txId1', tokenId: 'token1', balance: -5, timestamp: 1000 },
+    { address: 'address1', txId: 'txId2', tokenId: 'token1', balance: 5, timestamp: 1000 },
+    { address: 'address1', txId: 'txId3', tokenId: 'token1', balance: 10, timestamp: 1000 },
+    { address: 'address2', txId: 'txId4', tokenId: 'token2', balance: -5, timestamp: 1000 },
+    { address: 'address2', txId: 'txId5', tokenId: 'token2', balance: 50, timestamp: 1000 },
```
andreabadesso, line 1994: Same here

tests/utils.ts
```diff
@@ -574,27 +580,36 @@ export const addToAddressTable = async (
...
+    INSERT INTO \`address_tx_history\`(\`address\`, \`tx_id\`,
+                                       \`token_id\`, \`balance\`,
+                                       \`timestamp\`, \`voided\`)
     VALUES ?`,
-  [entries]);
+  [payload]);
 };
 
-export const addToAddressTxHistoryTable = async (
+export const addToAddressBalanceTable = async (
```
andreabadesso, line 604: Just changed the order of the method on the file with the `addToAddressTxHistoryTable` that had its parameter refactored

r4mmer (approved):

src/db/index.ts
```diff
@@ -2076,9 +2086,35 @@ export const rebuildAddressBalancesFromUtxos = async (
    ON DUPLICATE KEY UPDATE
     locked_balance = VALUES(locked_balance),
     locked_authorities = VALUES(locked_authorities),
-    timelock_expires = VALUES(timelock_expires),
-    transactions = transactions + VALUES(\`transactions\`)
+    timelock_expires = VALUES(timelock_expires)
    `, [addresses]);
+
+  if (txList.length === 0) {
```
pedroferreira1: If you return here, then it means that there are no txs affected, right? But you've updated the tx count to 0 anyway, is it correct?

src/db/index.ts
```diff
@@ -2076,9 +2086,35 @@ export const rebuildAddressBalancesFromUtxos = async (
...
+    return;
+  }
+
+  const addressTransactionCount: StringMap<number> = await getAffectedAddressTxCountFromTxList(mysql, txList);
+  const finalTxCount = oldAddressTokenTransactions.map(({ address, tokenId, transactions }) => {
+    const diff = addressTransactionCount[`${address}_${tokenId}`] || 0;
```
pedroferreira1, line 2099: So the idea is that you get the old count X and all the txs that are voided now Y and the new count is X - Y, right?

But if this reorg made a previous voided tx to become confirmed now, the new count should be X - Y + 1, right?

Am I missing something?

src/db/index.ts
andreabadesso: You are right

This should never happen, added the check on the first line of the method and throwing an error if this happen, as it would delete balances without rebuilding them

da9d7c0e7b0602c368ab533f55b2f66e3c433992

src/db/index.ts
andreabadesso: As we discussed yesterday, this is indeed a situation that we are not yet handling

We currently are not sending transaction updates in real-time from the daemon, we have an issue for that, and it will solve this situation (by unvoiding transactions and re-calculating their affected addresses balances and transaction count)

When we do, we should handle also this scenario (a transaction that was voided being unvoided), calculating the balances and transaction count properly

The way we void transactions that were not part of a re-org but were voided before being confirmed by a block is by using the `onHandleOldVoidedTxs` lambda, that is called every 20 minutes and voids checks if there is any transaction older than 20 minutes that were not confirmed by any block, voiding them after confirming if they are indeed voided on the fullnode (by using the transaction API)

Another situation we've discussed is the balance and transaction count re-calculation when a transaction destroys an authority, I've created an issue for that as well https://github.com/HathorNetwork/hathor-wallet-service/issues/268

tests/db.test.ts
```diff
@@ -1551,40 +1553,99 @@ test('rebuildAddressBalancesFromUtxos', async () => {
...
+  await addToAddressTxHistoryTable(mysql, txHistory);
+
+  // We are only using the txList parameter on `transactions` recalculation, so our balance
+  // checks should include txId3 and txId4, but the transaction count should not.
+  await rebuildAddressBalancesFromUtxos(mysql, [addr1, addr2], [txId3, txId4]);
 
   const addressBalances = await fetchAddressBalance(mysql, [addr1, addr2]);
 
-  expect(addressBalances[0].unlockedBalance).toStrictEqual(20);
+  expect(addressBalances[0].unlockedBalance).toStrictEqual(41);
```
pedroferreira1, line 1642: If we are only fixing the transaction count in this PR, why did the `unlockedBalance` changed?

pedroferreira1 (approved):

tests/db.test.ts
andreabadesso: Because I added more transactions (suming 21 for this address) to be able to re-calculate and test the result

PR #260, opened 2022-07-28
Release v0.15.0-alpha

tuliomir:
### Acceptance Criteria
- Release v0.15.0-alpha

### Security Checklist

### DIFF
--- a/.codebuild/buildspec.yml
+++ b/.codebuild/buildspec.yml
@@ -17,14 +17,20 @@ env:
     dev_WS_DOMAIN: "ws.dev.wallet-service.testnet.hathor.network"
     dev_NETWORK: "testnet"
     dev_LOG_LEVEL: "info"
+    dev_EXPLORER_SERVICE_LAMBDA_ENDPOINT: "https://lambda.eu-central-1.amazonaws.com"
+    dev_WALLET_SERVICE_LAMBDA_ENDPOINT: "https://lambda.eu-central-1.amazonaws.com"
     testnet_DEFAULT_SERVER: "https://wallet-service.private-nodes.testnet.hathor.network/v1a/"
     testnet_WS_DOMAIN: "ws.testnet.wallet-service.hathor.network"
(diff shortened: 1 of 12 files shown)


.codebuild/buildspec.yml
```diff
@@ -17,14 +17,20 @@ env:
     dev_WS_DOMAIN: "ws.dev.wallet-service.testnet.hathor.network"
     dev_NETWORK: "testnet"
     dev_LOG_LEVEL: "info"
+    dev_EXPLORER_SERVICE_LAMBDA_ENDPOINT: "https://lambda.eu-central-1.amazonaws.com"
```
luislhl, line 20: I suspect this should not be needed as the aws-sdk will have the right defaults for the endpoint.

We had to add this because it's required somewhere?

luislhl (approved):

.codebuild/buildspec.yml
tuliomir: This endpoint was actually hardcoded on the lambda invocation functions.
It was extracted to environment variables after [this](https://github.com/HathorNetwork/hathor-wallet-service/pull/235#discussion_r931535932) and [this](https://github.com/HathorNetwork/hathor-wallet-service/pull/235#discussion_r932249142) comments.

If you have a suggestion on how to improve this, maybe we could create a new issue for it?

luislhl: I guess something like this would work:

```js
    const endpoint = process.env.EXPLORER_SERVICE_LAMBDA_ENDPOINT;
    const region = process.env.AWS_REGION;
    
    const lambda = new Lambda({
      apiVersion: '2015-03-31',
      endpoint,
      region
    });
```

Then we wouldn't need the env var set in the deployed environments, only when testing locally, because it wouldn't be included in the object if it is `undefined`. And `AWS_REGION` is always available, so we wouldn't need to set it either, but we need to use it so the aws-sdk knows which endpoint to use.

But I didn't really test this to confirm

PR #269, opened 2022-08-08
fix: stop deleting address balance rows

andreabadesso:
## Motivation

[The queries we use to re-build](https://github.com/HathorNetwork/hathor-wallet-service/blob/9ba7f085fa50121d0912556f66237ece3f12ece9/src/db/index.ts#L2012) the `address_balance` table after a re-org or voided transaction are currently removing the affected rows

The problem with this approach is that if an address does not have at least one utxo, locked or unlocked, we would never add its row to the `address_balance` table again.

This is a problem for the new `big_addresses` feature that won't be able to display the `total_spent` and `total_received` information as it won't find the `address_balance` row for the addresses that are in this condition

An example is the `HCxPZcbWhiUT3zCx9rYbwGkQA1bzZKeEUe` address, that probably belongs to a pool that usually sends out every htr that it mines

Both queries for rebuilding the balances will return 0 results, causing it to never re-create the `address_balance` row.

### Acceptance Criteria
- When a re-org happens, we should not delete `address_balance` rows for affected addresses, instead we should set values to 0 (or NULL)

### Security Checklist

### DIFF
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -460,9 +460,6 @@ export const validateAddressBalances = async (mysql: ServerlessMysql, addresses:
   const addressBalances: AddressBalance[] = await fetchAddressBalance(mysql, addresses);
   const addressTxHistorySums: AddressTotalBalance[] = await fetchAddressTxHistorySum(mysql, addresses);
 
-  // if we have the full history of transactions, the number of rows must match:
-  assert.strictEqual(addressBalances.length, addressTxHistorySums.length);
-
   for (let i = 0; i < addressTxHistorySums.length; i++) {
     const addressBalance: AddressBalance = addressBalances[i];
     const addressTxHistorySum: AddressTotalBalance = addressTxHistorySums[i];
--- a/tests/txProcessor.test.ts
+++ b/tests/txProcessor.test.ts
@@ -517,7 +517,7 @@ test('onHandleVoidedTxRequest', async () => {
 
   // Balance should be back to 2500 as the transactions that spent the original utxo were voided and we should
   // have total of one transaction as both txId2 and txId3 were voided.
-  await expect(checkAddressBalanceTable(mysql, 1, addr, token, 2500, 0, 0, 1)).resolves.toBe(true);
+  await expect(checkAddressBalanceTable(mysql, 2, addr, token, 2500, 0, 0, 1)).resolves.toBe(true);
 }, 20000);
 
 test('txProcessor should rollback the entire transaction if an error occurs on balance calculation', async () => {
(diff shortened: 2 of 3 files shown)


src/commons.ts
```diff
@@ -460,9 +460,6 @@ export const validateAddressBalances = async (mysql: ServerlessMysql, addresses:
   const addressBalances: AddressBalance[] = await fetchAddressBalance(mysql, addresses);
   const addressTxHistorySums: AddressTotalBalance[] = await fetchAddressTxHistorySum(mysql, addresses);
 
-  // if we have the full history of transactions, the number of rows must match:
-  assert.strictEqual(addressBalances.length, addressTxHistorySums.length);
```
andreabadesso, line 464: This assert is no longer valid as we might have `address_balance` rows for addresses that don't have any valid  `address_tx_history` rows, here is an example:

```
> SELECT address, token_id, transactions FROM address_balance;

+---------+----------+--------------+
| address | token_id | transactions |
+---------+----------+--------------+
| address | tokenId  |            1 |
| other   | tokenId  |            0 |
+---------+----------+--------------+
2 rows in set (0.004 sec)

> SELECT * FROM address_tx_history;

+---------+-------+----------+---------+------------+--------+
| address | tx_id | token_id | balance | timestamp  | voided |
+---------+-------+----------+---------+------------+--------+
| address | txId1 | tokenId  |    2500 |          0 |      0 |
| address | txId2 | tokenId  |    -500 | 1590504219 |      1 |
| address | txId3 | tokenId  |    -500 | 1590503219 |      1 |
| other   | txId2 | tokenId  |     500 | 1590504219 |      1 |
| other   | txId3 | tokenId  |     500 | 1590503219 |      1 |
+---------+-------+----------+---------+------------+--------+
5 rows in set (0.002 sec)
```

The address `other` had records for `address_tx_history` but they got voided on a re-org, which updated the `address_balance` row for this address to zero for all columns and voided all `address_tx_history` rows

This is very unlikely to happen in production, but we should expect it

tests/txProcessor.test.ts
```diff
@@ -517,7 +517,7 @@ test('onHandleVoidedTxRequest', async () => {
 
   // Balance should be back to 2500 as the transactions that spent the original utxo were voided and we should
   // have total of one transaction as both txId2 and txId3 were voided.
-  await expect(checkAddressBalanceTable(mysql, 1, addr, token, 2500, 0, 0, 1)).resolves.toBe(true);
+  await expect(checkAddressBalanceTable(mysql, 2, addr, token, 2500, 0, null, 1)).resolves.toBe(true);
```
andreabadesso, line 520: The `null` part is a bugfix, shouldn't have any practical effect, but we are now setting the `timelock_expires` column to `NULL` instead of `0` on the query above

r4mmer (approved):

pedroferreira1 (approved):

PR #273, opened 2022-08-11
feat: transactions count on token table

andreabadesso:
### Motivation

We have a feature on the explorer that displays the total number of transactions for a given token.

Currently, we are using the transactions column from the `address_balance` table to display it, by summing all values for a given `token_id`

The problem with this approach is that there are transactions with more than one address, so the final count of transactions will have duplicate transactions being counted

The idea here is to start counting the number of transactions on the `onNewTxRequest` lambda for each token and also fixing the column after a re-org or voided tx happens

### Acceptance Criteria
- We should add a counter to the `token` table to sum the number of transactions for each row
- We should handle re-orgs and voided transactions properly, reducing from this number of transactions the number of transactions voided

### Security Checklist

### DIFF
--- a/db/migrations/20220811222729-add-voided-index-to-address-tx-history.js
+++ b/db/migrations/20220811222729-add-voided-index-to-address-tx-history.js
@@ -1,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface) => queryInterface.addIndex(
+    'address_tx_history',
+    ['voided'], {
+      name: 'address_tx_history_voided_idx',
+      fields: 'voided',
+    },
+  ),
+  down: async (queryInterface) => queryInterface.removeIndex(
+    'address_tx_history',
+    'address_tx_history_voided_idx',
+  ),
+};
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -191,6 +191,20 @@ export const getAddressBalanceMap = (
   return addressBalanceMap;
 };
 
+export const getTokenListFromInputsAndOutputs = (inputs: TxInput[], outputs: TxOutput[]): string[] => {
+  const tokenIds = new Set<string>([]);
+
+  for (const input of inputs) {
+    tokenIds.add(input.token);
+  }
+
+  for (const output of outputs) {
+    tokenIds.add(output.token);
+  }
+
+  return [...tokenIds];
+};
+
 /**
  * Get the map of token balances for each wallet.
  *
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -191,6 +191,13 @@ export const getAddressBalanceMap = (
   return addressBalanceMap;
 };
 
+/**
+ * Gets a list of tokens from a list of inputs and outputs
+ *
+ * @param inputs - The transaction inputs
+ * @param outputs - The transaction outputs
+ * @returns A list of tokens present in the inputs and outputs
+ */
 export const getTokenListFromInputsAndOutputs = (inputs: TxInput[], outputs: TxOutput[]): string[] => {
   const tokenIds = new Set<string>([]);
 
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -198,7 +198,7 @@ export const getAddressBalanceMap = (
  * @param outputs - The transaction outputs
  * @returns A list of tokens present in the inputs and outputs
  */
-export const getTokenListFromInputsAndOutputs = (inputs: TxInput[], outputs: TxOutput[]): string[] => {
+export const getTokenListFromInputsAndOutputs = (inputs: TxInput[], outputs: TxOutputWithIndex[]): string[] => {
   const tokenIds = new Set<string>([]);
 
   for (const input of inputs) {
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -1255,7 +1255,6 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
     params.push(tokenIds);
   }
 
-  // use LEFT JOIN as HTR token ('00') won't be on the token table, so INNER JOIN would never match it
   const query = `
     SELECT NULL AS total_received,
            w.unlocked_balance AS unlocked_balance,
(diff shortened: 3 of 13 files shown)


src/db/index.ts
```diff
@@ -1255,8 +1255,20 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
...
+           w.unlocked_balance AS unlocked_balance,
+           w.locked_balance AS locked_balance,
+           w.unlocked_authorities AS unlocked_authorities,
+           w.locked_authorities AS locked_authorities,
+           w.timelock_expires AS timelock_expires,
+           w.transactions AS transactions,
+           w.token_id AS token_id,
+           token.name AS name,
+           token.symbol AS symbol
+      FROM (${subquery}) w
```
andreabadesso, line 1269: Had to specify fields as the `transactions` column would be ambiguous and default to the `wallet_balance` transactions count

src/db/index.ts
```diff
@@ -1255,8 +1255,20 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
...
+INNER JOIN token ON w.token_id = token.id
```
andreabadesso, line 1270: We now have the HTR token on the token table by default, so it is safe to use an `INNER JOIN`

src/db/index.ts
```diff
@@ -1255,8 +1255,20 @@ export const getWalletBalances = async (mysql: ServerlessMysql, walletId: string
     params.push(tokenIds);
   }
 
-  // use LEFT JOIN as HTR token ('00') won't be on the token table, so INNER JOIN would never match it
-  const query = `SELECT * FROM (${subquery}) w LEFT JOIN token ON w.token_id = token.id;`;
+  const query = `
+    SELECT NULL AS total_received,
```
andreabadesso, line 1259: This column is missing on the `wallet_balance` table, this will be fixed on another PR by @r4mmer

db/migrations/20220811222729-add-voided-index-to-address-tx-history.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface) => queryInterface.addIndex(
+    'address_tx_history',
```
r4mmer, line 5: We need to add this index to the models

src/db/index.ts
r4mmer: Created a new [PR](https://github.com/HathorNetwork/hathor-wallet-service/pull/274) for this.

tests/utils.ts
```diff
@@ -458,6 +458,51 @@ export const checkWalletBalanceTable = async (
   return true;
 };
 
+export const checkTokenTable = async (
```
r4mmer, line 469: Do you think we should split this helper in 2?
A `checkTokenTable` which checks that we have the expected number of tokens (and maybe add a list of token ids or names).
And a `checkTokenEntry` which actually checks a single entry (id, symbol, name and transactions).

tests/db.test.ts
```diff
@@ -2121,3 +2132,20 @@ test('getAffectedAddressTxCountFromTxList', async () => {
...
+
+  await addToTokenTable(mysql, [
+    { id: token1.id, name: token1.name, symbol: token1.symbol, transactions: token1.transactions },
+    { id: token2.id, name: token2.name, symbol: token2.symbol, transactions: token2.transactions },
+  ]);
+
+  await incrementTokensTxCount(mysql, ['token1', '00', 'token2']);
+
+  await expect(checkTokenTable(mysql, 2, token1.id, token1.symbol, token1.name, token1.transactions + 1)).resolves.toBe(true);
+  await expect(checkTokenTable(mysql, 2, token1.id, token1.symbol, token1.name, token1.transactions + 1)).resolves.toBe(true);
```
pedroferreira1: Should this be for `token2`?

tests/db.test.ts
```diff
@@ -2121,3 +2132,20 @@ test('getAffectedAddressTxCountFromTxList', async () => {
...
+});
```
pedroferreira1, line 2177: Should we add one more check for '00' token?

src/commons.ts
```diff
@@ -191,6 +191,27 @@ export const getAddressBalanceMap = (
...
 };
 
+/**
+ * Gets a list of tokens from a list of inputs and outputs
+ *
+ * @param inputs - The transaction inputs
+ * @param outputs - The transaction outputs
+ * @returns A list of tokens present in the inputs and outputs
+ */
+export const getTokenListFromInputsAndOutputs = (inputs: TxInput[], outputs: TxOutput[]): string[] => {
```
pedroferreira1: I think the outputs list from the `txProcessor` method that calls this method here is `TxOutputWithIndex`

tests/db.test.ts
andreabadesso: Yes we should, done in fee2282070e680130ebef288db37a929e7e582c9

tests/db.test.ts
andreabadesso: Fixed in fee2282070e680130ebef288db37a929e7e582c9

src/commons.ts
andreabadesso: You are right, done in 82cf24deb017cfd328f3498a023bb764c6fc1617

db/migrations/20220811222729-add-voided-index-to-address-tx-history.js
andreabadesso: Thanks, done in 54d2700d7c045995c5e06e8d2c4084b52c915d92

tests/utils.ts
andreabadesso: Refactored the `checkTokenTable` to receive a list of tokens and check for equality, what do you think?

058d05127fcbea1256810931d72338a81c6abe1e

src/commons.ts
```diff
@@ -191,6 +191,27 @@ export const getAddressBalanceMap = (
...
+ */
+export const getTokenListFromInputsAndOutputs = (inputs: TxInput[], outputs: TxOutputWithIndex[]): string[] => {
+  const tokenIds = new Set<string>([]);
+
+  for (const input of inputs) {
+    tokenIds.add(input.token);
+  }
+
+  for (const output of outputs) {
+    tokenIds.add(output.token);
```
r4mmer, line 209: Seeing that we need to catch the HTR token to increase the count, should we use `output.token || '00'`? Same on input?

r4mmer (approved):

pedroferreira1 (approved):

PR #293, opened 2022-10-03
chore: exit if migration fails

andreabadesso:
### Acceptance Criteria
- We should make sure that if the migration fails the build will fail without deploying the lambdas

### Security Checklist

### DIFF
--- a/.codebuild/buildspec.yml
+++ b/.codebuild/buildspec.yml
@@ -101,6 +101,10 @@ phases:
           done
 
           make migrate;
+          if [ $? -ne 0 ]; then
+            echo "Dev migration failed"
+            exit 1
+          fi
           make deploy-lambdas-dev-testnet;
         fi
       - |
@@ -111,6 +115,10 @@ phases:
           done
 
           make migrate;
+          if [ $? -ne 0 ]; then
+            echo "Testnet migration failed"
+            exit 1
+          fi
           make deploy-lambdas-testnet;
         fi
       - |
@@ -121,6 +129,10 @@ phases:
           done
 
           make migrate;
+          if [ $? -ne 0 ]; then
+            echo "Mainnet migration failed"
+            exit 1
+          fi
           make deploy-lambdas-mainnet;
         fi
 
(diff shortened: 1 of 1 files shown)


r4mmer (approved):

.codebuild/buildspec.yml
```diff
@@ -115,6 +117,7 @@ phases:
         fi
       - |
         if expr "${GIT_REF_TO_DEPLOY}" : "v.*" >/dev/null; then
+          set -e;
```
luislhl: I would say that dropping the yaml items and putting everything in a single item could allow us to use only one `set -e` at the beginning. Like in the example below.

But I'm not really sure since I didn't test. I'm actually ok with the current solution as well.

```yaml
  build:
    commands:
      - |
        set -e;
        if expr "${GIT_REF_TO_DEPLOY}" : "dev" >/dev/null; then
          # Gets all env vars with `dev_` prefix and re-exports them without the prefix
          for var in "${!dev_@}"; do
            export ${var#dev_}=${!var}
          done
          make migrate;
          make deploy-lambdas-dev-testnet;
        fi;
        
        if expr "${GIT_REF_TO_DEPLOY}" : "master" >/dev/null; then
          # Gets all env vars with `testnet_` prefix and re-exports them without the prefix
          for var in "${!testnet_@}"; do
            export ${var#testnet_}=${!var}
          done
          make migrate;
          make deploy-lambdas-testnet;
        fi;
        
        if expr "${GIT_REF_TO_DEPLOY}" : "v.*" >/dev/null; then
          # Gets all env vars with `mainnet_` prefix and re-exports them without the prefix
          for var in "${!mainnet_@}"; do
            export ${var#mainnet_}=${!var}
          done
          make migrate;
          make deploy-lambdas-mainnet;
        fi;
```

luislhl (approved):

.codebuild/buildspec.yml
andreabadesso: Done and tested in 5a01c517ec5088f29573ac6adeb5e3456cefc0d5

PR #299, opened 2022-11-09
feat: create push_devices table

alexruzenhack:
### Acceptance Criteria
- create the table `push_devices`.

### Security Checklist

### DIFF
--- a/db/migrations/20221108235926-create-pushdevices.js.js
+++ b/db/migrations/20221108235926-create-pushdevices.js.js
@@ -1,0 +1,50 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.createTable('push_devices', {
+      device_id: {
+        type: Sequelize.STRING(256),
+        allowNull: false,
+        primaryKey: true,
+      },
+      push_provider: {
+        type: Sequelize.ENUM(['ios', 'android']),
+        allowNull: false,
+        defaultValue: 'android',
+      },
+      wallet_id: {
+        type: Sequelize.STRING(64),
+        allowNull: false,
+        references: {
+          model: 'wallet',
+          key: 'id',
+        },
+      },
+      enable_push: {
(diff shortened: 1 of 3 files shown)


db/migrations/20221108235926-create-pushdevices.js.js
```diff
@@ -0,0 +1,50 @@
...
+    await queryInterface.createTable('push_devices', {
+      device_id: {
+        type: Sequelize.STRING(256),
+        allowNull: false,
+        primaryKey: true,
+      },
+      push_provider: {
+        type: Sequelize.ENUM(['ios', 'android']),
+        allowNull: false,
+        defaultValue: 'android',
```
tuliomir: I don't think we should have a default value here. If some code tries to register a device without specifying the provider, I think it's better to throw an error than to suppose a value.

What do you think about this?

tuliomir (changes requested):
I think we should also update the [`/DATABASE.md` file](https://github.com/HathorNetwork/hathor-wallet-service/blob/master/DATABASE.md) with the table's description.

It's a very helpful consolidated reference for understanding the database as a whole.

db/migrations/20221108235926-create-pushdevices.js.js
```diff
@@ -0,0 +1,50 @@
...
+        defaultValue: false,
+      },
+      updated_at: {
+        type: 'TIMESTAMP',
+        allowNull: false,
+        defaultValue: Sequelize.literal('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'),
+      },
+    });
+  },
+  async down(queryInterface, Sequelize) {
```
andreabadesso:
```suggestion
  async down(queryInterface) {
```

andreabadesso (changes requested):

alexruzenhack:
Task of https://github.com/HathorNetwork/hathor-wallet-service/issues/300

tuliomir (approved):

andreabadesso (approved):

PR #301, opened 2022-11-10
feat: create push/register lambda

alexruzenhack:
### Acceptance Criteria
- create push/register lambda

### Security Checklist


src/api/pushRegister.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  if (error) {
+    const details = error.details.map((err) => ({
+      message: err.message,
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
+  const body: PushRegister = value;
```
tuliomir: I wish there was a way to insert this validation within Joi, like a `bodySchema.validate<PushRegister>()`, but I couldn't find a way to do it.

It seems this is actually the best syntax for this case.

src/db/index.ts
```diff
@@ -2664,3 +2665,43 @@ export const incrementTokensTxCount = async (
...
+  );
+};
+
+/**
+ * Remove any record of push notification device given a device ID.
+ *
+ * @param mysql - Database connection
+ * @param deviceId - The device ID
+ */
+export const removeAllPushDeviceByDeviceId = async (mysql: ServerlessMysql, deviceId: string): Promise<void> => {
```
tuliomir: I would just suggest having the name in plural, since we will potentially affect multiple rows
```suggestion
export const removeAllPushDevicesByDeviceId = async (mysql: ServerlessMysql, deviceId: string): Promise<void> => {
```

src/api/pushRegister.ts
```diff
@@ -0,0 +1,66 @@
...
+import { ApiError } from '@src/api/errors';
+import { closeDbAndGetError, warmupMiddleware } from '@src/api/utils';
+import { removeAllPushDeviceByDeviceId, registerPushDevice } from '@src/db';
+import { getDbConnection } from '@src/utils';
+import { walletIdProxyHandler } from '@src/commons';
+import middy from '@middy/core';
+import cors from '@middy/http-cors';
+import Joi from 'joi';
+import { PushRegister } from '@src/types';
+import _ from 'lodash';
```
tuliomir: An unused import "`_`" is being warned by the CI

tests/pushRegister.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      pushProvider,
+      enablePush: true,
+      enableShowAmounts: false,
+    });
+
+    const result = await register(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(400);
+    expect(returnBody.success).toStrictEqual(false);
```
tuliomir, line 185: I feel we should check the `returnBody` for the error message, to make sure the client is having adequate feedback on the failure.

tests/db.test.ts
```diff
@@ -2175,3 +2178,51 @@ test('incrementTokensTxCount', async () => {
     transactions: htr.transactions + 1,
   }])).resolves.toBe(true);
 });
+
+test('registerPushDevice', async () => {
```
tuliomir, line 2217: While I like this smoke test here, I feel we should have a more extensive testing suite on the dedicated `pushRegister.test` file.

I can think of some cases would be nice to have:
- check if the values added to the database are correct when
  - there is no row to be deleted
  - there is already a row with the same deviceId and walletId, to ensure the updated values are persisted
- check that the error message of missing foreign key is being treated when trying to register for a wallet-id that is not on the `wallet` table
- ensuring that only the correct rows are affected by a delete command

tuliomir (changes requested):

src/types.ts
```diff
@@ -672,3 +672,13 @@ export interface Miner {
   lastBlock: string;
   count: number;
 }
+
+type iOSPushProvider = 'ios';
```
pedroferreira1: Maybe an `enum` would fit better here, what do you think? Something similar to

```
export enum TxProposalStatus {
  OPEN = 'open',
  SENT = 'sent',
  SEND_ERROR = 'send_error',
  CANCELLED = 'cancelled',
}
```

src/db/index.ts
```diff
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+/* eslint-disable camelcase */
```
pedroferreira1: Why do you need to add this?

src/db/index.ts
pedroferreira1: Why? Isn't `device_id` a primary key in the `push_devices` table?

src/api/pushRegister.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const mysql = getDbConnection();
+
+const bodySchema = Joi.object({
+  pushProvider: Joi.string().pattern(new RegExp('^(?:android|ios)$')).required(),
+  deviceId: Joi.string().max(256).required(),
+  enablePush: Joi.boolean().optional(),
```
pedroferreira1: Is the body does not have this parameter, what will we set in the db?

The method `registerPushDevice` expects a boolean here, will it have `undefined`? If yes, maybe we should have here `false` as default.

(the same applies to `enableShowAmounts`)

tests/utils.ts
```diff
@@ -68,11 +68,13 @@ export const cleanDatabase = async (mysql: ServerlessMysql): Promise<void> => {
     'wallet_balance',
     'wallet_tx_history',
     'miner',
+    'push_devices',
   ];
-
+  await mysql.query('SET FOREIGN_KEY_CHECKS = 0');
```
pedroferreira1, line 73: Why did you add this?

src/api/pushRegister.ts
alexruzenhack: Ok, agree. I was relying on the database default, which is false, but you are right, better protect the value from the beginning.

src/db/index.ts
alexruzenhack: Yes, it is. I see this removeAll as a double check. In the future, if we want to enable 1:N wallet->device push notification we just remove `device_id` as the primary key.

src/types.ts
alexruzenhack: Humm.. sure, it works better.

tests/utils.ts
alexruzenhack: Because the table `push_devices` has a relationship with `wallet`, therefore the delete does not work with active constraints, then I need to deactivate it before deleting and reactive it after deleting.

src/db/index.ts
alexruzenhack: Because lint was accusing and error for camelcase rule. But it is not necessary anymore.

tests/pushRegister.test.ts
alexruzenhack: I have added error message assert because I think the details of validation should not be asserted to avoid making the tests fragile.

tests/db.test.ts
alexruzenhack: I believe the update should not be tested because it is a very basic operation in the database realm.

The test to verify the right delete is good.

alexruzenhack: I implemented a wallet verification before calling the push device registration. I believe it will be cheaper than handling errors.

src/api/pushRegister.ts
pedroferreira1: I saw how you implemented the default value and it's okay but if there's a way to add this to the `bodySchema` using jest (a default value to each optional parameter), I think would be better.

src/db/index.ts
```diff
@@ -2664,3 +2664,62 @@ export const incrementTokensTxCount = async (
...
+
+/**
+ * Verify the existence of a wallet by its ID.
+ *
+ * @param mysql - Database connection
+ * @param walletId - The wallet linked to device
+ */
+export const existsWallet = async (
+  mysql: ServerlessMysql,
+  deviceId: string,
```
pedroferreira1, line 2777: Should this be `walletId`? And this db field has index?

alexruzenhack: It is a primary key, then it is indexed. I will fix the name.

tests/pushRegister.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    }]);
+
+    const payloadWithoutEnablePush = {
+      deviceId: 'device1',
+      pushProvider: 'android',
+      enableShowAmounts: false,
+    };
+    const event = makeGatewayEventWithAuthorizer('my-wallet', null, payloadWithoutEnablePush);
+
+    const result = await register(event, null, null) as APIGatewayProxyResult;
```
tuliomir, line 72: I like this test suite, but could not feel confident that the results on the database are actually what we expect.

From the tests we know that passing a payload without push does not _throw_, but does it add the corrrect value to the table?

I think a simple assert like below could fix this:
```ts
await expect(checkPushDevicesTable(mysql, 1, {
    walletId: 'my-wallet',
    deviceId: 'device1',
    pushProvider: 'android',
    enablePush: false,
    enableShowAmounts: false,
  })).resolves.toBe(true);
```
What do you think about this?

src/api/pushRegister.ts
alexruzenhack: Hi, it is already done.

tuliomir (approved):

pedroferreira1 (approved):

src/db/index.ts
```diff
@@ -2664,3 +2664,62 @@ export const incrementTokensTxCount = async (
...
+ *
+ * @param mysql - Database connection
+ * @param deviceId - The device ID
+ */
+export const removeAllPushDevicesByDeviceId = async (mysql: ServerlessMysql, deviceId: string): Promise<void> => {
+  await mysql.query(
+    `
+    DELETE FROM \`push_devices\`
+    WHERE device_id = ?
+    `, [deviceId],
```
```suggestion
    `
    DELETE
      FROM \`push_devices\`
     WHERE device_id = ?
    `, [deviceId],
```

src/db/index.ts
```diff
@@ -2664,3 +2664,62 @@ export const incrementTokensTxCount = async (
...
+ * @param mysql - Database connection
+ * @param walletId - The wallet linked to device
+ */
+export const existsWallet = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+) : Promise<boolean> => {
+  const [{ count }] = await mysql.query(
+    `
+    SELECT COUNT(1) as \`count\` FROM \`wallet\` pd WHERE id= ?`,
```
```suggestion
    SELECT COUNT(1) as \`count\`
      FROM \`wallet\` pd
     WHERE id= ?`,

```

tests/utils.ts
```diff
@@ -900,3 +902,56 @@ export const getAuthData = (now: number): any => {
...
+
+  // now fetch the exact entry
+  const baseQuery = `
+    SELECT *
+      FROM \`push_devices\`
+    WHERE \`wallet_id\` = ?
+      AND \`device_id\` = ?
+      AND \`push_provider\` = ?
+      AND \`enable_push\` = ?
+      AND \`enable_show_amounts\` = ?
```
andreabadesso:
```suggestion
    SELECT *
      FROM \`push_devices\`
     WHERE \`wallet_id\` = ?
       AND \`device_id\` = ?
       AND \`push_provider\` = ?
       AND \`enable_push\` = ?
       AND \`enable_show_amounts\` = ?
```

andreabadesso (changes requested):

src/db/index.ts
```diff
@@ -2664,3 +2664,74 @@ export const incrementTokensTxCount = async (
...
+ */
+export const existsWallet = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+) : Promise<boolean> => {
+  const [{ count }] = (await mysql.query(
+    `
+    SELECT COUNT(1) as \`count\`
+      FROM \`wallet\` pd
+     WHERE id= ?`,
```
```suggestion
     WHERE id = ?`,
```

andreabadesso (approved):

PR #302, opened 2022-11-11
feat: create push/update lambda

alexruzenhack:
### Acceptance Criteria
- create push/update lambda

### Security Checklist

### DIFF
--- a/src/api/pushUpdate.ts
+++ b/src/api/pushUpdate.ts
@@ -8,13 +8,13 @@
 import { APIGatewayProxyHandler } from 'aws-lambda';
 import { ApiError } from '@src/api/errors';
 import { closeDbAndGetError, warmupMiddleware } from '@src/api/utils';
-import { registerPushDevice } from '@src/db';
+import { existsPushDevice, updatePushDevice } from '@src/db';
 import { getDbConnection } from '@src/utils';
 import { walletIdProxyHandler } from '@src/commons';
 import middy from '@middy/core';
 import cors from '@middy/http-cors';
 import Joi from 'joi';
-import { PushRegister } from '@src/types';
+import { PushUpdate } from '@src/types';
 
 const mysql = getDbConnection();
 
@@ -45,11 +45,16 @@ export const update: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (
     return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
   }
 
-  const body: PushRegister = value;
-  await registerPushDevice(mysql, {
+  const body: PushUpdate = value;
+
+  const deviceExists = await existsPushDevice(mysql, body.deviceId, walletId);
+  if (!deviceExists) {
+    return closeDbAndGetError(mysql, ApiError.DEVICE_NOT_FOUNT);
+  }
+
+  await updatePushDevice(mysql, {
     walletId,
     deviceId: body.deviceId,
-    pushProvider: body.pushProvider,
     enablePush: body.enablePush,
     enableShowAmounts: body.enableShowAmounts,
     enableOnlyNewTx: false,
--- a/src/api/pushUpdate.ts
+++ b/src/api/pushUpdate.ts
@@ -13,16 +13,26 @@ import { getDbConnection } from '@src/utils';
 import { walletIdProxyHandler } from '@src/commons';
 import middy from '@middy/core';
 import cors from '@middy/http-cors';
-import Joi from 'joi';
+import Joi, { ValidationError } from 'joi';
 import { PushUpdate } from '@src/types';
 
 const mysql = getDbConnection();
 
-const bodySchema = Joi.object({
-  deviceId: Joi.string().max(256).required(),
-  enablePush: Joi.boolean().optional(),
-  enableShowAmounts: Joi.boolean().optional(),
-});
+class PushUpdateInputValidator {
+  static readonly bodySchema = Joi.object({
+    deviceId: Joi.string().max(256).required(),
+    enablePush: Joi.boolean().optional(),
+    enableShowAmounts: Joi.boolean().optional(),
+  });
+
+  static validate(payload: unknown): { value: PushUpdate, error: ValidationError } {
+    const { value, error } = PushUpdateInputValidator.bodySchema.validate(payload, {
+      abortEarly: false, // We want it to return all the errors not only the first
+      convert: true, // We need to convert as parameters are sent on the QueryString
+    });
+    return { value: { enablePush: false, enableShowAmounts: false, ...value }, error };
+  }
+}
 
 /*
  * Update a device to receive push notification.
@@ -30,10 +40,7 @@ const bodySchema = Joi.object({
  * This lambda is called by API Gateway on POST /push/register
  */
 export const update: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (walletId, event) => {
-  const { value, error } = bodySchema.validate(event.body, {
-    abortEarly: false, // We want it to return all the errors not only the first
-    convert: true, // We need to convert as parameters are sent on the QueryString
-  });
+  const { value: body, error } = PushUpdateInputValidator.validate(event.body);
 
   if (error) {
     const details = error.details.map((err) => ({
@@ -43,8 +50,6 @@ export const update: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (
 
     return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
   }
-
-  const body: PushUpdate = value;
 
   const deviceExists = await existsPushDevice(mysql, body.deviceId, walletId);
   if (!deviceExists) {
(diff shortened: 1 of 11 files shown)


src/api/pushUpdate.ts
```diff
@@ -0,0 +1,67 @@
...
+import middy from '@middy/core';
+import cors from '@middy/http-cors';
+import Joi from 'joi';
+import { PushUpdate } from '@src/types';
+
+const mysql = getDbConnection();
+
+const bodySchema = Joi.object({
+  deviceId: Joi.string().max(256).required(),
+  enablePush: Joi.boolean().optional(),
```
pedroferreira1: Same comment about the default value in the previous PR. If we decide to change anything there, then we must change here as well

src/db/index.ts
```diff
@@ -2666,6 +2666,28 @@ export const incrementTokensTxCount = async (
   `, [tokenList]);
 };
 
+/**
+ * Verify the existence of a device registered for a given wallet.
+ *
+ * @param mysql - Database connection
+ * @param input - Input of push device register
```
pedroferreira1: We don't have this parameter

src/db/index.ts
```diff
@@ -2705,3 +2727,29 @@ export const removeAllPushDeviceByDeviceId = async (mysql: ServerlessMysql, devi
     `, [deviceId],
   );
 };
+
+/**
+ * Update existing push device given a wallet.
+ *
+ * @param mysql - Database connection
+ * @param input - Input of push device register
+ * @param deviceId - The device to verify existence
```
pedroferreira1: We don't have this parameter

src/db/index.ts
```diff
@@ -2705,3 +2727,29 @@ export const removeAllPushDeviceByDeviceId = async (mysql: ServerlessMysql, devi
...
+ * @param walletId - The wallet linked to device
```
pedroferreira1: We don't have this parameter

tests/db.test.ts
```diff
@@ -2198,7 +2231,47 @@ test('registerPushDevice', async () => {
...
+  await expect(checkPushDevicesTable(mysql, 1, {
+    walletId,
+    deviceId,
+    pushProvider,
+    enablePush,
+    enableShowAmounts,
+  })).resolves.toBe(true);
+});
+
+test('updatePushDevice', async () => {
```
tuliomir: I would also suggest adding tests to failure cases, to improve our code security:
- Missing parameters on the inputs
- Invalid parameters
- Error handling, ensure a developer-friendly error was raised to the controller function
- When there are multiple devices for the same wallet, if only the correct device is updated

tests/pushUpdate.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      deviceId,
+      enablePush: false,
+      enableShowAmounts: false,
+    });
+
+    const result = await update(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(400);
+    expect(returnBody.success).toStrictEqual(false);
```
tuliomir, line 199: We should also assert the error message raised here, to ensure it's helpful to the client.

tests/pushUpdate.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      deviceId,
+      enablePush: false,
+      enableShowAmounts: false,
+    });
+
+    const result = await update(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(404);
+    expect(returnBody.success).toStrictEqual(false);
```
tuliomir, line 229: Message should also be asserted here

src/api/pushUpdate.ts
```diff
@@ -0,0 +1,67 @@
...
+    enableShowAmounts: body.enableShowAmounts,
+  });
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ success: true }),
+  };
+}))
+  .use(cors())
+  .use(warmupMiddleware());
```
tuliomir: The `serverless.yml` file indicates that this route will not have the `walletWarmer` enabled.
Should we actually use this middleware here?

tuliomir (changes requested):

src/db/index.ts
```diff
@@ -2705,3 +2727,29 @@ export const removeAllPushDeviceByDeviceId = async (mysql: ServerlessMysql, devi
...
+    walletId: string,
+    enablePush: boolean,
+    enableShowAmounts: boolean,
+  },
+) : Promise<void> => {
+  await mysql.query(
+    `
+    UPDATE \`push_devices\`
+    SET enable_push = ?, enable_show_amounts = ?
+    WHERE device_id = ? AND wallet_id = ?`,
```
pedroferreira1: Do we have index for `device_id`and `wallet_id`? Should we have it?

alexruzenhack: No, we don't need a composed index for now, because the cardinality for device_id related to wallet_id is 1. If we change this relation allowing 1 to N then we can rethink a composed index. However, even for a 1-to-N relation, we need to take into consideration the cardinality involved in the average case.

tests/db.test.ts
alexruzenhack: > Missing parameters on the inputs

We can guarantee that all the parameters will be passed from lambda. As this method is not used in any other place nor intended for it, I believe it would be overkill.

> Invalid parameters

We already validated these parameters in the lambda. I don't see the necessity to validate our own code inputs, because I consider it to be a "safe environment".

> Error handling, ensure a developer-friendly error was raised to the controller function

I don't see an error path for this code. This is a basic database operation. I believe this kind of error should be held by a global error handler. Because there is nothing the controller can do with it.

> When there are multiple devices for the same wallet, if only the correct device is updated

Ok! Got it.

tests/db.test.ts
```diff
@@ -2240,38 +2241,98 @@ test('registerPushDevice', async () => {
...
+      enableShowAmounts,
+    });
+
+    await expect(checkPushDevicesTable(mysql, 1, {
+      walletId,
+      deviceId: deviceToUpdate,
+      pushProvider,
+      enablePush: true,
+      enableShowAmounts,
+    })).resolves.toBe(true);
```
tuliomir, line 2316: I think it's good to validate also that the `device2` was not updated.

tuliomir (approved):

tests/db.test.ts
alexruzenhack: Done!

pedroferreira1 (approved):

src/db/index.ts
```diff
@@ -2665,6 +2665,27 @@ export const incrementTokensTxCount = async (
...
+ * @param walletId - The wallet linked to device
+ */
+export const existsPushDevice = async (
+  mysql: ServerlessMysql,
+  deviceId: string,
+  walletId: string,
+) : Promise<boolean> => {
+  const [{ count }] = await mysql.query(
+    `
+    SELECT COUNT(1) as \`count\` FROM \`push_devices\` pd WHERE device_id = ? AND wallet_id = ?`,
```
```suggestion
SELECT COUNT(1) as \`count\`
  FROM \`push_devices\` pd
 WHERE device_id = ?
   AND wallet_id = ?
```

src/db/index.ts
```diff
@@ -2706,11 +2727,35 @@ export const removeAllPushDevicesByDeviceId = async (mysql: ServerlessMysql, dev
...
+    walletId: string,
+    enablePush: boolean,
+    enableShowAmounts: boolean,
+  },
+) : Promise<void> => {
+  await mysql.query(
+    `
+    UPDATE \`push_devices\`
+    SET enable_push = ?, enable_show_amounts = ?
+    WHERE device_id = ? AND wallet_id = ?`,
```
```suggestion
UPDATE \`push_devices\`
   SET \`enable_push\` = ?,
       \`enable_show_amounts\` = ?
 WHERE \`device_id\` = ?
   AND \`wallet_id\` = ?
```

tests/pushUpdate.test.ts
```diff
@@ -0,0 +1,231 @@
...
+      enablePush: false,
+      enableShowAmounts: false,
+    });
+
+    const result = await update(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(400);
+    expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('invalid-payload');
```
andreabadesso: Import and use `ApiError.INVALID_PAYLOAD`

tests/pushUpdate.test.ts
```diff
@@ -0,0 +1,231 @@
...
+      enablePush: false,
+      enableShowAmounts: false,
+    });
+
+    const result = await update(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(404);
+    expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('device-not-found');
```
andreabadesso: Import and use `ApiError.DEVICE_NOT_FOUND`

andreabadesso (changes requested):

andreabadesso (approved):

PR #303, opened 2022-11-11
feat: create push/unregister lambda

alexruzenhack:
### Acceptance Criteria
- create push/unregister lambda

### Security Checklist

### DIFF
--- a/src/api/pushUnregister.ts
+++ b/src/api/pushUnregister.ts
@@ -24,11 +24,10 @@ class PushUpdateUnregisterValidator {
   });
 
   static validate(payload: unknown): { value: PushDelete, error: ValidationError } {
-    const { value, error } = PushUpdateUnregisterValidator.bodySchema.validate(payload, {
+    return PushUpdateUnregisterValidator.bodySchema.validate(payload, {
       abortEarly: false, // We want it to return all the errors not only the first
       convert: true, // We need to convert as parameters are sent on the QueryString
-    });
-    return { value: { enablePush: false, enableShowAmounts: false, ...value }, error };
+    }) as { value: PushDelete, error: ValidationError };
   }
 }
 
--- a/src/api/pushUnregister.ts
+++ b/src/api/pushUnregister.ts
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { APIGatewayProxyHandler } from 'aws-lambda';
+import { ApiError } from '@src/api/errors';
+import { closeDbAndGetError, warmupMiddleware } from '@src/api/utils';
+import { unregisterPushDevice } from '@src/db';
+import { getDbConnection } from '@src/utils';
+import { walletIdProxyHandler } from '@src/commons';
+import middy from '@middy/core';
+import cors from '@middy/http-cors';
+import Joi from 'joi';
+import { PushDelete } from '@src/types';
+
+const mysql = getDbConnection();
+
+const bodySchema = Joi.object({
+  deviceId: Joi.string().max(256).required(),
+});
+
+/*
+ * Unregister a device to recive push notification.
+ *
+ * This lambda is called by API Gateway on POST /push/unregister
+ */
+export const unregister: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (walletId, event) => {
+  const { value, error } = bodySchema.validate(event.body, {
+    abortEarly: false, // We want it to return all the errors not only the first
+    convert: true, // We need to convert as parameters are sent on the QueryString
+  });
+
+  // TODO: validate with tulio if the solution can be that or must follow strictly the design.
+  if (error) {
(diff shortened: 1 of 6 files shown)


src/db/index.ts
```diff
@@ -2753,3 +2753,24 @@ export const updatePushDevice = async (
     [input.enablePush, input.enableShowAmounts, input.deviceId, input.walletId],
   );
 };
+
+/**
+ * Unregister push device for a given wallet.
+ *
+ * @param mysql - Database connection
+ * @param input - Input of push device register
```
pedroferreira1: We don't have this parameter

tests/db.test.ts
```diff
@@ -2299,3 +2300,33 @@ test('removeAllPushDeviceByDeviceId', async () => {
...
+
+  await expect(checkPushDevicesTable(mysql, 1, {
+    walletId,
+    deviceId,
+    pushProvider,
+    enablePush,
+    enableShowAmounts,
+  })).resolves.toBe(true);
+
+  await unregisterPushDevice(mysql, deviceId, walletId);
```
pedroferreira1: Maybe add another `checkPushDevicesTable` and expect it to be `false` to validate it has unregistered the device?

src/api/pushUnregister.ts
```diff
@@ -0,0 +1,54 @@
...
+  const body: PushDelete = value;
+  await unregisterPushDevice(mysql, body.deviceId, walletId);
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ success: true }),
+  };
+}))
+  .use(cors())
+  .use(warmupMiddleware());
```
tuliomir: Same question about the walletWarmer that is defined on the `serverless.yml` as disabled, but added here to the handler.
Should we have this?

tests/db.test.ts
```diff
@@ -2299,3 +2300,33 @@ test('removeAllPushDeviceByDeviceId', async () => {
   await removeAllPushDeviceByDeviceId(mysql, deviceId);
   await expect(checkPushDevicesTable(mysql, 0)).resolves.toBe(true);
 });
+
+test('unregisterPushDevice', async () => {
```
tuliomir: I feel we should also have tests for:
- the behavior when there is no device to unregister
- assert that, when multiple devices are registered for the same wallet, only the correct device is removed

tests/pushUnregister.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+beforeEach(async () => {
+  await cleanDatabase(mysql);
+});
+
+afterAll(async () => {
+  await closeDbConnection(mysql);
+});
+
+test('update push device given a wallet', async () => {
```
tuliomir:
```suggestion
test('unregister push device for a given wallet', async () => {
```

tests/pushUnregister.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    pushProvider,
+    enablePush: true,
+    enableShowAmounts,
+  })).resolves.toBe(true);
+
+  const event = makeGatewayEventWithAuthorizer(walletId, null, {
+    deviceId,
+  });
+
+  const result = await unregister(event, null, null) as APIGatewayProxyResult;
```
tuliomir, line 65: While not necessary, I feel we should have a test that tries to delete a device that was never registered, to ensure its result will be a success.

tuliomir (approved):

src/api/pushUnregister.ts
```diff
@@ -0,0 +1,59 @@
...
+  static readonly bodySchema = Joi.object({
+    deviceId: Joi.string().max(256).required(),
+  });
+
+  static validate(payload: unknown): { value: PushDelete, error: ValidationError } {
+    const { value, error } = PushUpdateUnregisterValidator.bodySchema.validate(payload, {
+      abortEarly: false, // We want it to return all the errors not only the first
+      convert: true, // We need to convert as parameters are sent on the QueryString
+    });
+    return { value: { enablePush: false, enableShowAmounts: false, ...value }, error };
```
pedroferreira1: What are those `enablePush: false, enableShowAmounts: false` values? Do we use them here?

andreabadesso: Apparently they are default values

@alexruzenhack You can use `Joi` to set default values, like `.default(true)`

src/db/index.ts
```diff
@@ -2750,6 +2750,26 @@ export const updatePushDevice = async (
...
+ */
+export const unregisterPushDevice = async (
+  mysql: ServerlessMysql,
+  deviceId: string,
+  walletId: string,
+) : Promise<void> => {
+  await mysql.query(
+    `
+    DELETE FROM \`push_devices\`
+    WHERE device_id = ? AND wallet_id = ?`,
```
```suggestion
    DELETE
      FROM \`push_devices\`
     WHERE \`device_id\` = ?
       AND \`wallet_id\` = ?`,
```

andreabadesso (changes requested):

pedroferreira1 (approved):

andreabadesso (approved):

andreabadesso (approved):

PR #305, opened 2022-11-12
feat: check wallet addresses api

andreabadesso:
### Acceptance Criteria
- We should create a new lambda to check if a list of addresses belongs to the wallet calling the lambda

### Security Checklist

### DIFF
--- a/src/api/addresses.ts
+++ b/src/api/addresses.ts
@@ -38,6 +38,15 @@ const checkMineBodySchema = Joi.object({
  */
 export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (walletId, event) => {
   const status = await getWallet(mysql, walletId);
+
+  // If the wallet is not started or ready, we can skip the query on the address table
+  if (!status) {
+    return closeDbAndGetError(mysql, ApiError.WALLET_NOT_FOUND);
+  }
+  if (!status.readyAt) {
+    return closeDbAndGetError(mysql, ApiError.WALLET_NOT_READY);
+  }
+
   const eventBody = (function parseBody(body) {
     try {
       return JSON.parse(body);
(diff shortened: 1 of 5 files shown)


src/api/addresses.ts
```diff
@@ -7,20 +7,89 @@
...
+    return closeDbAndGetError(mysql, ApiError.WALLET_NOT_READY);
+  }
+
+  const eventBody = (function parseBody(body) {
+    try {
+      return JSON.parse(body);
+    } catch (e) {
+      return null;
+    }
+  }(event.body));
```
alexruzenhack, line 58: This conversion is needed? I'm not applying it to my lambda implementations. Should I?

alexruzenhack (approved):

src/api/addresses.ts
andreabadesso: It is! Even with `convert: true`, Joi is not accepting a string (that comes from the event)

It's weird that it is working on your PRs

src/api/addresses.ts
```diff
@@ -7,20 +7,90 @@
...
 import { closeDbConnection, getDbConnection } from '@src/utils';
 import { walletIdProxyHandler } from '@src/commons';
 import middy from '@middy/core';
 import cors from '@middy/http-cors';
 
 const mysql = getDbConnection();
 
+const checkMineBodySchema = Joi.object({
+  addresses: Joi.array()
+    .items(Joi.string().regex(/^[A-HJ-NP-Za-km-z1-9]*$/).min(34).max(34))
```
tuliomir, line 29: I haven't seen this rule for addresses anywhere in our documentation or code. Researching a bit I found that this is the [regex for bitcoin addresses](https://stackoverflow.com/a/24205650/1123007), which we also use here.

I feel some documentation and regex constant should be added about this on our lib, even if it just points out to the original bitcoin definitions. That way we could use something like
```js
Joi.string().regex(libConstants.addressRegex).min(34).max(34)
```

src/api/addresses.ts
tuliomir: Thinking about this, maybe this problem only arises when the first deploy is made to the dev environment.
Up until then, all local tests can be made using a JSON object instead of its stringified version.

tests/api.test.ts
```diff
@@ -153,6 +153,99 @@ test('GET /addresses', async () => {
...
+    maxGap: 5,
+    createdAt: 10000,
+    readyAt: 10001,
+  }]);
+
+  await addToAddressTable(mysql, [
+    { address: ADDRESSES[0], index: 0, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[1], index: 1, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[2], index: 3, walletId: 'my-wallet', transactions: 0 },
+    { address: ADDRESSES[3], index: 3, walletId: 'my-wallet', transactions: 0 },
```
tuliomir: Why are we overwriting index `3` here?

andreabadesso: Fixed! Thanks

src/api/addresses.ts
andreabadesso: Added a comment on 37ae52825a8b0957f3e17b448deb74e61b8b9cae to explain the regex

tuliomir (approved):

PR #309, opened 2022-11-14
feat: create lambda wallet/getTxById

alexruzenhack:
### Acceptance Criteria
- create lambda wallet/getTxById

### Security Checklist


serverless.yml
```diff
@@ -478,6 +478,17 @@ functions:
...
+        enabled: false
+  getTxById:
+    handler: src/api/txById.get
+    events:
+      - http:
+          path: wallet/getTxById
+          method: get
+          cors: true
+          authorizer: ${self:custom.authorizer.walletBearer}
     warmup:
```
tuliomir, line 503: Maybe this route could have the warmup active. I think it's worth at least studying the benefits and the costs of it.

src/db/index.ts
```diff
@@ -2774,3 +2775,30 @@ export const unregisterPushDevice = async (
...
+ * @param mysql - Database connection
+ * @param txId - A transaction ID
+
+ * @returns A transaction if found, return null otherwise
+ */
+export const getTransactionById = async (
+  mysql: ServerlessMysql,
+  txId: string,
+): Promise<Tx|null> => {
+  const result: DbSelectResult = await mysql.query(
```
tuliomir: The query here must be more complex to be able to retrieve the transaction's balances for the current wallet as well. 

As proposed by the design, it should have a similar interface to the [`getWalletTxHistory` method](https://github.com/HathorNetwork/hathor-wallet-service/blob/master/src/db/index.ts#L1331-L1384), but fetching every relevant token for this specific wallet.

tests/txById.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const result = await get(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(400);
+    expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('invalid-payload');
+  });
+});
+
+describe('statusCode:404', () => {
```
tuliomir, line 109: I suggest having a `statusCode:500` test as well, mocking the method calls made internally and ensuring a treated response is delivered to the middleware.

tuliomir (changes requested):

tests/txById.test.ts
alexruzenhack: Do we have a global error handler? Because beyond this scenario I don't see a way to go wrong, except for connection errors and things like that. Unknown errors should be globally caught imo.

alexruzenhack: Also, tests are good, but not too much. I agree with this quote ["Write tests. Not too many. Mostly integration."](https://twitter.com/rauchg/status/807626710350839808?s=20&t=dI2uUted0o5yBWErR9Zbgg)

src/db/index.ts
pedroferreira1: We also need the check if this tx belongs to the `walletId` from the request, otherwise it's 404

serverless.yml
alexruzenhack: I believe @andreabadesso can give us a hand about it

andreabadesso: We are currently only warming up the lambdas called during the wallet load so we don't impact the perceived loading time

I would suggest not warming it up to avoid unnecessary lambda calls (the warmup process is a lambda call every 5 minutes)

src/db/index.ts
```diff
@@ -2774,3 +2771,72 @@ export const unregisterPushDevice = async (
...
+    SELECT
+      transaction.tx_id AS tx_id,
+      transaction.timestamp AS timestamp,
+      transaction.version AS version,
+      transaction.voided AS voided,
+      transaction.height AS height,
+      transaction.weight AS weight,
+      wallet_tx_history.balance AS balance,
+      wallet_tx_history.token_id AS token_id,
+      wallet_tx_history.wallet_id AS wallet_id
```
tuliomir: Do we need this column on the results? We have filtered by a single `wallet_id` already.

src/db/index.ts
```diff
@@ -2774,3 +2771,72 @@ export const unregisterPushDevice = async (
...
+    FROM wallet_tx_history
+    INNER JOIN transaction ON transaction.tx_id = wallet_tx_history.tx_id
```
tuliomir: Taking a double look at the front-end design and checking this query here it seems the design did not consider the client's need for the tokens' names and symbols on the response.

On the front-end, this will be the desired results:
Front end exhibition

To achieve this, we will need to add two new columns here, otherwise the client application will have nowhere to fetch these data from:
```sql
SELECT
  ...
  token.name AS token_name
  token.symbol AS token_symbol
FROM ...
INNER JOIN token ON wallet_tx_history.token_id = token.id
WHERE ...
```

tests/db.test.ts
```diff
@@ -2289,44 +2354,185 @@ test('removeAllPushDeviceByDeviceId', async () => {
...
+      tokenId: token2,
+      txId: txId1,
+      version: version1,
+      voided: false,
+      walletId: walletId1,
+      weight: weight1,
+    });
+  });
+
+  it('should return null when there is no record', async () => {
```
tuliomir: it actually returns an empty array, not `null`, right?

tests/txById.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  const event = makeGatewayEventWithAuthorizer(walletId1, null, {
+    txId: txId1,
+  });
+
+  const result = await get(event, null, null) as APIGatewayProxyResult;
+  const returnBody = JSON.parse(result.body as string);
+
+  expect(result.statusCode).toStrictEqual(200);
+  expect(returnBody.success).toStrictEqual(true);
+  expect(returnBody.tx).toHaveLength(2);
```
tuliomir: The property `tx` seems to indicate we have two transactions on the response, and could be misleading.

Maybe we could rename it to something that makes it clear that this property is related to a list of token balances? Two possibilities I see are:
- Changing the `tx` property to something like `{ txId: string, tokens: [] }`
- Just changing the `tx` property name to `txTokens`, as received by the db util function

What do you think?

tuliomir (commented):
The current state of the PR has more code than what is related to the `getTxById` method.

I think it's a good idea to separate it to isolate the review only on the PR's main subject.

src/api/pushUnregister.ts
```diff
@@ -7,31 +7,38 @@
...
+  static readonly bodySchema = Joi.object({
+    deviceId: Joi.string().max(256).required(),
+  });
+
+  static validate(payload: unknown): { value: PushDelete, error: ValidationError } {
+    const { value, error } = PushUpdateUnregisterValidator.bodySchema.validate(payload, {
+      abortEarly: false, // We want it to return all the errors not only the first
+      convert: true, // We need to convert as parameters are sent on the QueryString
+    });
+    return { value: { enablePush: false, enableShowAmounts: false, ...value }, error };
```
andreabadesso: Where do `enablePush` and `enableShowAmounts` come from? They are not in the `bodySchema`

Should they be constants? I'm a little confused on what this is doing

src/db/index.ts
```suggestion
    SELECT transaction.tx_id AS tx_id,
           transaction.timestamp AS timestamp,
           transaction.version AS version,
           transaction.voided AS voided,
           transaction.height AS height,
           transaction.weight AS weight,
           wallet_tx_history.balance AS balance,
           wallet_tx_history.token_id AS token_id,
           wallet_tx_history.wallet_id AS wallet_id
      FROM wallet_tx_history
INNER JOIN transaction ON transaction.tx_id = wallet_tx_history.tx_id

```

src/db/index.ts
```diff
@@ -2774,3 +2771,72 @@ export const unregisterPushDevice = async (
...
+* @param mysql - Database connection
+* @param walletId - The wallet linked to device
+*/
+export const existsWallet = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+) : Promise<boolean> => {
+  const [{ count }] = await mysql.query(
+    `
+    SELECT COUNT(1) as \`count\` FROM \`wallet\` pd WHERE id= ?`,
```
```suggestion
SELECT COUNT(1) as \`count\`
  FROM \`wallet\` pd
 WHERE id = ?
```

tests/db.test.ts
```diff
@@ -2241,45 +2243,108 @@ test('registerPushDevice', async () => {
...
+    await expect(checkPushDevicesTable(mysql, 0)).resolves.toBe(true);
+  });
 });
 
 test('removeAllPushDeviceByDeviceId', async () => {
   expect.hasAssertions();
 
   const walletId = 'wallet1';
-  const deviceId = 'device1';
+  // eslint-disable-next-line @typescript-eslint/naming-convention, camelcase
```
andreabadesso: Why don't you just call it `deviceId1` or `deviceIdOne`?

tests/pushRegister.test.ts
```diff
@@ -73,6 +182,7 @@ describe('statusCode:400', () => {
 
     expect(result.statusCode).toStrictEqual(400);
     expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('invalid-payload');
```
andreabadesso: Import and use `ApiError.INVALID_PAYLOAD`

tests/pushRegister.test.ts
```diff
@@ -101,5 +211,26 @@ describe('statusCode:400', () => {
...
```
andreabadesso: Same here

tests/pushRegister.test.ts
```diff
@@ -101,5 +211,26 @@ describe('statusCode:400', () => {
...
+      enablePush: true,
+      enableShowAmounts: false,
+    });
+
+    const result = await register(event, null, null) as APIGatewayProxyResult;
+    const returnBody = JSON.parse(result.body as string);
+
+    expect(result.statusCode).toStrictEqual(404);
+    expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('wallet-not-found');
```
andreabadesso: Import and use `ApiError.WALLET_NOT_FOUND`

tests/pushUpdate.test.ts
```diff
@@ -96,6 +196,7 @@ describe('statusCode:400', () => {
 
     expect(result.statusCode).toStrictEqual(400);
     expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('invalid-payload');
```
andreabadesso: Same here

tests/pushUpdate.test.ts
```diff
@@ -125,5 +226,6 @@ describe('statusCode:404', () => {
 
     expect(result.statusCode).toStrictEqual(404);
     expect(returnBody.success).toStrictEqual(false);
+    expect(returnBody.error).toStrictEqual('device-not-found');
```
andreabadesso: Same here

src/db/index.ts
andreabadesso: Screenshot 2022-11-23 at 02 09 04

Wow, this is annoying me more than it should. Is this happening on the production app?

serverless.yml
```diff
@@ -478,6 +478,17 @@ functions:
           method: delete
           cors: true
           authorizer: ${self:custom.authorizer.walletBearer}
+    warmup:
+      walletWarmer:
+        enabled: false
+  getTxById:
```
andreabadesso, line 495: We should add the `authorizer` key here to only allow querying if the user is authenticated

andreabadesso (changes requested):

andreabadesso:
Can you link the design on this PR? Took me a while to find it as it is on the internal-issues project:

https://github.com/HathorNetwork/internal-issues/issues/124

src/db/index.ts
alexruzenhack: Well, I have followed the walletTxHistory. But you right.

tests/txById.test.ts
alexruzenhack: I chose `txTokens` to represent the list, this is a good name indeed.

serverless.yml
alexruzenhack: How do I do it? Can you provide an example?

andreabadesso (approved):

src/db/index.ts
```diff
@@ -2785,30 +2802,33 @@ export const unregisterPushDevice = async (
  * @param mysql - Database connection
  * @param txId - A transaction ID
  * @param walletId - The wallet related to the transaction
- * @returns A transaction if found, return null otherwise
+ * @returns A a list of tokens for a transaction if found, return an empty list otherwise
```
tuliomir:
```suggestion
 * @returns A list of tokens for a transaction if found, return an empty list otherwise
```

tests/db.test.ts
tuliomir: Should we change this test's title?

tuliomir (approved):

src/api/pushSendNotificationToDevice.ts
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { Handler } from 'aws-lambda';
+import { closeDbConnection, getDbConnection } from '@src/utils';
+import Joi, { ValidationError } from 'joi';
+import { SendNotificationToDevice, PushProvider } from '@src/types';
+import { getPushDevice } from '@src/db';
```
tuliomir:
```suggestion
import { getPushDevice, unregisterPushDevice } from '@src/db';
```

andreabadesso (approved):

tuliomir (approved):

PR #311, opened 2022-11-22
feat: create sendToFcm util

alexruzenhack:
### Acceptance Criteria
- create sendToFcm util - send push notification to Android
- handle invalid device id - unregister invalid device id

### Security Checklist


src/types.ts
```diff
@@ -717,7 +717,7 @@ export interface SendNotificationToDevice {
   deviceId: string,
   title: string,
   description: string,
-  metadata: Record<string, unknown>,
+  metadata: Record<string, string>,
```
tuliomir, line 720: Here we're enforcing we will only have a 1-level-deep object as metadata.

It serves all of our needs, but since we're restricting the contents I believe a simple comment would help future developers identify this limitation to avoid any errors.

src/utils/pushnotification.utils.ts
```diff
@@ -1,5 +1,8 @@
 import { Lambda } from 'aws-sdk';
 import { SendNotificationToDevice } from '@src/types';
+import { credential, initializeApp, messaging, ServiceAccount } from 'firebase-admin';
+import { MulticastMessage } from 'firebase-admin/messaging';
+import serviceAccount from '@src/utils/fcm.config.json';
```
tuliomir: Since we're adding a new file outside our repository, I think it deserves its own section on the `readme.md` project file.

Also, a verification to check if the file is not available and a custom error / explicit warning on this context would help developers troubleshoot any problems they might encounter because of this feature.

src/utils/pushnotification.utils.ts
```diff
@@ -14,15 +17,40 @@ if (!process.env.STAGE) {
...
+    if (multicastResult.failureCount === 0) {
+      return { success: true };
+    }
+
+    const { 0: { error } } = multicastResult.responses;
+    if (/token-not-registered/.test(error?.code || '')) {
+      return { success: false, errorMessage: PushNotificationError.INVALID_DEVICE_ID };
+    }
+
+    return { success: false, errorMessage: PushNotificationError.UNKNOWN };
```
tuliomir, line 65: If we're going to hide the actual message from the rest of the application, I think we should log it here.

andreabadesso (approved):

src/utils/pushnotification.utils.ts
```diff
@@ -2,21 +2,32 @@ import { Lambda } from 'aws-sdk';
 import { SendNotificationToDevice } from '@src/types';
 import { credential, initializeApp, messaging, ServiceAccount } from 'firebase-admin';
 import { MulticastMessage } from 'firebase-admin/messaging';
-import serviceAccount from '@src/utils/fcm.config.json';
 import createDefaultLogger from '@src/logger';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const serviceAccount = require('@src/utils/fcm.config.json');
```
tuliomir, line 8: I understand this is the simplest way to load a JSON into the application, but it also triggers some static code rules.

I think we can keep it this way and later, if necessary, convert this to a full `fsPromises.readFile` + `JSON.parse` call.

tests/db.test.ts
```diff
@@ -2644,11 +2646,217 @@ describe('getPushDevice', () => {
...
+
+    // get settings querying only devices not loaded on database, resulting on empty list
+    const notRegisteredWalletIdList = devicesToNotLoad.map((each) => each.walletId);
+    const result = await getPushDeviceSettingsList(mysql, notRegisteredWalletIdList);
+
+    // assert settings
+    expect(result).toStrictEqual([]);
+  });
+
+  it('should return a list of seetings even when some wallet ids are not found', async () => {
```
tuliomir:
```suggestion
  it('should return a list of settings even when some wallet ids are not found', async () => {
```

package.json
```diff
@@ -7,7 +7,7 @@
     "jest": "jest --runInBand --collectCoverage --detectOpenHandles --forceExit",
     "lint": "eslint src/ tests/ --ext .js,.jsx,.ts,.tsx src tests",
     "lint-fix": "eslint src/ tests/ --fix --ext .js,.jsx,.ts,.tsx src tests",
-    "check-types": "tsc --noemit",
+    "check-types": "tsc --noemit --skipLibCheck",
```
tuliomir, line 10: Why was that necessary?

I was expecting it to have some relation to the `bitcore-lib` that we usually have conflicts with, but found no issue with it.

alexruzenhack: It will avoid running type checks on the libs in the node_modules path.

tuliomir (approved):

PR #323, opened 2022-12-07
feat: implement db.routine to delete stale push devices

alexruzenhack:
### Acceptance Criteria
- lambda triggered by cron-job to clean up stale push devices.

### Security Checklist


serverless.yml
```diff
@@ -510,6 +510,13 @@ functions:
         enabled: false
   txPushNotificationRequested:
     handler: src/api/txPushNotificationRequested.handleRequest
+    warmup:
+      walletWarmer:
+        enabled: false
+  dbroutineDeleteStalePushDevices:
+    handler: src/db.routine.cleanStalePushDevices
+    events:
+      - schedule: cron(17 3 */15 * * *)
```
tuliomir: A comment with the interval written here would be helpful too

src/db.routines.ts
```diff
@@ -0,0 +1,39 @@
+/**
```
tuliomir: What do you think about moving this file to the `db/` folder?

src/db/index.ts
```diff
@@ -2945,3 +2945,33 @@ export const getPushDeviceSettingsList = async (
...
+ *
+ * @param mysql - Database connection
+ * @returns - total of stale device from now
+ */
+export const countStalePushDevices = async (mysql): Promise<number> => {
+  const [staleDevice] = await mysql.query(
+    `
+    SELECT COUNT(device_id) as count
+      FROM \`push_devices\`
+     WHERE updated_at > UNIX_TIMESTAMP(date_sub(now(), interval 1 month))`,
```
tuliomir, line 2960: Not that it's really necessary, but I think it would be nice to have a constant for all queries that may want to interact with potentially stale devices, should they grow beyond those two.

What do you think of:
```js
const SQL_STALE_DEVICE_TIMESTAMP = 'UNIX_TIMESTAMP(date_sub(now(), interval 1 month))'
```

and then on the queries:
```mysql
SET stale_timestamp = (?)

SELECT
  # ...
  WHERE updated_at > stale_timestamp
```

tuliomir (approved):

src/db.routines.ts
andreabadesso: Or maybe something like `cron/`, `schedules/`

src/db/index.ts
```diff
@@ -2945,3 +2945,33 @@ export const getPushDeviceSettingsList = async (
...
+ *
+ * @param mysql - Database connection
+ */
+export const deleteStalePushDevices = async (mysql) => {
+  await mysql.query(
+    `
+    DELETE
+      FROM \`push_devices\`
+      WHERE updated_at > UNIX_TIMESTAMP(date_sub(now(), interval 1 month))`,
+  );
```
```suggestion
    DELETE
      FROM \`push_devices\`
     WHERE updated_at > UNIX_TIMESTAMP(date_sub(now(), interval 1 month))`,
```

andreabadesso (changes requested):

src/db/index.ts
alexruzenhack: What is the use case you have in mind?

src/db.routines.ts
alexruzenhack: Imo the folder `db/` is better to keep consistency, but I add `cron` as a prefix in the file name.

src/db/index.ts
tuliomir: Nothing specific, just thinking on the DRY principle

alexruzenhack: I see. Well, let these cases come naturally. We don't know if we will really need it in the future.

andreabadesso (approved):

PR #326, opened 2022-12-08
chore: add snapshots to error details

alexruzenhack:
### Acceptance Criteria
- Add inline snpatshot to validation error details in push register, update and unregister tests

### Security Checklist

### DIFF
--- a/tests/pushRegister.test.ts
+++ b/tests/pushRegister.test.ts
@@ -184,6 +184,16 @@ describe('statusCode:400', () => {
     expect(result.statusCode).toStrictEqual(400);
     expect(returnBody.success).toStrictEqual(false);
     expect(returnBody.error).toStrictEqual(ApiError.INVALID_PAYLOAD);
+    expect(returnBody.details).toMatchInlineSnapshot(`
+Array [
+  Object {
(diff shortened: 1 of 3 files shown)


tuliomir (approved):
I like the additional validations, but wouldn't they be easier to read/maintain if we used something like `expect().toStrictlyEqual()` and passed the object in plain javascript instead of serialized? Ex.:

```js
expect(returnBody.details).toStrictlyEqual([
  {
    message: '"pushProvider" with value (...)',
    path: [ 'pushProvider' ],
  }
]);
```

alexruzenhack:
> I like the additional validations, but wouldn't they be easier to read/maintain if we used something like expect().toStrictlyEqual() and passed the object in plain javascript instead of serialized?

More readable yes, but not easier to maintain. The inline snapshot is a great tool to maintain a check over data that have more potential to change. To update it we jest need to run jest with `-u` flag.

andreabadesso (approved):
Didn't know about `toMatchInlineSnapshot`, nice!

We could refactor this in many other test files, can you open an issue to do that later?

alexruzenhack:
> Didn't know about `toMatchInlineSnapshot`, nice!
> 
> We could refactor this in many other test files, can you open an issue to do that later?

Sure, here it is: https://github.com/HathorNetwork/hathor-wallet-service/issues/330

PR #327, opened 2022-12-09
feat: refactor push/unregister to receive deviceId by path param

alexruzenhack:
### Acceptance Criteria
- change the endpoint to receive `deviceId` through path param

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -485,10 +485,14 @@ functions:
     handler: src/api/pushUnregister.unregister
     events:
       - http:
-          path: push/unregister
-          method: post
+          path: push/unregister/{deviceId}
+          method: delete
           cors: true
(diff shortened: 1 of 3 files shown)


tuliomir (approved):

andreabadesso:
Where is this coming from? Are we having problems with `DELETE` in any of our wallets?

I don't like the idea of using `POST` to delete a resource, `DELETE` makes much more sense, doesn't it?

alexruzenhack:
> I don't like the idea of using POST to delete a resource, DELETE makes much more sense, doesn't it?

Yes, it does. I did this change to make it work. But you are right, I can refactor to keep the DELETE verb but receive a path param. I will work on it then.

andreabadesso (approved):

PR #329, opened 2022-12-11
feat: refactor getTxById to receive txId by path param

alexruzenhack:
### Acceptance Criteria
- pass txId by path param, it will simplify the endpoint usage and comply with HTTP REST spec
- change route to wallet/transactions/{txId}

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -496,7 +496,7 @@ functions:
     handler: src/api/txById.get
     events:
       - http:
-          path: wallet/txById/{txId}
+          path: wallet/transactions/{txId}
           method: get
           cors: true
           authorizer: ${self:custom.authorizer.walletBearer}
--- a/serverless.yml
+++ b/serverless.yml
@@ -496,7 +496,7 @@ functions:
     handler: src/api/txById.get
     events:
       - http:
-          path: wallet/getTxById/{txId}
+          path: wallet/txById/{txId}
           method: get
           cors: true
           authorizer: ${self:custom.authorizer.walletBearer}
--- a/serverless.yml
+++ b/serverless.yml
@@ -496,10 +496,14 @@ functions:
     handler: src/api/txById.get
     events:
       - http:
-          path: wallet/getTxById
+          path: wallet/getTxById/{txId}
           method: get
           cors: true
           authorizer: ${self:custom.authorizer.walletBearer}
+          request:
+            parameters:
+              paths:
+                txId: true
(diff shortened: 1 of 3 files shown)


serverless.yml
```diff
@@ -496,10 +496,14 @@ functions:
     handler: src/api/txById.get
     events:
       - http:
-          path: wallet/getTxById
+          path: wallet/txById/{txId}
```
andreabadesso: What do you think about refactoring it to be something line:

`wallet/transactions/{txId}`

I think it looks a lot more `restful`

andreabadesso (changes requested):

serverless.yml
alexruzenhack: @andreabadesso, you're right. I didn't use `/transactions` because I was afraid to use it, but it is indeed better.

andreabadesso (approved):

serverless.yml
andreabadesso: @alexruzenhack thanks!

Don't forget to update the acceptance criteria

src/api/txById.ts
```diff
@@ -20,7 +20,7 @@ const mysql = getDbConnection();
 
 class TxByIdValidator {
   static readonly bodySchema = Joi.object({
-    txId: Joi.string().required(),
+    txId: Joi.string().min(1).required(),
```
andreabadesso: I guess we can use `.min(32).max(32)` here since `txId` will always be a `sha256` hash

andreabadesso (changes requested):

src/api/txById.ts
alexruzenhack: Are you sure it is 32 characters? Doesn't it 64?

andreabadesso: Oops, it's in hexadecimal string format, so yes, 64 bytes

Sorry

src/api/txById.ts
```diff
@@ -20,7 +20,7 @@ const mysql = getDbConnection();
 
 class TxByIdValidator {
   static readonly bodySchema = Joi.object({
-    txId: Joi.string().required(),
+    txId: Joi.string().min(32).max(32).required(),
```
tuliomir: From previous conversations I see that this should be `min(64).max(64)`

tuliomir (approved):

andreabadesso (approved):

PR #340, opened 2022-12-28
fix: pushRegister and pushUpdate

alexruzenhack:
### Acceptance Criteria
- Add JSON parser to event.body on pushRegister and pushUpdate lambdas, because API Gateway stringfy the payload
- Add wallet/ path as prefix for push/ paths, because auth guard has the policy to allow only wallet/ and tx/ paths

### Security Checklist

### DIFF
--- a/src/api/pushRegister.ts
+++ b/src/api/pushRegister.ts
@@ -40,7 +40,15 @@ class PushRegisterInputValidator {
  * This lambda is called by API Gateway on POST /push/register
  */
 export const register: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (walletId, event) => {
-  const { value: body, error } = PushRegisterInputValidator.validate(event.body);
+  const eventBody = (function parseBody(body) {
+    try {
+      return JSON.parse(body);
+    } catch (e) {
+      return null;
+    }
+  }(event.body));
+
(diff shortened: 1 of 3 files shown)


andreabadesso (approved):

src/api/pushRegister.ts
```diff
@@ -40,7 +40,15 @@ class PushRegisterInputValidator {
  * This lambda is called by API Gateway on POST /push/register
  */
 export const register: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (walletId, event) => {
-  const { value: body, error } = PushRegisterInputValidator.validate(event.body);
+  const eventBody = (function parseBody(body) {
```
tuliomir, line 43: I can see that almost every API handler has to do this very same code to work with the way AWS Lambda sends the events.

I suggest opening an _issue_ to refactor all of them into an util function, so that future implementations can easily reuse them and avoid rework by not parsing the JSON on the first time.

tuliomir (approved):

src/api/pushRegister.ts
alexruzenhack: Yes, it could be a middleware to deliver a JSON in the body

PR #360, opened 2023-01-04
chore: add test to logger.error on firebase initialization

alexruzenhack:
### Acceptance Criteria
- add test to logger.error on firebase initialization in order to increase coverage

### Security Checklist

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -5,7 +5,8 @@ on:
     paths-ignore:
       - '**.md'
   push:
-    branch: [master, dev]
+    tags: ['v**']
+    branches: [master, dev]
 jobs:
   build:
     runs-on: ubuntu-20.04
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -1,7 +1,6 @@
 name: PR Validation
 on:
   pull_request:
-    types: [opened, synchronize]
     paths-ignore:
       - '**.md'
   push:
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -1,11 +1,11 @@
 name: PR Validation
 on:
   pull_request:
-    branches: [master, dev]
     paths-ignore:
       - '**.md'
   push:
-    branches: [master, dev]
+    paths-ignore:
+      - '**.md'
 jobs:
   build:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


luislhl:
I think we have a problem on `https://github.com/HathorNetwork/hathor-wallet-service/blob/master/.github/workflows/pr-validation.yml`

You can see that in this PR, the tests didn't run yet.

This is because the config in the workflow is preventing them from running for the first commit of each PR.

We should create another PR to fix this.

The fix is to remove these lines:
```
    types: [ready_for_review, synchronize]
    branches: [master, dev]
```

The `types` config is what causes the problem. We should just use default (by removing this line).

The `branches` config, if kept, will prevent PRs that have other bases from running tests as well.
I think it doesn't make sense.

.github/workflows/pr-validation.yml
```diff
@@ -1,11 +1,11 @@
 name: PR Validation
 on:
   pull_request:
-    types: [ready_for_review, synchronize]
-    branches: [master, dev]
+    types: [opened, synchronize]
```
luislhl: The github actions default is `[opened, synchronize, reopened]`

Did you leave `reopened` out for a specific reason?
It seems that we should have it.

But I don't think this is a big deal, so I'll not block the PR for this

alexruzenhack: Oh, I didn't know about the default. I will ripe it off then.

tests/utils/pushnotification.utils.test.ts
```diff
@@ -37,6 +37,25 @@ describe('PushNotificationUtils', () => {
...
   });
 
+  // test firebase initialization error
+  it('firebase initialization error', () => {
+    expect.hasAssertions();
+
+    // load local env
+    process.env.PUSH_NOTIFICATION_ENABLED = 'true';
+    logger.error.mockReset();
+    initFirebaseAdminMock.mockReset();
```
luislhl, line 47: Maybe it's better to add these 2 `mockReset` calls to the `afterEach` function as well, so we make sure we won't be leaving traces that could impact other tests. 

What do you think?

Unless there is some other process in place somewhere to reset all mocks.

luislhl (approved):

tests/utils/pushnotification.utils.test.ts
alexruzenhack: This is a good point, but I don't like to generalize with few applications, in this case only this test needs this reset.

andreabadesso (approved):

PR #361, opened 2023-01-05
feat: added a proxy API to fetch a transaction from the fullnode

andreabadesso:
### Acceptance Criteria
- We should have "proxied" methods for querying the fullnode
- We should have the same global rate limits as we have on our public fullnodes

## On using lambdas instead of ApiGateway directly

We could setup the proxy directly on `ApiGateway`, removing the need for the proxy lambdas (which are currently only making the request and returning the response transparently

That would avoid entirely the lambda calls and decrease our costs

The problem with that approach is that it is currently not supported by the serverless syntax, there is a [serverless plugin](https://www.serverless.com/plugins/serverless-apigateway-service-proxy) to configure proxy integrations on the ApiGateway but it does not support [HTTP integrations](https://docs.aws.amazon.com/apigateway/latest/developerguide/setup-http-integrations.html), which would be our usecase

I could try to implement it using the `CloudFormation` syntax on the serverless file, but it would take some time since I'm not quite used to it and would need some time to study it.

Keeping that and the fact that this API is needed to finish the QA on the wallet-desktop and deploy the wallet-service facade for internal testing, I would suggest deploying the current solution using lambdas and create an issue to refactor this to use `ApiGateway` directly

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -573,6 +573,8 @@ functions:
             parameters:
               paths:
                 txId: true
+          throttling:
+            maxRequestsPerSecond: 50
     warmup:
       walletWarmer:
         enabled: false
@@ -588,6 +590,8 @@ functions:
             parameters:
               paths:
                 txId: true
+          throttling:
+            maxRequestsPerSecond: 10
     warmup:
       walletWarmer:
         enabled: false
@@ -599,6 +603,8 @@ functions:
           method: get
           cors: true
           authorizer: ${self:custom.authorizer.walletBearer}
+          throttling:
+            maxRequestsPerSecond: 20
     warmup:
       walletWarmer:
         enabled: false
--- a/serverless.yml
+++ b/serverless.yml
@@ -420,6 +420,17 @@ functions:
     warmup:
       walletWarmer:
         enabled: false
+  proxyGetTxById:
+    handler: src/api/fullnodeProxy.getTransactionById
+    events:
+      - http:
+        path: wallet/proxy/transactions/{txId}
+        method: get
+        cors: true
+        authorizer: ${self:custom.authorizer.walletBearer}
+    warmup:
+      walletWarmer:
+        enabled: false
   wsConnect:
     handler: src/ws/connection.connect
     timeout: 1
--- a/serverless.yml
+++ b/serverless.yml
@@ -420,17 +420,6 @@ functions:
     warmup:
       walletWarmer:
         enabled: false
-  proxyGetTxById:
-    handler: src/api/fullnodeProxy.getTransactionById
-    events:
-      - http:
-        path: wallet/proxy/transactions/{txId}
-        method: get
-        cors: true
-        authorizer: ${self:custom.authorizer.walletBearer}
-    warmup:
-      walletWarmer:
-        enabled: false
   wsConnect:
     handler: src/ws/connection.connect
     timeout: 1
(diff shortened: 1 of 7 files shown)


serverless.yml
```diff
@@ -561,6 +561,53 @@ functions:
...
+          path: wallet/proxy/transactions/{txId}/confirmation_data
+          method: get
+          cors: true
+          authorizer: ${self:custom.authorizer.walletBearer}
+          request:
+            parameters:
+              paths:
+                txId: true
+          throttling:
+            maxRequestsPerSecond: 10
```
pedroferreira1, line 594: Why 50 in the first and 10 in this one?

src/api/fullnodeProxy.ts
```diff
@@ -0,0 +1,141 @@
...
+  });
+
+  if (error) {
+    const details = error.details.map((err) => ({
+      message: err.message,
+      path: err.path,
+    }));
+
+    return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
+  }
```
pedroferreira1: These lines here are exactly the same for all the methods (except for the validator, that can be a parameter). I feel we should refactor it. What do you think?

serverless.yml
andreabadesso: I copied the public fullnode's configuration, it's different for each path

src/api/fullnodeProxy.ts
andreabadesso: I agree, added an utils method to help with validation in b92552775732b92ffcb2246be428dede78f804b0

Eventually, we should refactor all other APIs to use it.

src/api/utils.ts
```diff
@@ -92,3 +101,33 @@ export const pushProviderRegexPattern = (): RegExp => {
   const options = entries.join('|');
   return new RegExp(`^(?:${options})$`);
 };
+
+export const validateParams = (
+  validator: Schema,
+  params: unknown,
```
pedroferreira1, line 107: What's the difference between `any` and `unknown`?

src/api/utils.ts
```diff
@@ -92,3 +101,33 @@ export const pushProviderRegexPattern = (): RegExp => {
...
+  validatorOptions: ValidationOptions = {
+    abortEarly: false,
+    convert: false,
+  },
+): ParamValidationResult => {
+  const result: ValidationResult = validator.validate(params, validatorOptions);
+
+  const { error, value } = result;
+
+  if (result.error) {
```
pedroferreira1:
```suggestion
  if (error) {
```

pedroferreira1 (approved):

src/api/fullnodeProxy.ts
```diff
@@ -0,0 +1,137 @@
...
+});
+
+/*
+ * Get a transaction from the fullnode
+ *
+ * This lambda is called by API Gateway on GET /wallet/proxy/transactions/:id
+ */
+export const getTransactionById: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (_walletId: string, event) => {
+  const params = event.pathParameters || {};
+  const validationResult: ParamValidationResult = validateParams(txIdValidator, params);
```
tuliomir: Since we'll need to cast the value results later everytime, it would be nice to be able to set them here, like `ParamValidationResult<GetTxByIdParams>`

src/api/utils.ts
tuliomir: I've read a good number of answers to this question, and [this one](https://stackoverflow.com/a/51439876/1123007) seems to me the most concise.

In short: `unknown` forces the developer to check the contents of the property at runtime. `any` just turns off the type checking.

src/fullnode.ts
```diff
@@ -31,7 +31,35 @@ export const create = (baseURL = BASE_URL): any => {
...
+
+    return response.data;
+  };
+
+  return {
+    api, // exported so we can mock it on the tests
+    downloadTx,
+    getConfirmationData,
+    queryGraphvizNeighbors,
+  };
```
tuliomir, line 62: Since we're explicitly returning the method's contents, there should be no need for the `: any` declaration on line 18, that is raising issues with the lint.

tuliomir (approved):

src/api/fullnodeProxy.ts
andreabadesso: Way better, thank you!

96a3ea9329bb88ed0a9aed698bc5b13166522786

src/api/utils.ts
andreabadesso: 06142eaa0c86e0a94ea10841f90d5f16caee08a5

PR #372, opened 2023-01-24
fix: push notification

alexruzenhack:
### Acceptance Criteria
- chore: add permissions to invoke push lambdas
- chore: rename lambda to txPushRequested
- fix: lambda sendNotificationToDevice
- fix: lambda txPushRequested
- feat: improve pushnotification.utils.js
- feat: add push notification feature toggle in the onNewTxRequest lambda

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,7 @@
     "source-map-support": "^0.5.19",
     "tiny-secp256k1": "^2.2.1",
     "uuid": "^8.3.0",
-    "winston": "^3.7.2",
-    "bitcore-lib": "8.25.31"
+    "winston": "^3.7.2"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.95",
@@ -75,6 +74,7 @@
     "typescript": "^4.9.3",
     "typescript-eslint": "0.0.1-alpha.0",
     "webpack": "^4.29.0",
-    "webpack-node-externals": "^1.7.2"
+    "webpack-node-externals": "^1.7.2",
+    "bitcore-lib": "8.25.31"
   }
 }
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "source-map-support": "^0.5.19",
     "tiny-secp256k1": "^2.2.1",
     "uuid": "^8.3.0",
-    "winston": "^3.7.2"
+    "winston": "^3.7.2",
+    "bitcore-lib": "8.25.31"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.95",
@@ -49,7 +50,6 @@
     "@types/node": "^18.0.4",
     "@typescript-eslint/eslint-plugin": "^3.3.0",
     "@typescript-eslint/parser": "^3.3.0",
-    "bitcore-lib": "8.25.31",
     "dotenv": "^10.0.0",
     "eslint": "^7.27.0",
     "eslint-config-airbnb-base": "^14.2.1",
--- a/src/utils/pushnotification.utils.ts
+++ b/src/utils/pushnotification.utils.ts
@@ -50,10 +50,6 @@ if (!process.env.FIREBASE_CLIENT_X509_CERT_URL) {
   logger.error('[ALERT] env.FIREBASE_CLIENT_X509_CERT_URL can not be null or undefined.');
 }
 
-if (!process.env.PUSH_ALLOWED_PROVIDERS) {
-  logger.error('[ALERT] env.PUSH_ALLOWED_PROVIDERS can not be null or undefined.');
-}
-
 export function buildFunctionName(functionName: string): string {
   return `hathor-wallet-service-${process.env.STAGE}-${functionName}`;
 }
--- a/src/utils/pushnotification.utils.ts
+++ b/src/utils/pushnotification.utils.ts
@@ -80,7 +80,7 @@ const PUSH_NOTIFICATION_ENABLED = process.env.PUSH_NOTIFICATION_ENABLED;
  * PUSH_ALLOWED_PROVIDERS=android,ios
  * @remarks
  * In the test this constant works like the environment variable constants.
- * It need to be reloaded after change the underlying environment variable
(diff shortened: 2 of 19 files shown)


package.json
```diff
@@ -41,15 +41,15 @@
     "source-map-support": "^0.5.19",
     "tiny-secp256k1": "^2.2.1",
     "uuid": "^8.3.0",
-    "winston": "^3.7.2"
+    "winston": "^3.7.2",
+    "bitcore-lib": "8.25.31"
```
andreabadesso: Why is this listed as a dependency? We only use it on the tests

tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "lib": ["es2017"],
     "removeComments": true,
     "moduleResolution": "node",
+    "module": "commonjs",
```
andreabadesso, line 6: What is this for?

andreabadesso (changes requested):

tsconfig.json
alexruzenhack: This is to enable dynamic imports using the ecma6 import syntax, like this: https://github.com/HathorNetwork/hathor-wallet-service/pull/372/files#diff-d79c36e7f5deb2d97946b3313d23d929ef33fb0790d7fa03602a9e44e3a05b5aR247

package.json
alexruzenhack: I did it to trying to resolve the bitcore-lib issue. I have restored it to dev dependencies now.

tests/pushSendNotificationToDevice.test.ts
```diff
@@ -25,16 +26,37 @@ initFirebaseAdminMock();
...
+    PUSH_ALLOWED_PROVIDERS: 'android,ios',
+  };
   spyOnSendToFcm.mockClear();
   spyOnLoggerError.mockClear();
   await cleanDatabase(mysql);
+  jest.resetModules();
+});
+
+afterEach(() => {
+  process.env = initEnv;
```
tuliomir: Since each test will overwrite the `process.env` anyway, maybe it would be better to move this environment-reverting code to `afterAll`, wouldn't it?

tests/txPushNotificationRequested.test.ts
```diff
@@ -26,6 +74,58 @@ afterAll(async () => {
 });
 
 describe('success', () => {
+  it('should alert when invoke send notification fails', async () => {
+    expect.hasAssertions();
+
+    const walletId = 'wallet1';
+    await addToWalletTable(mysql, [buildWallet({ id: walletId })]);
+
+    // device with disabled enableShowAmounts, resulting in a generic notification
```
tuliomir: This comments leads the reader to think there will be a consequence to the `enableShowAmounts` property here, but the test ignores this and just tests an error handling, right?

If this is correct, maybe just removing this comment would be better for the reader.

src/utils/pushnotification.utils.ts
```diff
@@ -74,6 +74,17 @@ const FIREBASE_AUTH_PROVIDER_X509_CERT_URL = process.env.FIREBASE_AUTH_PROVIDER_
 const FIREBASE_CLIENT_X509_CERT_URL = process.env.FIREBASE_CLIENT_X509_CERT_URL;
 /** Local feature toggle that disable the push notification by default */
 const PUSH_NOTIFICATION_ENABLED = process.env.PUSH_NOTIFICATION_ENABLED;
+const PUSH_ALLOWED_PROVIDERS = (() => {
```
tuliomir, line 92: The format this data is stored caused some weird bits of code on the testing, especially the need to `await import()` in the middle of the code, which for me indicates a testing code smell ( even though it works just fine! ).

I suggest one of two paths here, since this PR is already quite big and the feature has chained ongoing PR's:
- Changing `PUSH_ALLOWED_PROVIDERS` to a function `getPushAllowedProviders()` that fetch the most up-to-date environment variables
- Adding an explanatory JSDocs here to this constant, making it clear that it only stores the environment values once and, should they change on testing, another import is needed to update them ( with a reference to the already existing test code )

If deciding for a code change only later, an issue could be open to keep track of this as well.

tests/pushSendNotificationToDevice.test.ts
alexruzenhack: Hummm... you right!

src/utils/pushnotification.utils.ts
alexruzenhack: I did it to avoid calling `process.env` during the lambda's handler processing. Otherwise, this would be another blocking call. Besides that, constants are good to optimize lambda execution, after all the constants are executed in the INIT phase and can be reutilized in warmed instances.

The way to test it is the same for the `env` variables. But sure, I will improve the docstring. However, given it is better to keep it as a constant, should I open an issue?

tuliomir: No need for issues, just the helper comments will be fine :+1:

src/utils/pushnotification.utils.ts
```diff
@@ -74,6 +74,21 @@ const FIREBASE_AUTH_PROVIDER_X509_CERT_URL = process.env.FIREBASE_AUTH_PROVIDER_
 const FIREBASE_CLIENT_X509_CERT_URL = process.env.FIREBASE_CLIENT_X509_CERT_URL;
 /** Local feature toggle that disable the push notification by default */
 const PUSH_NOTIFICATION_ENABLED = process.env.PUSH_NOTIFICATION_ENABLED;
+/**
+ * Controls which providers are allowed to send notification when it is enabled
+ * @example
+ * PUSH_ALLOWED_PROVIDERS=android,ios
+ * @remarks
+ * In the test this constant works like the environment variable constants.
+ * It need to be reloaded after change the underlying environment variable
```
tuliomir:
```suggestion
 * It needs to be reloaded after changing the underlying environment variable
```

tuliomir (approved):

andreabadesso (approved):

PR #374, opened 2023-01-24
chore: using mysql2 as library for serverless-mysql

andreabadesso:
## Motivation

The last update of the `mysql` library we were using was in Jan 23, 2020

### Acceptance Criteria
- We should use `mysql2` as the library for serverless-mysql (mysql is the default)
- We should remove the mysql dependency

### Security Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "hathor-wallet-service",
-      "version": "1.23.0-beta",
+      "version": "1.23.1-beta",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -383,7 +383,7 @@ export const initWalletTxHistory = async (mysql: ServerlessMysql, walletId: stri
   const results: DbSelectResult = await mysql.query(
     `SELECT \`tx_id\`,
             \`token_id\`,
-            SUM(\`balance\`) AS balance,
+            CAST(SUM(\`balance\`) AS INTEGER) AS balance,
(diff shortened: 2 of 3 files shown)


src/db/index.ts
```diff
@@ -383,7 +383,7 @@ export const initWalletTxHistory = async (mysql: ServerlessMysql, walletId: stri
   const results: DbSelectResult = await mysql.query(
     `SELECT \`tx_id\`,
             \`token_id\`,
-            SUM(\`balance\`) AS balance,
+            CAST(SUM(\`balance\`) AS INTEGER) AS balance,
```
andreabadesso, line 386: `mysql2` is returning `SUM`, `MIN` and `MAX` as strings, so we can either cast it on mysql itself or on javascript, with `parseInt`, I prefer doing it directly in mysql

package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "hathor-wallet-service",
-      "version": "1.23.0-beta",
+      "version": "1.23.1-beta",
```
andreabadesso, line 9: The last bump didn't update this, totally unrelated to the PR

src/db/index.ts
r4mmer: Casting to integer is fine but the `SUM` returns a `DECIMAL` (which the client converts to a string) because of a posible overflow and precision issues. Maybe we should check no cast has this problem, like a sum of output values that can go over the mysql native integer highest number.

r4mmer (commented):
There are some warnings like "Argument 'mysql' should be typed" which can be fixed if we use the mysql2 types

PR #375, opened 2023-01-24
chore: increased ws timeouts to 2s

andreabadesso:
## Motivation

We are receiving timeout alerts for the websocket lambdas on wallet-service

Upon investigation, I've detected that the timeouts are currently configured to 1s

It does make sense to have a low timeout value for those APIs as they should run in under 50ms per call

The problem is that those lambdas need to query the MySQL database to fetch the walletId from the requester

The timeout is happening because the connection to the database (which happens only once in a while for multiple lambda calls) takes up to 1s because of the TLS handshake.

Since the lambda always times out, the connection never gets re-used, causing every lambda call to create a new MySQL connection and timeout because of the TLS handshake time.

This is not affecting the wallets because the timeout happens after the client gets a response, so that's why it took us a while to receive this error

### Acceptance Criteria
- We should increase the timeouts on the websocket lambdas to fix https://github.com/HathorNetwork/on-call-incidents/issues/87
- We should create alarms if the average of duration of websocket lambdas is greater than 1s

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -68,6 +68,17 @@ custom:
         treatMissingData: notBreaching
         alarmActions:
           - minor
+      wsLambdasExecutionDuration:
+        description: "WebSocket lambda took too long to execute"
+        namespace: 'AWS/Lambda'
+        metric: "Duration"
+        statistic: Average
+        threshold: 1000 # This is in milliseconds
+        period: 60
+        evaluationPeriods: 1
+        comparisonOperator: GreaterThanThreshold
+        alarmActions:
+          - minor
     alarms: # Alarms that will be applied to all functions
       - majorFunctionErrors
       - minorFunctionErrors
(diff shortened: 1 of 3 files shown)


serverless.yml
```diff
@@ -437,18 +448,22 @@ functions:
     warmup:
       walletWarmer:
         enabled: false
+    alarms:
+      - wsLambdasExecutionDuration
```
luislhl, line 452: Maybe add a comment to make it clear that this gets merged with the global alarms?

luislhl (approved):

serverless.yml
andreabadesso: Done in a3fa311d756faca71800fd6c0b0ea8eb954368df

Thanks!

tuliomir (approved):

PR #377, opened 2023-01-27
fix: throw error when htr is requested on getTokenDetails

andreabadesso:
## Motivation

Since we now have htr on our token table, we should ignore it on `getTokenDetails` to prevent unnecessary database queries when it is requested

Fixes https://github.com/HathorNetwork/on-call-incidents/issues/87

### Acceptance Criteria
- Return an error when HTR is requested

### Security Checklist

### DIFF
--- a/src/api/tokens.ts
+++ b/src/api/tokens.ts
@@ -13,7 +13,7 @@ import {
 } from '@src/types';
 import { getDbConnection } from '@src/utils';
 import { ApiError } from '@src/api/errors';
-import { closeDbAndGetError, warmupMiddleware } from '@src/api/utils';
+import { closeDbAndGetError, warmupMiddleware, txIdJoiValidator } from '@src/api/utils';
 import Joi from 'joi';
 import { constants } from '@hathor/wallet-lib';
 import middy from '@middy/core';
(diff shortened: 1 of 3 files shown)


luislhl (approved):

src/api/tokens.ts
```diff
@@ -70,6 +70,14 @@ export const getTokenDetails = middy(walletIdProxyHandler(async (walletId, event
   const tokenId = value.token_id;
   const tokenInfo: TokenInfo = await getTokenInformation(mysql, tokenId);
 
+  if (tokenId === constants.HATHOR_TOKEN_CONFIG.uid) {
```
tuliomir, line 71: Another low-cost possibility here would be checking `if tokenId.length < 64` before making any queries elsewhere.

That way any actual invalid token would be rejected promptly without having to query any external service to reach this same conclusion.

tuliomir (approved):

src/api/tokens.ts
msbrogli: It seems something useful for other parts of this project. Maybe create a method `isValidTxId(...)` so you can reuse it later.

andreabadesso: @msbrogli @tuliomir refactored in e129fb741dbd064b8c896d1864279a484a31ce58

andreabadesso: @msbrogli @tuliomir refactored in e129fb741dbd064b8c896d1864279a484a31ce58

tuliomir (approved):

PR #378, opened 2023-01-30
refactor: alert using sqs

andreabadesso:
### Acceptance Criteria
- We should send alerts directly to SQS instead of just logging with an `ALERT` string

### Security Checklist

### DIFF
--- a/src/api/pushSendNotificationToDevice.ts
+++ b/src/api/pushSendNotificationToDevice.ts
@@ -8,10 +8,10 @@
 import { Handler } from 'aws-lambda';
 import { closeDbConnection, getDbConnection } from '@src/utils';
 import Joi, { ValidationError } from 'joi';
-import { Severity, SendNotificationToDevice } from '@src/types';
+import { Severity, SendNotificationToDevice, PushProvider } from '@src/types';
 import { getPushDevice, unregisterPushDevice } from '@src/db';
 import createDefaultLogger from '@src/logger';
-import { PushProvider, PushNotificationUtils, PushNotificationError } from '@src/utils/pushnotification.utils';
+import { PushNotificationUtils, PushNotificationError } from '@src/utils/pushnotification.utils';
 import { addAlert } from '@src/utils/alerting.utils';
 
 const mysql = getDbConnection();
--- a/src/api/pushSendNotificationToDevice.ts
+++ b/src/api/pushSendNotificationToDevice.ts
@@ -8,10 +8,11 @@
 import { Handler } from 'aws-lambda';
 import { closeDbConnection, getDbConnection } from '@src/utils';
 import Joi, { ValidationError } from 'joi';
-import { SendNotificationToDevice } from '@src/types';
+import { Severity, SendNotificationToDevice } from '@src/types';
 import { getPushDevice, unregisterPushDevice } from '@src/db';
 import createDefaultLogger from '@src/logger';
-import { PushNotificationUtils, PushNotificationError, isPushProviderAllowed } from '@src/utils/pushnotification.utils';
+import { PushProvider, PushNotificationUtils, PushNotificationError } from '@src/utils/pushnotification.utils';
+import { addAlert } from '@src/utils/alerting.utils';
 
 const mysql = getDbConnection();
 
@@ -63,15 +64,27 @@ export const send: Handler<unknown, { success: boolean, message?: string, detail
 
   if (!pushDevice) {
     closeDbConnection(mysql);
-    logger.error('[ALERT] Device not found.', {
+    await addAlert(
+      'Device not found while trying to send notification',
+      '-',
+      Severity.MINOR,
+      { deviceId: body.deviceId },
+    );
+    logger.error('Device not found.', {
       deviceId: body.deviceId,
     });
     return { success: false, message: 'Failed due to device not found.' };
   }
 
-  if (!isPushProviderAllowed(pushDevice.pushProvider)) {
+  if (pushDevice.pushProvider !== PushProvider.ANDROID) {
     closeDbConnection(mysql);
-    logger.error('[ALERT] Provider invalid.', {
+    await addAlert(
+      'Invalid provider error while sending push notification',
+      '-',
+      Severity.MINOR,
+      { deviceId: body.deviceId, pushProvider: pushDevice.pushProvider },
+    );
+    logger.error('Provider invalid.', {
       deviceId: body.deviceId,
       pushProvider: pushDevice.pushProvider,
     });
--- a/src/api/pushSendNotificationToDevice.ts
+++ b/src/api/pushSendNotificationToDevice.ts
@@ -8,10 +8,10 @@
 import { Handler } from 'aws-lambda';
 import { closeDbConnection, getDbConnection } from '@src/utils';
 import Joi, { ValidationError } from 'joi';
-import { Severity, SendNotificationToDevice, PushProvider } from '@src/types';
+import { Severity, SendNotificationToDevice } from '@src/types';
 import { getPushDevice, unregisterPushDevice } from '@src/db';
 import createDefaultLogger from '@src/logger';
-import { PushNotificationUtils, PushNotificationError } from '@src/utils/pushnotification.utils';
+import { isPushProviderAllowed, PushNotificationUtils, PushNotificationError } from '@src/utils/pushnotification.utils';
 import { addAlert } from '@src/utils/alerting.utils';
 
 const mysql = getDbConnection();
@@ -76,7 +76,7 @@ export const send: Handler<unknown, { success: boolean, message?: string, detail
     return { success: false, message: 'Failed due to device not found.' };
   }
 
-  if (pushDevice.pushProvider !== PushProvider.ANDROID) {
+  if (!isPushProviderAllowed(pushDevice.pushProvider)) {
     closeDbConnection(mysql);
     await addAlert(
       'Invalid provider error while sending push notification',
--- a/src/commons.ts
+++ b/src/commons.ts
@@ -57,8 +57,10 @@ import {
   WalletBalance,
   Transaction,
   WalletBalanceValue,
+  Severity,
 } from '@src/types';
 import { Logger } from 'winston';
+import { addAlert } from '@src/utils/alerting.utils';
 
 import {
   getUnixTimestamp,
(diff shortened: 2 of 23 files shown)


src/api/pushSendNotificationToDevice.ts
```diff
@@ -63,15 +64,27 @@ export const send: Handler<unknown, { success: boolean, message?: string, detail
...
+      { deviceId: body.deviceId },
+    );
+    logger.error('Device not found.', {
       deviceId: body.deviceId,
     });
     return { success: false, message: 'Failed due to device not found.' };
   }
 
-  if (!isPushProviderAllowed(pushDevice.pushProvider)) {
+  if (isPushProviderAllowed(pushDevice.pushProvider)) {
```
alexruzenhack: 🚫 Here the condition should have a negative `!`. Only log the error if the provider is not allowed.
```javascript
if (!isPushProviderAllowed(pushDevice.pushProvider)) {
```

src/types.ts
```diff
@@ -789,3 +789,12 @@ export interface WalletBalanceValue {
   addresses: string[],
   walletBalanceForTx: TokenBalanceValue[],
 }
+
+export enum Severity {
```
alexruzenhack, line 797: 🚲 I believe it would be nice to have JSDoc here explaining this severity is related to the alert system, because the severity definition for the log may be different.

src/utils.ts
```diff
@@ -357,3 +357,22 @@ export const getAddressFromXpub = (xpubkey: string): string => {
...
   }).address;
 };
+
+/**
+ * Validates if a list of env variables are set in the environment. Throw if at least
+ * one of them is missing
+ *
+ * @param envVariables - A list of variables to check
+ */
+export const validateEnvVariables = (envVariables: string[]): void => {
```
alexruzenhack: 🚲 A validation function implies a boolean return, which is not the case here. Should the name be changed to convey a better semantic? I suggest `assertEnvVariablesExistence`, but this is a `bike`.

src/utils/alerting.utils.ts
```diff
@@ -0,0 +1,53 @@
...
+        DataType: 'String',
+        StringValue: '--',
+      },
+    },
+  };
+
+  sqs.sendMessage(params, (err) => {
+    if (err) {
+      logger.error('[ALERT] Erroed while sending message to the alert sqs queue');
+      logger.error(err);
```
alexruzenhack: 💭 Why use 2 error logs? Why not only 1:
```javascript
logger.error('[ALERT] Erroed while sending message to the alert sqs queue', err);
```

src/utils/pushnotification.utils.ts
```diff
@@ -233,6 +227,12 @@ export class PushNotificationUtils {
     // Event InvocationType returns 202 for a successful invokation
     const walletIdList = Object.keys(walletBalanceValueMap);
     if (response.StatusCode !== 202) {
+      await addAlert(
+        'Error on PushNotificationUtils',
+        `${SEND_NOTIFICATION_FUNCTION_NAME} lambda invoke failed for wallets`,
```
alexruzenhack: 🚫 Here should be `ON_TX_PUSH_NOTIFICATION_REQUESTED_FUNCTION_NAME` in the place of `SEND_NOTIFICATION_FUNCTION_NAME`

src/ws/connection.ts
```diff
@@ -20,32 +20,47 @@ import {
...
+    await closeDbConnection(mysql);
+  } catch (e) {
+    await addAlert(
+      'Captured error on connect websocket lambda',
+      '-',
+      Severity.MINOR,
+      { error: e.message },
+    );
+
+    logger.error(e);
```
alexruzenhack: 💭 Should the error log also have an explanation?
```javascript
logger.error('Captured error on connect websocket lambda', e);
```

tests/utils/alerting.utils.mock.ts
```diff
@@ -0,0 +1,4 @@
+export const mockedAddAlert = jest.fn();
+export default jest.mock('@src/utils/alerting.utils', () => ({
+  addAlert: mockedAddAlert.mockReturnValue(Promise.resolve()),
+}));
```
alexruzenhack, line 4: ⭐ Nice!

tests/utils/pushnotification.utils.test.ts
```diff
@@ -69,7 +70,11 @@ describe('PushNotificationUtils', () => {
       // reload module
       const { PushNotificationUtils } = await import('@src/utils/pushnotification.utils');
 
-      expect(logger.error).toHaveBeenLastCalledWith('[ALERT] env.WALLET_SERVICE_LAMBDA_ENDPOINT can not be null or undefined.');
+      expect(mockedAddAlert).toHaveBeenLastCalledWith(
```
alexruzenhack, line 74: 🚲 It is working, but it would be nice to have the `mockedAddAlert` mock reset `beforeEach`, [like in the line `36`](https://github.com/HathorNetwork/hathor-wallet-service/pull/378/files#diff-da2fa7c63ab1c8c03df04018ca79ce81c497c3c20bf41cea89659689dd2684d7R36).

alexruzenhack (changes requested):
There are only 2 required changes, some bikes you may want to address and other thoughts.

src/utils.ts
alexruzenhack: ⭐ But nice function!

andreabadesso: I agree with the name change, done in fba51be743dd63d6e336f684c081b0649d15985e

src/utils/alerting.utils.ts
andreabadesso: Not sure, refactored in 0bd59b087878707537898b48d65b8d12db224bf9

src/utils/pushnotification.utils.ts
andreabadesso: Thanks, done in 230285c7b9395145474a8a9f45942ff64c5dffcb

src/ws/connection.ts
andreabadesso: Done in 45d242d26ad848c1501fcc565a393d0afee809a2

tests/utils/pushnotification.utils.test.ts
andreabadesso: Done in b5c0046872637f2215e0ef24fbcab0e12cbdae46

src/api/pushSendNotificationToDevice.ts
andreabadesso: Wow, not sure what happened here, fixed in 86821c0a

src/commons.ts
```diff
@@ -515,7 +517,13 @@ export const handleReorg = async (mysql: ServerlessMysql, logger: Logger): Promi
   });
 
   if ((currentHeight - height) > WARN_MAX_REORG_SIZE) {
-    logger.error(`[ALERT] A reorg with ${currentHeight - height} blocks has been detected`);
+    logger.error(`A reorg with ${currentHeight - height} blocks has been detected`);
+    await addAlert(
+      'Big reorg detected',
+      `A reorg with ${currentHeight - height} blocks has been detected`,
+      Severity.MINOR,
```
tuliomir, line 524: Should we have multiple alerts with different severities for an increasing amount of blocks being _re-org_'d? A reorg of 2 blocks being `minor`, but a reorg of 20 blocks being `major` or something.

The place for this kind of implementation is not in a refactor PR, but I wanted to bring up the subject on the code here.

src/types.ts
tuliomir: Agreed. A link to the document containing the most up-to-date definitions (if available) would be a great addition.

src/utils/pushnotification.utils.ts
tuliomir: I'm confused: I'm reviewing a version more recent than this `230285c` commit and the value is back to `SEND_NOTIFICATION_FUNCTION_NAME`. No commit seems to have reverted it.

Can you please double check this has the correct value now?

src/utils/alerting.utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  metadata?: unknown,
+): Promise<void> => {
+  const logger = createDefaultLogger();
+  const preparedMessage = {
+    title,
+    message,
+    severity,
+    metadata,
+    environment: process.env.NETWORK,
+    application: process.env.APPLICATION_NAME,
```
tuliomir, line 37: I would suggest adding a call to `assertEnvVariablesExistence()` to ensure these variables are present on startup, so that we don't send insufficient data to the alert manager.

src/utils/pushnotification.utils.ts
```diff
@@ -203,6 +191,12 @@ export class PushNotificationUtils {
 
     // Event InvocationType returns 202 for a successful invokation
     if (response.StatusCode !== 202) {
+      await addAlert(
+        'Error on PushNotificationUtils',
+        `${ON_TX_PUSH_NOTIFICATION_REQUESTED_FUNCTION_NAME} lambda invoke failed for device: ${notification.deviceId}`,
```
alexruzenhack: ⛔ Here should be `SEND_NOTIFICATION_FUNCTION_NAME`

src/utils/pushnotification.utils.ts
alexruzenhack: I've realized the problem. The commit above changes the constant in the wrong place 😅, where the commit made the change was correct.

The `invokeSendNotificationHandlerLambda` function must have the following:
- `SEND_NOTIFICATION_FUNCTION_NAME`

The `invokeOnTxPushNotificationRequestedLambda` function must have the following:
- `ON_TX_PUSH_NOTIFICATION_REQUESTED_FUNCTION_NAME`

src/utils/pushnotification.utils.ts
alexruzenhack: See the comment that follows.

alexruzenhack (changes requested):
Still have a required change to make. These little things to change are the devil.

src/commons.ts
andreabadesso: I guess we can, I don't remember and couldn't find where we discussed this current setting

I'll create an issue so we track and discuss this

src/types.ts
andreabadesso: I agree, done in 516a2b4dc58ebc3d84de73ec107c2c886c9c6b80

src/utils/pushnotification.utils.ts
andreabadesso: Oops! Fixed in 071b8d52104138a5defee2a1c17b9157b2b920ac

tuliomir (approved):

alexruzenhack (approved):

PR #387, opened 2023-02-06
chore: add manual deploy for dev-testnet environment

alexruzenhack:
### Acceptance Criteria
- add manual deploy for dev-testnet environment

This addition in the CD pipeline aims to test a delivery passing through all the CI/CD pipeline.

### Security Checklist

### DIFF
--- a/.codebuild/buildspec.yml
+++ b/.codebuild/buildspec.yml
@@ -146,17 +146,20 @@ phases:
           for var in "${!testnet_@}"; do
             export ${var#testnet_}="${!var}"
           done
-
           make migrate;
           make deploy-lambdas-testnet;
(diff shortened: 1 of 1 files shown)


luislhl:
With the proposed changes, we will have something like:
```
if (...)
[...]

if (...)
[...]
elif (...)
[...]
else
[...]
```

We need to join the first if to the `if..elif..else` chain as well, otherwise the `dev` branch will be always deployed 2 times.

Deleting the first `if` would have the same result.

luislhl (changes requested):
See my previous comment

luislhl (approved):

andreabadesso (approved):

PR #402, opened 2023-02-28
chore: missing env variables

andreabadesso:
### Acceptance Criteria
- Refactored the `LOG_LEVEL` to default to `debug`
- Refactored the `addAlert` method to create the `QUEUE_URL` from env variables (`ACCOUNT_ID`, `ALERT_MANAGER_REGION`, `ALERT_MANAGER_TOPIC`)

Note: After review is done, we need to add the new envs to the secret manager on our AWS account

### Security Checklist


src/utils/alerting.utils.ts
```diff
@@ -50,9 +59,13 @@ export const addAlert = async (
     },
   };
 
-  sqs.sendMessage(params, (err) => {
-    if (err) {
-      logger.error('[ALERT] Erroed while sending message to the alert sqs queue', err);
-    }
+  await new Promise<void>((resolve) => {
```
andreabadesso, line 62: During tests I noticed that sometimes the lambda would end causing the alert to never be sent

tuliomir: I suggest adding a comment to explain this promise on the code, just like you did here on the PR.

That way, future developers will understand why not to simplify the code by removing the seemingly unnecessary promise.

tuliomir (approved):

.github/workflows/pr-validation.yml
```diff
@@ -81,6 +81,9 @@ jobs:
           EXPLORER_SERVICE_LAMBDA_ENDPOINT: https://lambda.eu-central-1.amazonaws.com
           WALLET_SERVICE_LAMBDA_ENDPOINT: https://lambda.eu-central-1.amazonaws.com
           APPLICATION_NAME: wallet-service-test
+          ACCOUNT_ID: 1
```
alexruzenhack, line 84: I couldn't find the `ACCOUNT_ID` reference on `.codebuild/buildspec.yml`, or on `serverless.yml`.

alexruzenhack, line 84: I couldn't find the `ACCOUNT_ID` reference on `.codebuild/buildspec.yml`, or on `serverless.yml`.

.github/workflows/pr-validation.yml
```diff
@@ -81,6 +81,9 @@ jobs:
...
+          ALERT_MANAGER_REGION: unknown-region-1
+          ALERT_MANAGER_TOPIC: alert-manager-void
```
alexruzenhack, line 86: Can we add this to `.env.example`?

alexruzenhack (changes requested):
Laking `ACCOUNT_ID` declaration on `buildspec.yml` and `serverless.yml`

alexruzenhack (approved):

.github/workflows/pr-validation.yml
alexruzenhack: It is already declared.

src/utils/alerting.utils.ts
andreabadesso: Actually, thinking more about this I think this is kind of implicit, it was an error before for the `addAlert` to be async but leak this callback after it resolves, right?

Adding a comment there would be like:

"This async method is really async because we are returning a promise"

.github/workflows/pr-validation.yml
andreabadesso: We didn't have one, but added in 3828dff6340d798119364aae45c970cffff6b9ed

PR #406, opened 2023-04-02
fix: using the latest block timestamp to detect if we have voided mem…

andreabadesso:
Fixes https://github.com/HathorNetwork/on-call-incidents/issues/102

## Explanation

Currently, to detect transactions that got voided before getting confirmed by a block, we run a cronjob every 20 minutes to check if we have transactions without `height` on the database, if we do, we query the fullnode to check if it's really voided.

Since we are having larger block times because of the volatile hashrate on our mainnet, transactions are taking more than 20 minutes to get confirmed by a block and this is triggering alerts (false positives).

I changed it to check all transactions 20 minutes older than the latest block timestamp.

### Acceptance Criteria
- We should use the best block timestamp to query the database for mempool voided txs

### Security Checklist

### DIFF
--- a/src/mempool.ts
+++ b/src/mempool.ts
@@ -7,8 +7,7 @@
 
 import 'source-map-support/register';
 import {
-  getBlockByHeight,
-  getLatestHeight,
+  getLatestBlockByHeight,
   getMempoolTransactionsBeforeDate,
   updateTx,
 } from '@src/db';
@@ -19,7 +18,6 @@ import {
   fetchBlockHeight,
   closeDbConnection,
   getDbConnection,
-  getUnixTimestamp,
 } from '@src/utils';
 import createDefaultLogger from '@src/logger';
 import { addAlert } from '@src/utils/alerting.utils';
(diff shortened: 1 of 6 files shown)


src/mempool.ts
```diff
@@ -36,12 +38,15 @@ export const onHandleOldVoidedTxs = async (): Promise<void> => {
   const logger = createDefaultLogger();
 
   const VOIDED_TX_OFFSET: number = parseInt(process.env.VOIDED_TX_OFFSET, 10) * 60; // env is in minutes
-  const now: number = getUnixTimestamp();
-  const date: number = now - VOIDED_TX_OFFSET;
+  const latestHeight: number = await getLatestHeight(mysql);
+  const bestBlock: Block = await getBlockByHeight(mysql, latestHeight);
```
luislhl: Should we create a `getLatestBlockByHeight` to avoid 2 queries in the database?

luislhl (approved):

glevco (approved):

src/mempool.ts
andreabadesso: Good idea, done in cf9f855622dee8dea4a162ea52a378dba08e1b8b

luislhl (approved):

PR #407, opened 2023-04-03
chore: support new mainnet-staging environment

luislhl:
### Acceptance Criteria
- Remove old uneeded variables from all environments: `ADMINTOKEN`
- Add new env vars and new logic to support the mainnet-staging environment on the `buildspec` file. We want it to be deployed for release-candidate tags like `v1.2.3-rc.1`

### Security Checklist


alexruzenhack (approved):

.codebuild/buildspec.yml
```diff
@@ -35,6 +35,17 @@ env:
...
     testnet_APPLICATION_NAME: "wallet-service-testnet"
+    mainnet_staging_DEFAULT_SERVER: "https://wallet-service.private-nodes.hathor.network/v1a/"
+    mainnet_staging_WS_DOMAIN: "ws.staging.wallet-service.hathor.network"
+    mainnet_staging_NETWORK: "mainnet"
+    mainnet_staging_LOG_LEVEL: "debug"
+    mainnet_staging_NFT_AUTO_REVIEW_ENABLED: "false"
+    mainnet_staging_EXPLORER_STAGE: "mainnet-staging"  # This stage doesn't exist in explorer-service and we don't need it because we have disabled the integration.
+    mainnet_staging_EXPLORER_SERVICE_LAMBDA_ENDPOINT: "https://lambda.eu-central-1.amazonaws.com"
+    mainnet_staging_WALLET_SERVICE_LAMBDA_ENDPOINT: "https://lambda.eu-central-1.amazonaws.com"
+    mainnet_staging_PUSH_NOTIFICATION_ENABLED: "true"
```
andreabadesso, line 46: Should we disable it? Or are we going to configure a different firebase project to also have it in this staging env?

luislhl: I discussed this with @alexruzenhack before creating this PR, and I believe he consulted with @pedroferreira1 as well, and it was decided that we should enable them and use the same Firebase project as used in the `mainnet-production`.

I myself prefer not to give an opinion on this because I don't have a full picture of how this works.

I think you guys are in a best position to decide about this and I can just change the configuration if needed.

But I'm open to participate in the discussions as well, even to better understand how things work and be able to give my opinion.

andreabadesso: I think it would be great to have push notifications on this staging env, didn't know you had discussed this previously

Thanks

andreabadesso (approved):

PR #408, opened 2023-04-05
feat: flag to disable nft auto review

luislhl:
### Acceptance Criteria
- We should have a flag to disable the nft auto-review

### Security Checklist


src/utils/nft.utils.ts
```diff
@@ -17,7 +17,19 @@ export const MAX_METADATA_UPDATE_RETRIES: number = parseInt(process.env.MAX_META
  * A helper for generating and updating a NFT Token's metadata.
  */
 
+/** This env-var based feature toggle can be used to disable this feature */
+export const isNftAutoReviewEnabled = (): boolean => process.env.NFT_AUTO_REVIEW_ENABLED === 'true';
```
tuliomir, line 21: I'll raise the question as I've seen it before on the Wallet Service code, but I'm unsure about the real gains it provides for the trouble: should we transform this `process.env` query into a constant?

This is implemented on the `PUSH_NOTIFICATION_ENABLED` feature flag and the rationale is that the lambda keeps this result in memory and avoid re-requesting it from the environment variables on each subsequent call.
https://github.com/HathorNetwork/hathor-wallet-service/blob/59a6daa12cc2bc40c009f2b35ea3b9fbec68a358/src/utils/pushnotification.utils.ts#L74-L75

Do you have an opinion about this, @andreabadesso ?

andreabadesso: But `process.env` is already in memory, what's the difference?

tuliomir: I found the complex environment code that I was thinking about when I raised this issue. It's the `PUSH_ALLOWED_PROVIDERS` created on #372 .
https://github.com/HathorNetwork/hathor-wallet-service/blob/59a6daa12cc2bc40c009f2b35ea3b9fbec68a358/src/utils/pushnotification.utils.ts#L91-L99

So it's about another situation, with more business logic involved. I believe, then, that this PR's situation of only storing a boolean is good the way it is.

tuliomir (approved):

andreabadesso (approved):

Issue #291, opened 2022-09-14
chore: make the deploy fail when an error is found on the deploy


andreabadesso:
More details on https://github.com/HathorNetwork/on-call-incidents/issues/65

luislhl:
I've tried reproducing this after the changes in https://github.com/HathorNetwork/hathor-wallet-service/pull/409, but I couldn't.

My test consisted of creating a wrong migration like the one below, and running the build:

```js
'use strict';

module.exports = {
  up: async (queryInterface, Sequelize) => {
    await queryInterface.removeColumn('push_devices', 'wefwefwef');
  },

  down: async (queryInterface, Sequelize) => {
    await queryInterface.removeColumn('push_devices', 'wefwefwef');
  }
};
```

It just errored out with `ERROR: Can't DROP 'wefwefwef'; check that column/key exists` and stopped the build, like expected.

So I would say we could close this issue already.

This is the link to the build logs: https://eu-central-1.console.aws.amazon.com/codesuite/codebuild/769498303037/projects/hathor-wallet-service/build/hathor-wallet-service%3A4e8241bc-6baf-44b7-ab30-cbd418e3c4f6/?region=eu-central-1

luislhl:
Reviewers @andreabadesso and @alexruzenhack

andreabadesso:
Oh! This was fixed in https://github.com/HathorNetwork/hathor-wallet-service/pull/293, I forgot to close this issue, sorry!

Approved

alexruzenhack:
Ack.

PR #410, opened 2023-04-11
chore: improve rollback mechanism

luislhl:
### Acceptance Criteria
- We should have a way to easily and explicitly rollback the wallet-service to a previous version, including the testnet-production and mainnet-production environments.

### Security Checklist


.codebuild/build.sh
```diff
@@ -1,17 +1,34 @@
...
 echo "Building git ref ${GIT_REF_TO_DEPLOY}..."
 
-if expr "${GIT_REF_TO_DEPLOY}" : "master" >/dev/null; then
+# Checks whether there is a file called "rollback_production", which is used by our other CodeBuild to indicate that this is a rollback
+if [ -f "rollback_production" ]; then
+    # Gets all env vars with `testnet_` prefix and re-exports them without the prefix
+    for var in "${!testnet_@}"; do
+        export ${var#testnet_}="${!var}"
+    done
+    make migrate;
```
luislhl: Maybe we don't need this since we are just rolling back the application and there will be no migrations to run.

alexruzenhack: Should we rollback the migration? Is it safe?

docs/SOP.md
alexruzenhack, line 1: I changed the channel reference to `wallet-service-deploys`: https://github.com/HathorNetwork/hathor-wallet-service/pull/410/commits/5b48f5ca6bab2b821088269d1bd1592957215801

docs/SOP.md
```diff
@@ -34,6 +34,16 @@ Then, you need to add it in [.codebuild/buildspec.yml](https://github.com/Hathor
...
+## Creating a new DB migration
+
+To create a new DB migration, run:
+
+```bash
+npx sequelize migration:generate --name "<migrate-name>"
+```
+
+It will create an empty migration file for you. You should include your migration logic there.
+
```
alexruzenhack: Should we also mention how to test the migration in the build? As you did here https://github.com/HathorNetwork/hathor-wallet-service/issues/291#issuecomment-1504023966?

.codebuild/build.sh
```diff
@@ -0,0 +1,61 @@
...
+    make deploy-lambdas-testnet;
+    send_slack_message "Rollback performed on testnet-production to: ${GIT_REF_TO_DEPLOY}"
+
+    # Gets all env vars with `mainnet_` prefix and re-exports them without the prefix
+    for var in "${!mainnet_@}"; do
+        export ${var#mainnet_}="${!var}"
+    done
+    make migrate;
+    make deploy-lambdas-mainnet;
+    send_slack_message "Rollback performed on mainnet-production to: ${GIT_REF_TO_DEPLOY}"
```
alexruzenhack: Should we have a mechanism to only rollback the mainnet?

alexruzenhack (approved):
I have some comments, but they don't block the PR.

docs/SOP.md
luislhl: Thanks

.codebuild/build.sh
luislhl: We could have separate rollbacks for them. I think it's better indeed. Will change that.

.codebuild/build.sh
luislhl: That's a good question, @andreabadesso also raised this on Slack.

Personally I would prefer not to. I think we would be probably rolling back the migrations when there is no need to. And I think some migrations could be problematic to undo, be it because they will delete something or because they could take a long time to run.

If the changes in the database are backwards compatible, there is no reason to undo them.

So I suggest we adopt a process of letting a human decide whether he/she wants to undo the migrations when performing a rollback.

If we always take care of only changing the database in a backwards compatible way, we will rarely need to undo migrations during rollbacks. This is called the [Expand / Contract model](https://www.red-gate.com/blog/how-to-avoid-rollbacks-in-database-devops#expand-contract) of performing database changes.

But I'm obviously open to adjusting the CI/CD to other ways of working with database rollbacks that the team prefers. I would say the best approach is the one the wallet-service team will be most comfortable with.

But to undo migrations in this rollback script, we would have to first identify which ones need to be undone. @andreabadesso suggested finding the migrations to undo by running a diff of the current tag and the one to where we are rolling back. This could work, but I would need more time to implement and test.

docs/SOP.md
luislhl: Actually my intention there was to test what happens in the CI/CD script when the migration fails, it was not to test the migration.

I think testing the migration should be something the dev will initially do on his/her own machine.

Then we will have the builds of the `dev` and `mainnet-staging` environments, which will run the migrations in their databases. This also serves as a test.

But if you want I can add a command to tell how to run the new migration locally. What do you think?

.codebuild/build.sh
alexruzenhack: I believe we should adopt the principle of "backward compatibility", at least regarding database. This way we can avoid this complexity. With this principle we can keep focus on software control over features using the feature toggle strategy and other strategies that suits our needs.

If @andreabadesso agree we can write it down as a project principle.

docs/SOP.md
alexruzenhack: You right!
And yes, it would be good to have a command to run the migration, but I believe a reference to [local-database](https://github.com/HathorNetwork/hathor-wallet-service/blob/dev/README.md#local-database) is sufficient.

.codebuild/build.sh
andreabadesso: I agree with keeping it simple and adopting the principle of backwards compatibility on new migrations

andreabadesso (approved):

alexruzenhack (approved):

.codebuild/build.sh
luislhl: Great. I had added something about keeping this backwards compatibility in the release guide: https://github.com/HathorNetwork/ops-tools/pull/636/files#diff-fcef9272de4009302f091a39b5b816d0c4dd2965b357a036bfc470c9f674b53eR45

I think you could check there and if you think it's needed, we can improve or add more info there.

andreabadesso (approved):

alexruzenhack (approved):

PR #412, opened 2023-04-14
fix: clean voided tx before reinserting

andreabadesso:
### Acceptance Criteria
- Should fix https://github.com/HathorNetwork/on-call-incidents/issues/64
- Should clean the database for voided `address_tx_history`, `wallet_tx_history` and `tx_output` rows

### Security Checklist

### DIFF
--- a/src/txProcessor.ts
+++ b/src/txProcessor.ts
@@ -96,6 +96,7 @@ export const onNewTxEvent = async (event: SQSEvent): Promise<APIGatewayProxyResu
   // TODO not sure if it should be 'now' or max(now, tx.timestamp), as we allow some flexibility for timestamps
   const now = getUnixTimestamp();
   const blockRewardLock = parseInt(process.env.BLOCK_REWARD_LOCK, 10);
+
   for (const evt of event.Records) {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
@@ -344,11 +345,14 @@ const _unsafeAddNewTx = async (_logger: Logger, tx: Transaction, now: number, bl
   const voidedTx = await checkTxWasVoided(mysql, txId);
 
   if (voidedTx) {
+    logger.info(`Transaction ${txId} received and was voided on databse`, {
+      tx,
+    });
     // this tx was already in the database in the past as voided and is now valid
     // again, we need to cleanup the tx_output and address_tx_history tables so we
     // can safely add it again. Balances were already re-calculated on the handleReorg
     // method, so we don't need to handle that here.
-    await cleanupVoidedTx(mysql, dbTx);
+    await cleanupVoidedTx(mysql, txId);
   }
 
   let heightlock = null;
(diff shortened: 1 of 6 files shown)


src/txProcessor.ts
```diff
@@ -337,6 +341,20 @@ const _unsafeAddNewTx = async (_logger: Logger, tx: Transaction, now: number, bl
     return;
   }
 
+  // check if this tx was already on the database in the past and got voided:
+  const voidedTx = await checkTxWasVoided(mysql, txId);
+
+  if (voidedTx) {
+    logger.info(`Transaction ${txId} received and was voided on databse`, {
```
luislhl:
```suggestion
    logger.info(`Transaction ${txId} received and was voided on database`, {
```

src/txProcessor.ts
```diff
@@ -337,6 +341,20 @@ const _unsafeAddNewTx = async (_logger: Logger, tx: Transaction, now: number, bl
     return;
   }
 
+  // check if this tx was already on the database in the past and got voided:
+  const voidedTx = await checkTxWasVoided(mysql, txId);
```
luislhl, line 345: A few lines above, we do a `fetchTx` to get the tx from the `transaction` database, which also contains information about whether the tx is voided.

Couldn't we avoid adding a new query to the database if we used the results from `fetchTx` to check if it's voided?

The only problem is that `fetchTx` currently filters out voided transactions. But if we did another query that doesn't filter them out, then we could just check in the code whether it's voided or not, with just one query instead of two.

Does it make sense?

tests/db.test.ts
```diff
@@ -1769,6 +1772,83 @@ test('addTx, fetchTx, getTransactionsById and markTxsAsVoided', async () => {
...
+    balance: 0,
+    timestamp: 1,
+    voided: true,
+  }]);
+
+  await addToWalletTxHistoryTable(mysql, [
+    [walletId, txId, tokenId, 0, 0, true],
+  ]);
+
+  await cleanupVoidedTx(mysql, txId);
```
tuliomir, line 1839: For extra safety I would suggest adding a non-voided transaction and running this method on it, confirming that the relevant rows on these tables are kept unaltered.

src/txProcessor.ts
andreabadesso: It definetly would make sense, the problem is that ATM the transactions get DELETED from database during a reorg, as you can see [here](https://github.com/HathorNetwork/hathor-wallet-service/blob/1f6590e00d05371cf6712d171aca37be903e1037/src/db/index.ts#L2017) which is called from [here](https://github.com/HathorNetwork/hathor-wallet-service/blob/1f6590e00d05371cf6712d171aca37be903e1037/src/commons.ts#L537)

tests/db.test.ts
andreabadesso: Done in 3baed2dfd7e763dad1b5ac3ea6e02ecb49e1df0c

luislhl (approved):

tuliomir (approved):

PR #413, opened 2023-04-17
feat: add support to ios on FCM notification

alexruzenhack:
### Acceptance Criteria
- Add support to iOS notification

### Security Checklist

### DIFF
--- a/src/utils/pushnotification.utils.ts
+++ b/src/utils/pushnotification.utils.ts
@@ -163,6 +163,32 @@ export class PushNotificationUtils {
          */
         priority: 'high',
       },
+      apns: {
+        headers: {
+          /**
+           * FCM requires priority 5 for data message, other priority is reject with error.
+           * See https://firebase.google.com/docs/cloud-messaging/concept-options#setting-the-priority-of-a-message
+           *
+           */
+          'apns-priority': '5',
+        },
+        payload: {
+          /**
+           * Background notification flag.
+           * It is labeled as low priority and may not be delivered by the platform.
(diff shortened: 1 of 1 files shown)


tuliomir (approved):

src/utils/pushnotification.utils.ts
```diff
@@ -163,6 +163,32 @@ export class PushNotificationUtils {
          */
         priority: 'high',
       },
+      apns: {
+        headers: {
+          /**
+           * FCM requires priority 5 for data message, other priority is reject with error.
```
andreabadesso, line 169: Is this a FCM limitation? Or do we need explicit approval from Apple to push high priority notifications or something like this? I don't think this is a problem tbh, but would love to be able to send high priority push notifications

andreabadesso (approved):

src/utils/pushnotification.utils.ts
alexruzenhack: This is an Apple (APNs) constraint actually, but only for background message, which is the case here. The FCM call the background message as "data type message". FCM puts this strict rule over "apns-priority" only to avoid error messages on the APNs side.

alexruzenhack: We can push time-sensitive notification on iOS through the Notifee when we receive the background message from Firebase, though.

However, if this configuration somehow get in our way to shuffle time-sensitive notification, we should rethink the current notification architecture actually, because we execute localization in the client, therefore blocking us from send the notification as a foreground message from FCM directly.

PR #415, opened 2023-04-19
fix: add missing indexes

andreabadesso:
Note:

We should manually insert the `20230419031529-tx-ouput-add-spentby-voided-locked-index.js` migration to the `SequelizeMeta` table on the `mainnet` database to prevent it from running, as it already has those indexes

### Acceptance Criteria
- Adds indexes to the `spent_by`, `voided` and `locked` columns

### Security Checklist


tuliomir (approved):

db/migrations/20230419031529-tx-ouput-add-spentby-voided-locked-index.js
luislhl, line 1: What will happen when we run this migration in the mainnet database? Will it have duplicate indexes for the same columns?

andreabadesso: It will fail, we should add the migration to the SequelizeMeta table, I'll add this to the pr description

luislhl (approved):

luislhl (approved):

tuliomir (approved):

PR #425, opened 2023-06-01
feat: added getAddressAtIndex method and API

andreabadesso:
### Motivation

Currently the `getAddressAtIndex` method in the wallet service facade is broken since it expects `this.seed` to have the seed loaded and we don't keep it in memory, so it derives an address from an empty Mnemonic:

https://github.com/HathorNetwork/hathor-wallet-lib/blob/652b94e72021dd11cd0372d3a5b750d9d83f0e29/src/wallet/wallet.ts#L991

We could ask for the user's PIN and use his seed to derive the address at an index, but that would also force us to also ask the PIN on the fullnode facade, which is something we don't want to do.

### Acceptance Criteria
- We should be able to fetch an address by its index
- We should include `flake.nix`, `flake.lock` and `direnv` files to setup a working nodejs14 environment using nix flakes

### Security Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,24 @@
+{
+  description = "virtual environments";
+
+  inputs.devshell.url = "github:numtide/devshell";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+
+  outputs = { self, flake-utils, devshell, nixpkgs }:
+
+    flake-utils.lib.eachDefaultSystem (system: {
+      devShell =
+        let pkgs = import nixpkgs {
+          inherit system;
+
+          overlays = [ devshell.overlay ];
+        };
+        in
+        pkgs.devshell.mkShell {
+          packages = with pkgs; [
+            nixpkgs-fmt
+            nodejs-14_x
+          ];
+        };
+    });
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,21 @@
+/*
+This flake.nix file creates a virtual environment with the desired dependencies 
+in a reproducible way using Nix and Nix flakes (https://nixos.wiki/wiki/Flakes)
+
+Flakes are a feature in Nix that allows you to specify the dependencies of your
+project in a declarative and reproducible manner. It allows for better isolation,
+reproducibility, and more reliable upgrades.
+
+`direnv` is an environment switcher for the shell. It knows how to hook into 
+multiple shells (like bash, zsh, fish, etc...) to load or unload environment
+variables depending on the current directory. This allows project-specific
+environment variables without cluttering the "~/.profile" file.
+
+This flake file creates a shell with nodejs v14.x installed and should work
+on macOs, linux and windows
+*/
 {
-  description = "virtual environments";
+  description = "Flake that installs Node.js 14.x via direnv";
 
   inputs.devshell.url = "github:numtide/devshell";
   inputs.flake-utils.url = "github:numtide/flake-utils";
@@ -16,7 +32,6 @@
         in
         pkgs.devshell.mkShell {
           packages = with pkgs; [
-            nixpkgs-fmt
             nodejs-14_x
           ];
         };
--- a/src/api/addresses.ts
+++ b/src/api/addresses.ts
@@ -7,17 +7,19 @@
 
 import 'source-map-support/register';
 
-import Joi from 'joi';
+import Joi, { ValidationError } from 'joi';
 import { APIGatewayProxyHandler } from 'aws-lambda';
 import { ApiError } from '@src/api/errors';
 import { closeDbAndGetError, warmupMiddleware } from '@src/api/utils';
 import {
   getWallet,
   getWalletAddresses,
+  getAddressAtIndex as dbGetAddressAtIndex,
 } from '@src/db';
-import { AddressInfo } from '@src/types';
+import { AddressInfo, AddressAtIndexRequest, Severity } from '@src/types';
 import { closeDbConnection, getDbConnection } from '@src/utils';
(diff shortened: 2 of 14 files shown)


flake.nix
```diff
@@ -0,0 +1,24 @@
+{
```
luislhl, line 17: If this file accepts comments, I think we should explain what this is.

Maybe also mention in the README (or some other doc that we would link in the README) that we have these configs and how to use them.

src/db/index.ts
```diff
@@ -7,7 +7,7 @@
 import { strict as assert } from 'assert';
 import { ServerlessMysql } from 'serverless-mysql';
 import { get } from 'lodash';
-import { OkPacket } from 'mysql';
+import { OkPacket } from 'mysql2';
```
luislhl: Why this change was needed?

src/db/index.ts
```diff
@@ -3157,3 +3157,38 @@ export const getUnsentTxProposals = async (
...
+ */
+export const getAddressAtIndex = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+  index: number,
+): Promise<AddressInfo | null> => {
+  const addresses = await mysql.query<AddressInfo[]>(
+    `
+    SELECT \`address\`, \`index\`, \`transactions\`
+      FROM \`address\` pd
```
luislhl, line 3178: For an address to be available in this table, we need to have had a transaction involving the address, right?

I'm asking this because, if this is true, then this new API you're building won't be able to actually derive a new address at an index, right?

I don't know how this API will be used to evaluate whether this is a limitation or not

src/api/addresses.ts
```diff
@@ -7,17 +7,19 @@
...
 import {
   getWallet,
   getWalletAddresses,
+  getAddressAtIndex as dbGetAddressAtIndex,
 } from '@src/db';
-import { AddressInfo } from '@src/types';
+import { AddressInfo, AddressAtIndexRequest, Severity } from '@src/types';
 import { closeDbConnection, getDbConnection } from '@src/utils';
 import { walletIdProxyHandler } from '@src/commons';
+import { addAlert } from '@src/utils/alerting.utils';
```
tuliomir: Variables `Severity` and `addAlert` are not being used

src/db/index.ts
tuliomir: I'm uncertain if this change should be made here. I feel that having it in its own PR would make it easy for future developers to understand that:
1) No other changes on the application were necessary (in case of a one-line PR), or 
2) What changes had to be made with this lib update

src/types.ts
```diff
@@ -709,6 +709,10 @@ export interface PushDelete {
   deviceId: string,
 }
 
+export interface AddressAtIndexRequest {
+  index: number,
```
tuliomir: This parameter is optional, right?
```suggestion
  index?: number,
```

tests/api.test.ts
```diff
@@ -149,14 +149,48 @@ test('GET /addresses', async () => {
...
+  // we should be able to filter for a specific index
+  event = makeGatewayEventWithAuthorizer('my-wallet', {
+    index: String(addresses[0].index),
+  });
+  result = await addressesGet(event, null, null) as APIGatewayProxyResult;
+  returnBody = JSON.parse(result.body as string);
+
+  expect(result.statusCode).toBe(200);
+  expect(returnBody.success).toBe(true);
+  expect(returnBody.addresses).toContainEqual({
```
tuliomir: I think this test would be more precise using `toStrictEqual`, since we know the `addresses` array will contain exactly one element.

src/db/index.ts
andreabadesso: We have both `mysql` and `mysql2` on the project, `mysql2` is used by sequelize during migrations

Both `OkPacket` are exactly the same on both libraries, I changed this because I needed a type from `mysql2` but ended up not using it and forgot to rollback to `mysql` instead of `mysql2`

Changed `mysql2` to `mysql` in 79b8d38b5a7f3c82f114191cac70415806cbb9fe

src/db/index.ts
andreabadesso: You're partially correct

We also make sure that the wallet always have at least `MAX_GAP` addresses starting from the last used index

So you're correct that this API won't be able to actually derive a new address at an index, but that's the expected behavior since this is exactly what the old facade does, so this is definetly a limitation but is not a problem for our use case

I added a notice in the API docs in 724fde24511dc475c05a0af2691a49a7a2f18399

src/api/addresses.ts
andreabadesso: Fixed in 88e2a0d21e03a8aa62794c1ef23024706008b653


thanks

src/types.ts
andreabadesso: It is, thanks!

Fixed in 8b3a064baafd4fcd5dedc94b242b4613aa8c7792

tests/api.test.ts
andreabadesso: Fixed in f3c50182de3f3109285a1b2ff2ebf35bfc4be05d

Thanks!

Also fixed a bug when index was 0

flake.nix
andreabadesso: Done in 5c9ab39842230a66ef94e45dcbed248de953fef0

tests/api.test.ts
tuliomir: Maybe it would be a good idea to also include this on the tests?
It would help ensure this bug is not re-inserted later: mixing `0` with `falsey` on javascript is pretty common!

luislhl (approved):

tuliomir (approved):

PR #426, opened 2023-06-19
refactor: using the hathor prefix on verifyMessage

andreabadesso:
### Acceptance Criteria
- After https://github.com/HathorNetwork/hathor-wallet-lib/pull/546 is merged, we need to expect the new message prefix on the wallet-service to verify signatures

### Security Checklist

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -334,7 +334,16 @@ export const verifySignature = (
   try {
     const message = String(timestamp).concat(walletId).concat(address);
 
-    return bitcoinMessage.verify(message, address, Buffer.from(signature, 'base64'));
+    return bitcoinMessage.verify(
+      message,
+      address,
+      Buffer.from(signature, 'base64'),
+      // Different from bitcore-lib, bitcoinjs-lib does not prefix the messagePrefix
+      // length on the message, so we need to do this by using a "End of Transmission
+      // Block" with the length (22) in hex (17). This is the same thing that is done
+      // for the default Bitcoin message (\u0018Bitcoin Signed Message:\n).
+      '\u0017Hathor Signed Message:\n',
+    );
   } catch (e) {
     // Since this will try to verify the signature passing user input, the verify method might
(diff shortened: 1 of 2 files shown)


tests/api.test.ts
```diff
@@ -54,6 +54,10 @@ import {
 } from '@tests/utils';
 import fullnode from '@src/fullnode';
 
+// Monkey patch bitcore-lib
+
+bitcore.Message.MAGIC_BYTES = Buffer.from('Hathor Signed Message:\n');
```
r4mmer, line 59: Since we're using bitcoinjs for verification, do we need this?

andreabadesso: Since the lib is using bitcore-lib, I thought it would be a good idea to also use bitcore-lib to generate the mesages

src/utils.ts
```diff
@@ -334,7 +334,16 @@ export const verifySignature = (
...
 
-    return bitcoinMessage.verify(message, address, Buffer.from(signature, 'base64'));
+    return bitcoinMessage.verify(
+      message,
+      address,
+      Buffer.from(signature, 'base64'),
+      // Different from bitcore-lib, bitcoinjs-lib does not prefix the messagePrefix
+      // length on the message, so we need to do this by using a "End of Transmission
+      // Block" with the length (22) in hex (17). This is the same thing that is done
+      // for the default Bitcoin message (\u0018Bitcoin Signed Message:\n).
```
pedroferreira1, line 344: Why `0018` in the bitcore lib and `0017` in bitcoinjs?

andreabadesso: Hathor Signed Message:\n has 23 characters (0x17), bitcoin has 24

pedroferreira1: Ahhh I misread the comment. I thought it was comparing what we have with bitcore-lib and bitcoinjs

pedroferreira1 (approved):

r4mmer (approved):

